PR #13, opened 2023-06-17
Fix formatting of speakers.yml file

ilrock:
When running `rails db:seed` you'd encounter this error

```
Psych::SyntaxError: (<unknown>): mapping values are not allowed in this context at line 3 column 10
```

because of formatting of the `speakers.yml` file. This PR fixes the format in the file.


adrienpoly:
Thanks

PR #14, opened 2023-06-18
Fix db setup command in README.md

domchristie:


README.md
```diff
@@ -17,7 +17,7 @@ You can use the `.env.sample` file as a guide for the environment variables requ
 To prepare your database and seed content, run:
 
 ```
-bin/rails setup
+bin/rails db:setup
```
adrienpoly, line 20: Thanks for catching that, I think I wanted to write bin/setup

```suggestion
bin/setup
```

Issue #8, opened 2023-06-17
Fix the back navigation 


adrienpoly:
The back navigation doesn't get me back to where I was.

Two reproduction case:

### Navigation
1. navigate to /talks
2. select page 2
3. select a video
4. press back to the list

The user is taken back to the first page, it should be the second page.

### Search

1. navigate to /talks
2. search for something
3. select a video
4. press back to the list

Once again we should come back to the initial search results.
relates to #4

Ukaypromise:
I would love to handle this issue.

adrienpoly:
sure all yours @Ukaypromise

Ukaypromise:
Thanks @adrienpoly

adrienpoly:
closed by #26

PR #15, opened 2023-06-18
Use Turn to handle transitions

domchristie:
Firstly, love this project ❤️

This pull request demonstrates using Turn for handling View Transitions. Rather than using Stimulus controllers to change add a `view-transition-name` it server-renders them instead e.g. `style="view-transition-name: talk_123_image"`. This is rendered on both the cards and show templates.

The potential issue with this approach is if a card of the same talk is rendered more than once on the same page, it might error.

Turn handles preview visits so we can make use of Turbo caching, too.


app/javascript/entrypoints/application.js
```diff
@@ -12,24 +16,4 @@ Turbo.start();
 // Example: Import a stylesheet in app/frontend/index.css
 import "../../assets/stylesheets/application.tailwind.css";
 
-addEventListener("turbo:before-render", (event) => {
-  if (document.startViewTransition) {
-    event.preventDefault();
-
-    document.startViewTransition(() => {
-      event.detail.resume();
```
domchristie, line 20: Have you seen any weird race conditions with this approach? 

I _think_ the Turbo rendering process is asynchronous, and `event.detail.resume()` just tells Turbo to resume rendering, so there's possibility that the View Transitions API captures the markup before Turbo has render the new page.

It might be preferable to return a promise in the `startViewTransition` callback that resolves on `turbo:render`

adrienpoly: I don't think I could spot some weird race condition but  this is pretty new so my testing hasn't been extensive

app/javascript/entrypoints/application.js
```diff
@@ -12,24 +16,4 @@ Turbo.start();
...
-
-addEventListener("turbo:before-frame-render", (event) => {
-  if (document.startViewTransition) {
-    event.preventDefault();
-
-    document.startViewTransition(() => {
-      event.detail.resume();
-    });
-  }
-});
```
adrienpoly, line 33: There is a tiny regression by removing those lines. They were used to transition the frame navigation. The best example is when you perform a search the results are displayed in a frame and with a subtle fade transition.

Can we get the same thing with Turn?

app/views/layouts/application.html.erb
```diff
@@ -2,7 +2,6 @@
 <html lang="en">
   <head>
     <meta name="viewport" content="width=device-width,initial-scale=1">
-    <meta name="turbo-cache-control" content="no-cache">
```
adrienpoly, line 5: I noticed a very tiny blink of the right list of videos.
The root cause is that for now, this list is a random selection of 5 talks so it changes for each visit.
https://github.com/adrienpoly/rubyvideo/assets/7847244/9a8fcd89-4478-4793-856f-92e2bb173fbb

I see two solutions:
1. add a Stimulus controller and remove this list content on disconnect
2. change the way we generate this list to have the same result for every visit

app/views/talks/_card.html.erb
```diff
@@ -1,13 +1,13 @@
...
-  <%= link_to talk_path(talk, speaker_slug: local_assigns.fetch(:speaker_slug, nil)), class: "flex aspect-video", data: { controller: "view-transition", view_transition_name_value: "banner-img" } do %>
+  <%= link_to talk_path(talk, speaker_slug: local_assigns.fetch(:speaker_slug, nil)), class: "flex aspect-video" do %>
     <%= image_tag talk.thumbnail_sm,
                   srcset: ["#{talk.thumbnail_lg} 2x"],
                   id: dom_id(talk),
                   height: 174,
                   width: 310,
                   alt: "talk by #{talk.speakers.join(", ")}: #{talk.title}",
-                  style: view_transition_style(talk_id: talk.id, from_talk_id:  local_assigns.fetch(:from_talk_id, nil), name: "banner-img"),
+                  style: "view-transition-name: #{dom_id(talk)}_image",
```
adrienpoly, line 10:
```suggestion
                  style: "view-transition-name: #{dom_id(talk, :image)}",
```

app/views/talks/_card_horizontal.html.erb
```diff
@@ -2,12 +2,8 @@
...
 
   <%= link_to talk_path(talk),
-            class: "flex aspect-video shrink-0 relative",
-            data: {
-              controller: "view-transition",
-              view_transition_name_value: "card-horizontal-img"
-            } do %>
-    <%= image_tag talk.thumbnail_sm, srcset: ["#{talk.thumbnail_lg} 2x"], id: dom_id(talk), class: "w-full h-auto object-cover", loading: :lazy %>
+            class: "flex aspect-video shrink-0 relative" do %>
+    <%= image_tag talk.thumbnail_sm, srcset: ["#{talk.thumbnail_lg} 2x"], id: dom_id(talk), class: "w-full h-auto object-cover", style: "view-transition-name: #{dom_id(talk)}_image", loading: :lazy %>
```
adrienpoly, line 6:
```suggestion
    <%= image_tag talk.thumbnail_sm, srcset: ["#{talk.thumbnail_lg} 2x"], id: dom_id(talk), class: "w-full h-auto object-cover", style: "view-transition-name: #{dom_id(talk, :image)}", loading: :lazy %>
```

app/views/talks/_talk.html.erb
```diff
@@ -2,11 +2,9 @@
 
   <div <%= tag.attributes(
             class: "aspect-video banner-img card-horizontal-img relative",
-            style: "view-transition-name: card-horizontal-img",
+            style: "view-transition-name: #{dom_id(talk)}_image",
```
adrienpoly, line 5:
```suggestion
            style: "view-transition-name: #{dom_id(talk, :image)}",
```

adrienpoly (commented):
Thanks @domchristie for this PR it really makes it more simple. I have seen a few tiny things and will keep on playing with it a bit but loving it.


### Page preview
I am curious (I probably need to dig a bit into the code of Turn) how do you handle to previews during the page transition?

Let's imagine the transition last 500ms and the page response is 300ms
Will the transition start with the preview and then keeps on doing the animation for the last 200ms with the response? Having a hard time understanding how this all work 😄

domchristie:
> I am curious (I probably need to dig a bit into the code of Turn) how do you handle to previews during the page transition?
>
> Let's imaging the transition last 500ms and the page response is 300ms
> Will the transition start with the preview and then keeps on doing the animation for the last 200ms with the response?

Yep, that's it. I think this works because when the View Transition API transitions between the two versions, the "`::view-transition-new(root)` is a live representation of the new view" (https://developer.chrome.com/docs/web-platform/view-transitions/#how-these-transitions-work). So I suppose `::view-transition-new(root)` is the preview initially but then gets swaps out with the response.

app/views/layouts/application.html.erb
domchristie: Yes, this is an odd user experience. Another option would be to disable previews for that view. I often do this by setting `@turbo_cache_control = :no_preview` in the template.

app/javascript/entrypoints/application.js
adrienpoly: hum very strange for Meilisearch you should have a folder `data.ms` can you try 

- stop all Meillisearch process 
- deleting this folder
- start the app again
- rails c
- Talk.reindex!

app/views/layouts/application.html.erb
adrienpoly: will change that perfect solution

app/javascript/entrypoints/application.js
adrienpoly: And yes if you can revert this change from the test I did locally it does work with your changes

domchristie: I've not used meilisearch before. I installed it via brew, but the meilisearch-rails download/run link 404s so I gave up. (I'm not using docker.)

adrienpoly: I am not using docker neither. I think I used the cURL or brew. Is the Meillisearch process running on port 7700?

domchristie: No, if I run `meilisearch`, it just errors with the above message

domchristie: It's weird. I've never used it before, so no idea why I'm getting an upgrade error

adrienpoly:
I played again a bit, It works really well. I tried in 3G mode and couldn't spot any animation issues.

There is only one issue when navigating back from a search but that's because I always for back to the talks home page and not to the previous search results. Correction this #8 will help.
https://github.com/adrienpoly/rubyvideo/assets/7847244/9f9f9410-48bc-4f2c-9ff1-4130eba6eb49

Correct me if I am wrong but in the previous version I would add the view-transition-name on a click navigation event. The name was the same for all cards. As we must have a unique name on the page it has to be added dynamically on the click.
With this approach, it is possible to customize the transitions 

```css
::view-transition-old(banner-img),
::view-transition-new(banner-img) {
...
}
```

Your approach is way simpler with a unique name `image_talk_123` for each card.
but I don't think we can customize the animation with something like 

```css
::view-transition-old(image_talk_*),
::view-transition-new(image_talk_*) {
...
}
```

The good news is I couldn't really spot any difference between what I had before and what we have now. (the CSS was taken from the official examples, I didn't invent it 😄 )

adrienpoly:
merging it thanks for your help on this

domchristie:
> Correct me if I am wrong but in the previous version I would add the view-transition-name on a click navigation event. The name was the same for all cards. As we must have a unique name on the page it has to be added dynamically on the click.
> With this approach, it is possible to customize the transitions

Yep, that's correct, and it's not currently possible to customise these transitions with this approach. This will only apply when there are multiple elements on the same page that share the same semantics e.g. `talk-image` or `talk-title`. For simple cases, I've found that this approach is acceptable, and it's still possible to customise transitions of singular elements e.g. `navbar`. However I definitely think it'd be handy to have some hooks that let developers dynamically add/change `view-transition-name`s when an element is clicked.

PR #16, opened 2023-06-19
add turbo_action advance

### LINKED ISSUE #4: Update url after a search
When we search for something the Url should be updated to reflect this search 

ie `https://rubyvideo.dev/talks?q=aaron`

thomasvanholder:
Added a [turbo_action](https://turbo.hotwired.dev/reference/attributes) attribute to the search form so the search query is visible in the URL.

fixes #4 

CleanShot 2023-06-19 at 15 27 36

### DIFF
--- a/app/views/talks/index.html.erb
+++ b/app/views/talks/index.html.erb
@@ -1,16 +1,16 @@
-<div class="container my-8 w-full gap-4 flex flex-col">
-    <h1 class="title">Talks</h1>
+<div class="container flex flex-col w-full gap-4 my-8">
+  <h1 class="title">Talks</h1>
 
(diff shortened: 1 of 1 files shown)


adrienpoly:
Thanks!

looking at your PR I realize that we need to setup some linters on this project. Do you have the Tailwind prettier linter by default in your IDE?

thomasvanholder:
> 

Yes, indeed. 

We could add it in with a dev dependency, something like:

```
  "devDependencies": {
    "prettier": "^2.8.0",
    "prettier-plugin-tailwindcss": "^0.2.0"
  }
```

adrienpoly:
PR welcome

PR #19, opened 2023-06-20
Store vector embeddings for talks, and display related talks

### LINKED ISSUE #18: Display videos on similar topics?
Hi @adrienpoly, when viewing a video, it might be interesting to have links to other videos on the same topic(s). Would you merge a PR that brings this feature?

crohr:
This PR adds a "Talks you might be interested in" below each talk. This uses the newly released vector support in meilisearch to find talks that share some similarity with the current one.

Support for vector search is currently non-existant in meilisearch-rails and meilisearch-ruby so we cannot easily (?) fetch the similarity scores to keep only the most interesting ones, but the results are quite good as it is.

To make it work, you need to specify an OpenAI API key in `.env`, and launch a reindex as follows:

```
Talk.reembed!
```

Note that vector support is still a bit fiddly, so you may have to start from a fresh meilisearch database if it doesn't work for you on the first try (tip: inspect `GET localhost:7700/tasks` to see if anything is going wrong when indexing).

Demo below:

https://github.com/adrienpoly/rubyvideo/assets/6114/8a7978af-3585-49ae-837e-d8f906c44bf3

* Announcement post on meilisearch - https://github.com/meilisearch/product/discussions/621#discussioncomment-6183647

Closes #18.


app/models/talk.rb
```diff
@@ -49,14 +49,15 @@ class Talk < ApplicationRecord
...
+    add_attribute :_vector do
+      _vector # doesn't work yet
+    end
+    searchable_attributes [:title, :description, :_vector]
     sortable_attributes [:title]
 
     attributes_to_highlight ["*"]
   end
 
-  meilisearch enqueue: true
```
crohr, line 58: @adrienpoly this was actually overriding the previous meilisearch block definition (I spent a while understanding why my `_vector` attribute wouldn't index 😅 )

app/models/talk.rb
```diff
@@ -49,14 +49,16 @@ class Talk < ApplicationRecord
     attribute :speakers do
       speakers.pluck(:name)
     end
-    searchable_attributes [:title, :description]
+    add_attribute :_vectors do
+      vectors
+    end
+    searchable_attributes [:title, :description, :_vector]
```
Kerollmops, line 55: I am pretty sure you don't want to let users search into a list of floats 😄

crohr: Yes indeed :), a leftover from some previous trial and error, thanks for the catch!

config/initializers/meilisearch.rb
```diff
@@ -1,4 +1,4 @@
 MeiliSearch::Rails.configuration = {
   meilisearch_url: Rails.env.local? ? "http://localhost:7700" : "http://91.107.208.207:7700", # example: http://localhost:7700
-  meilisearch_api_key: ENV["MEILI_MASTER_KEY"]
+  meilisearch_api_key: Rails.env.local? ? nil : ENV["MEILI_MASTER_KEY"]
 }
```
adrienpoly, line 4: Thanks I will add this directly to main as it could create issues for users not having a correct local key

adrienpoly:
Thanks @crohr for this prototype it looks very promising.  

## Embeddings

To move forward if we want to have this in production I guess we could as a first step put in place the logic to compute the embeddings and store them in the talk model.  With some kind of logic to recompute them every time title / description changes. 
This should prevent recomputing the embeddings for every reindex

Then Meillisearch would just index that new column from our model.

Can this work or I am missing something?

## UI

I supposed you put the list below for testing. Ultimately this should feed the cards that are on the right and replace this random suggestion. But while developing let's keep it like that. At some point, we might want to deploy this feature to prod behind a feature flag so that will make it easier to test

## Devops

To release it to prod I will need to update the Meillisearch engine. We probably need to wait a bit as it seems pretty new and they highly recommend waiting.  This being said whatever result we get out of it will always be better than a random suggestion.

crohr:
> ## Embeddings
> To move forward if we want to have this in production I guess we could as a first step put in place the logic to compute the embeddings and store them in the talk model. With some kind of logic to recompute them every time title / description changes. This should prevent recomputing the embeddings for every reindex
> 
> Then Meillisearch would just index that new column from our model.
> 
> Can this work or I am missing something?

I tried to selectively tell meilisearch to ignore _vectors when title or description hasn't changed, but it doesn't seem to work. So yes storing there embedding in sqlite and recomputing with classic AR callbacks if title or description changes would work.

> ## UI
> I supposed you put the list below for testing. Ultimately this should feed the cards that are on the right and replace this random suggestion. But while developing let's keep it like that. At some point, we might want to deploy this feature to prod behind a feature flag so that will make it easier to test

Yes, kept it simple for now, you're the UI guy :) Also, I didn't really notice that videos on the right were supposed to be "more like the current one". I think it could make sense to keep both exploratory videos on the right and related videos below the one you've just visioned, but maybe I'm wrong.

> ## Devops
> To release it to prod I will need to update the Meillisearch engine. We probably need to wait a bit as it seems pretty new and they highly recommend waiting. This being said whatever result we get out of it will always be better than a random suggestion.

Not sure how you deploy the meilisearch container / process on the server, but yes since there is nothing in meilisearch that can't be reindexed from sqlite, I think it's ok to deploy alpha/beta software in that case.

Kerollmops:
Hey @crohr 👋 Do you plan to release this recommendation system? We improved the Vector Store solution since then 🧇

Kerollmops:
@adrienpoly Indeed, you need a v1.3.x and [we provide a Docker image](https://hub.docker.com/layers/getmeili/meilisearch/v1.3.2/images/sha256-576e2dd613337c87310ce292c2da7c00578e972984ad9ce2fdb14a5490b4248f?context=explore) and [all sort of binaries](https://github.com/meilisearch/meilisearch/releases/tag/v1.3.2) ☺️

crohr:
> Hey @crohr 👋 Do you plan to release this recommendation system? We improved the Vector Store solution since then 🧇

Sure, I will have another look this week, thanks for the update!

app/models/talk.rb
```diff
@@ -41,26 +52,56 @@ class Talk < ApplicationRecord
...
+  before_save :compute_embedding, if: :must_compute_embedding?
+
   # search
-  meilisearch do
+  meilisearch primary_key: :id, enqueue: true, raise_on_failure: Rails.env.development? do
     attribute :title
     attribute :description
     attribute :slug
-    attribute :video_id
+    # ⚠️ This `video_id` attribute makes indexing (silently) fail with v1.3.2 of meilisearch. Error message from meilisearch (GET /tasks):
```
crohr, line 62: @Kerollmops I wonder if there is a regression here with v1.3.2, since it fails to index even if I specifically set the `primary_key: :id` as above.

app/models/talk.rb
```diff
@@ -41,26 +52,56 @@ class Talk < ApplicationRecord
...
+    # attribute :video_id
     attribute :video_provider
     attribute :thumbnail_sm
     attribute :thumbnail_md
     attribute :thumbnail_lg
     attribute :speakers do
       speakers.pluck(:name)
     end
+    # ⚠️ This must return nil and not an empty array if no vector is available.
+    # Otherwise all other indexing tasks with non-zero vector arrays will silently fail, since the engine will expect all vectors to have the same length.
```
crohr, line 74: @Kerollmops this was by far the most time-consuming issue I have with meilisearch vector indexing: seems like meilisearch rejects any document that doesn't have a vector of the same size as the first one it has seen. If you mistakenly send a zero-sized vector (or incorrectly sized vector) for your first document, then all other documents with a different vector size will be rejected. This should be put in a big warning somewhere, or hopefully have some better error message in the server logs (had to inspect the tasks list to see this error).

app/models/talk.rb
```diff
@@ -41,26 +52,56 @@ class Talk < ApplicationRecord
...
+    filterable_attributes [:id]
 
     attributes_to_highlight ["*"]
   end
 
-  meilisearch enqueue: true
+  # Recomputes embedding for all talks that don't have one yet.
+  def self.reembed!(sleep_interval: 2.seconds, limit: nil)
+    # required for querying vectors (not indexing)
+    MeiliSearch::Rails.client.http_patch "/experimental-features", {vectorStore: true}
```
crohr, line 86: @Kerollmops could this be auto-enabled in some way if we start to send vectors?

crohr:
@adrienpoly ready to be reviewed, I've updated the description with the new task to run

app/models/talk.rb
Kerollmops: Unfortunately not. You must enable it by hand 😞

app/models/talk.rb
Kerollmops: I recently specified the primary key to index one of my datasets and haven't had issues. Are you sure it was introduced in v1.3.2? Is it related to the ruby SDK? I used the `primaryKey` query parameter with the `POST /indexes/{indexname}/documents` route.

brunoocasali: hi @crohr why do you need to specify the primary key?

app/models/talk.rb
Kerollmops: Thank you for the report @crohr. I shared your feedback with the team and will see what we can do. We [recently fixed a bug with the `_vectors` field](https://github.com/meilisearch/meilisearch/pull/4039) that was producing issues when specifying multiple vectors for a single document. This one will be released in v1.3.2.

app/models/talk.rb
crohr: It was just an attempt to fix the issue I'm explaining in the comment, but it doesn't make a difference (and shouldn't since this is the default)

adrienpoly:
thanks @crohr I made a bit of preparation work in #64 to integrate on the front end part the results of the suggestions and to isolate then into a frame so that initial page load for the Talk#show route is not coupled to this suggestion method.

My next step is to upgrade Meilisearch in the prod environment and it is not as plug and play as I hope it would be... Anyway it is not neither a very high traffic site so If search is down for a little time that should be ok 😄


Will try to look back into this soon. Thanks for your work

PR #21, opened 2023-06-22
Fix typo Sandi Metz  instead of Mets

HusseinMorsy:
see https://sandimetz.com


adrienpoly:
thanks. I corrected also the online db

PR #22, opened 2023-06-22
Retrieve video statistics from Youtube

### LINKED ISSUE #11: Retrieve the video statistics from Youtube
With the API we can have access to the video statistics.

It would be nice to retrieve the view comments and +1, to store them in the Talk model.

With this we can display the view on the card and add this as a search/sort metadata #3

ashwin47:
This PR adds video views and like counts to the Talk table.

Closes #11


app/clients/youtube/video_statistics.rb
```diff
@@ -0,0 +1,24 @@
+module Youtube
+  class VideoStatistics < Client
+    def list(video_id)
+      path = "/videos"
+      query = {
+        part: "statistics",
+        id: video_id
+      }
```
adrienpoly, line 8: Little naming reco:

We want to retrieve the statistics to do so we are querying the `/videos` how about naming the `class Videos < Client` and making it more general purpose (ie not returning only the stats information)?


Also I know the Youtube api documentation talk about a "list" for getting a single object. I understand that you name this method accordingly. 
I would prefer a more restful naming such as
`retrieveStatistic` or `getStatistic`

A list I always expect it to return multiple objects

app/clients/youtube/video_statistics.rb
```diff
@@ -0,0 +1,24 @@
...
+      if response.empty?
+        puts "Error: No video found with the given ID"
+        return nil
+      end
+
+      # Assuming there's only one item in the response, as the ID of a video is unique.
+      {
+        view_count: response.first["statistics"]["viewCount"],
+        like_count: response.first["statistics"]["likeCount"]
+      }
```
adrienpoly, line 21: What are the other values returned by the API. Should we return them from that method in case we need them in the future?

adrienpoly (approved):
Perfect thanks for this contribution.

I just had a few thoughts on some namings

Let me know what you think

app/clients/youtube/video_statistics.rb
ashwin47: Makes sense.

app/clients/youtube/video_statistics.rb
ashwin47: It has comment and favorite count additionally. Do we need those? (Not sure what is a favorite, probably something outdated)

adrienpoly:
thanks for taking this one. Will look next how to automate this

PR #26, opened 2023-06-23
Fix back navigation

Ukaypromise:
### Hello 👋🏽 @adrienpoly ,
I believe a better and easy way would be to use `:back`

The :back option in Ruby on Rails provides a convenient way to navigate back to the previous page. It automatically generates the appropriate URL based on the HTTP referer header.

Using :back can be a good option in many cases, especially when you want to implement a generic back button that takes the user back to wherever they came from. It simplifies the code and handles the navigation automatically.


adrienpoly:
Thanks that is an interesting approach. My only concern is when we would arrive from a Google search result then it would take us bac to this search result. Is there a way to provide a fallback?

Ukaypromise:
Okay I get it now. Just give me today, I will fix it properly @adrienpoly

Ukaypromise:
@adrienpoly Made some modifications.

adrienpoly:
Thanks, it does not cover the search results let's say I search for Aaron and visit a show and then come back I won't have the same results but it remains improvements. I'll probably open an issue to improve that further down the road once the search has been stabilized

PR #28, opened 2023-06-24
Show a few speakers

### LINKED ISSUE #10: Fix the card when they are lots of speakers
On the card we should only display a few speakers to prevent such layout

CleanShot 2023-06-17 at 07 31 11@2x

enderahmetyurt:
Screenshot 2023-06-24 at 08 55 11

Close #10

### DIFF
--- a/app/views/talks/_card.html.erb
+++ b/app/views/talks/_card.html.erb
@@ -27,9 +27,16 @@
               <%#= link_to talk.event.name, talk.event %>
               <%= content_tag :div, talk.event.name %>
             <% end %>
-            <% talk.speakers.each do |speaker| %>
+
+            <strong class="mt-5">Speakers:</strong>
+            <% talk.speakers.first(10).each do |speaker| %>
               <%= link_to speaker.name, speaker_path(speaker) %>
             <% end %>
+            <div class="my-2">
+              <% if talk.speakers.count > 10 %>
+                <span class="text-red">and <%= talk.speakers.count - 10 %> more speakers</span>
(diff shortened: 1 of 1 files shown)


marcoroth:
This looks great! 

I think it would be even nicer if it would show the first 2/3 speakers and additionally say "and 10 more" if there are any speakers remaining.

enderahmetyurt:
how does it look now @marcoroth

marcoroth:
Thank you, @enderahmetyurt! 

I was wondering if 10 people are not a bit too much, but this looks great to me otherwise!

enderahmetyurt:
> Thank you, @enderahmetyurt!
> 
> I was wondering if 10 people are not a bit too much, but this looks great to me otherwise!

you might be right. 5 would be enough.

enderahmetyurt:
@adrienpoly what do you think about it?

adrienpoly:
hello thanks for your proposal

I would prefer that we use most of the available space on the lines

here is a quick draft
CleanShot 2023-07-08 at 11 58 50@2x

```erb
            <strong class="mt-5">Speakers:</strong>
            <div class="line-clamp-2">
              <% talk.speakers.shuffle.each do |speaker| %>
                <%= link_to speaker.name, speaker_path(speaker), class: "whitespace-nowrap" %>
              <% end %>
            </div>
```

with comas in between that would be perfect

adrienpoly:
Thanks for your investigation on that one I did some follow up in #44 as I couldn't add commits to this branch.

PR #32, opened 2023-06-24
Configure CSS nesting for Tailwind

marcoroth:
This resolves the following warning:

```
[vite:css] Nested CSS was detected, but CSS nesting has not been configured correctly.
Please enable a CSS nesting plugin *before* Tailwind in your configuration.
See how here: https://tailwindcss.com/docs/using-with-preprocessors#nesting
8  |  @import "components/pagination.css";
9  |  @import "components/transition.css";
10 |  @import "components/typography.css";
   |       ^
11 |  @import "components/video.css";
```

See the Tailwind docs for reference: https://tailwindcss.com/docs/using-with-preprocessors#nesting

### DIFF
--- a/postcss.config.js
+++ b/postcss.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   plugins: {
-    'postcss-nested': {},
     'postcss-import': {},
+    'tailwindcss/nesting': {},
     tailwindcss: {},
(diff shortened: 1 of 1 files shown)


adrienpoly:
thanks again I was wondering where that was coming from

PR #35, opened 2023-06-25
Implementation of RSS and Atom feeds

### LINKED ISSUE #12: RSS feed for all new videos
It would be great to have RSS support to subscribe for all new videos.

krapans:
Both feeds were validated by https://validator.w3.org/feed/. Can be down by using snapshots from tests as well.

RSS

image

Atom

image

Closes #12


app/controllers/feed_controller.rb
```diff
@@ -0,0 +1,14 @@
+class FeedController < ApplicationController
+  skip_before_action :authenticate_user!
+
+  # GET /feed.rss
+  # GET /feed.atom
+  def index
+    @talks = Talk.all.order(created_at: :desc).includes(:speakers).limit(100)
```
krapans, line 7: We could add here in memory caching or persist on disk based on some kind of checksum. Though as SQLite is used in production, not sure if there is going to be a big impact of that. Let me know what you think about this.

test/controllers/feed_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "test_helper"
+
+class TalksControllerTest < ActionDispatch::IntegrationTest
```
krapans, line 3: Regarding the test suite, I made sure that these new tests passes and all is well, though on my local machine all test suite doesn't work, There are quite a few errors which seems to be not related to my changes even if I take in account fixtures.

image

image


Is this a known issue?

test/controllers/feed_controller_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "test_helper"
+
+class FeedControllerTest < ActionDispatch::IntegrationTest
```
krapans, line 3: Regarding the test suite, I made sure that these new tests passes and all is well, though on my local machine all test suite doesn't work, There are quite a few errors which seems to be not related to my changes even if I take in account fixtures.

image

image


Is this a known issue?

app/controllers/feed_controller.rb
adrienpoly: Rails cache is available with this app using Litestack cache (using Sqlite behind the scene). 
You could cache the views using normal Rails cache tools but not sure it really worth it at this point

test/controllers/feed_controller_spec.rb
adrienpoly: yes sorry about that I wanted to release the project ASAP and didn't clean all the test suites yet. https://github.com/adrienpoly/rubyvideo/issues/5
Someone is working on it

adrienpoly:
This looks awesome to me. I was wondering if we really need a feed route/controller vs using the existing talks_controller / route.

The advantage I see using the talks_controller is we could use the filtering https://github.com/adrienpoly/rubyvideo/issues/3 once they are available. And provide a feed icon for every search result. So you could add to your feed Railsconf only by example. 

Not a big RSS user so not sure how people are using this

app/controllers/feed_controller.rb
krapans: Yeah, makes sense. Then lets skip premature optimisation if its not needed.

krapans:
@adrienpoly, kind of makes sense that we would move it to talks controller, I don't see any problems with that. The only concern that I would have I guess is that `@pagy` object would be ignored in case of RSS, because it kind of wouldn't fit in that structure. 

Only worry is hear that we have a one entrypoint, though some formats would use @pagy object and some like RSS/Atom wouldn't. Not sure though if its a dealbreaker.

adrienpoly:
@krapans sorry to bother you with that but I made an error trying to rebase the branch and deleted some codes. If you still have the branch locally can you open a new PR. Sorry about that !

adrienpoly:
@krapans ok I could recover the missing file. Test suite is green now will give it a final test before merge thanks

ChaelCodes:
Thank you so much for opening this and exploring what RSS/Atom feeds could look like on the site. There are a lot of merge conflicts and I don't think we've settled on a clear path for how we want to handle feeds, even now.

We're getting over 100 videos in a month, some of them years old, and some of them recent. Some are recordings, some are just scheduled, and then we have parent and child talks too. There's a lot of unexpected edge cases.

I'm going to close this PR, because I don't think we're going to merge this version of RSS feeds, and I don't think we've decided what video RSS feeds should look like.

Thank you very much for your contribution, and I hope you consider contributing to RubyEvents in the future.

PR #37, opened 2023-06-28
Start cleaning up the test suite

Oli0li:
### What does this PR do?
This PR aims at fixing some issues in the test suite. It:

- removes tests and methods generated by the scaffold that are actually not needed
- fixes incorrect expectations in the tests
- comments out some failing tests that may be useful in the future but would first require a method to be changed or a heading to be added

### What is left to do?
The only tests still erroring on my end when running "bundle exec bin/rails test" are 6 tests using VCR under the test/clients/youtube folder, because of this error:

`
RuntimeError: Neutered Exception VCR::Errors::UnhandledHTTPRequestError:
` 

For your reference, I also noticed that some tests in test/system raise errors when I run them separately, because they are meant to check text on home#index, but there is only a page#home in the routes. These tests are:

- test/system/identity/emails_test.rb
- test/system/passwords_test.rb
- test/system/sessions_test.rb

They also have this warning displayed:

`
Rack::Handler is deprecated and replaced by Rackup::Handler
`
### Context
This PR was made to solve some issues raised in Issue https://github.com/adrienpoly/rubyvideo/issues/5: "Clean the test suite"

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -1,6 +1,6 @@
 class EventsController < ApplicationController
   skip_before_action :authenticate_user!, only: %i[index show]
-  before_action :set_event, only: %i[show edit update destroy]
+  before_action :set_event, only: %i[show edit update]
 
   # GET /events
   def index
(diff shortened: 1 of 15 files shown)


adrienpoly:
thanks for this is it a great addition

PR #39, opened 2023-06-29
add event page

### LINKED ISSUE #7: Create the event page
The event page should be a page displaying on la left the event information (date web site location) and on the right the videos from that event.

dedemenezes:
This PR **implements an events#show page** as discussed.

Close #7 

The page added has the **same structure as the speakers#show.** 
What I'm adding:
1. System test for the events#show, ensuring you can see the event name and talks for the event.
2. Event show page, displaying all talks to the specific event using pagination.
3. Link to events#show on the talks/card partial.

What I'm missing:
1. The back link on talks#show always take the user back to a speakers#show, at the moment it redirect to the first speaker for each talk. I see there is a PR trying to fix it (#26), so I was not sure if I should do it her.
2. The about information for the event is actually coming from the organisation as mentioned in #37. This means that the events#edit is buggy and that's why I kept the edit button as a dead link. Should I remove the button while there is no edit action?
3. The list of speakers is quite big. Should be fixed by #28 

Here is a printscreen how it looks like atm 👇 
localhost_3000_events_rails-conf-2012


adrienpoly:
Thanks this looks very cool and thanks for identifying the various blockers from the ongoing works. Will test it right away

dedemenezes:
Hey @adrienpoly and @marcoroth ✌️ I would love to fix the conflicts and make any adjustments that you think are needed. 
Probably I won't be able to make it this week because I'm moving but I'm fully available next week to work on this too.

adrienpoly:
yes that would be great also last I looked, it would be great to also have an index page in addition to the show page

dedemenezes:
Hey @adrienpoly and @marcoroth, I'm working on this PR this week and I'm wondering how to display the list of events (events#index). 

Considering there are no images or thumbnails associated with events, organizing them like in speakers#index might be more suitable. We can display the list with the number of talks for each event. What do you think?

marcoroth:
Hey @dedemenezes, I think it would be helpful to just have a list somewhere, we can always iterate on the design/features.

But as you say, a list of all the events with the number of talks sounds great to me!

marcoroth:
This look great, thank you @dedemenezes!

The only feedback I have is that it might be a bit hard to find the event you are looking for in that big list, maybe it might make sense to group them by organization? Something like this maybe?

CleanShot 2024-08-21 at 13 35 29

```html+erb
  <% @events.group_by(&:organisation).each do |organisation, events| %>
    <h1><%= organisation.name %></h1>

    <div id="<%= dom_id(organisation, "events") %>" class="grid sm:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4 gap-x-8 lg:gap-x-12 gap-y-2 min-w-full py-8">
      <%= render events %>
    </div>
  <% end %>
```

dedemenezes:
Thank you @marcoroth ✌️ 

I agree with you and it make a lot of sense since they are already associated. Thank you for the insight. I'll make the adjustments and push the new version.

dedemenezes:
Thank you for the commits you did removing duplicate links and fixing lint errors @adrienpoly. The only thing I changed addressing @marcoroth suggestion was the padding between elements: `py-8 => py-4 pb-8.`

marcoroth:
Awesome @dedemenezes, thank you, even better! 🙌🏼

adrienpoly:
@dedemenezes Thanks for this great addition. I have done some slight optimisations mostly to remove the n+1 : 
- eager load the `organisation` 
- add a counter_cache for the talks_count

This way the page loads much faster

marcoroth:
Thanks for the contribution @dedemenezes! 🎉

dedemenezes:
Thank you both for the help on this <3 I'll take a look into the code and issues to see how can I contribute more this project!

PR #46, opened 2023-07-22
Fix top padding

enderahmetyurt:
# Before
Screenshot 2023-07-22 at 09 28 57

# After
Screenshot 2023-07-22 at 09 35 32


adrienpoly:
Thanks

Issue #6, opened 2023-06-17
Add a burger menu in mobile


adrienpoly:
In mobile there is currently no menu available. We should have a burger menu with a dropdown available

code-death:
Hello Adrien can I work on this issue and can you please tell what are the contribution guidelines for this as I couldn't find anything related to issues, branches or PRs in the existing guidelines

awalsh3:
Hello! Wondering if I could take a stab at this? I am new to the community. No worries if it's already in progress!

PR #47, opened 2023-08-03
Burger menu

awalsh3:
ISSUE: #6

This proposed PR adds a burger menu with drop-down links for mobile rendering. This uses tailwind css and plain vanilla javascript. 

**A few questions/notes:**

- It needs some styling decisions. I've left it as close to the main navbar as possible for now. Happy to adjust where needed.
- The drop-down links inherit the same classes used by the _links.html.erb shared partial. To change the styling we could either add another partial for mobile links or add code to the existing links partial if any adjustments are needed?
- I am having a little formatting issue with the drop-down vs the margin of the existing page. I am new to Turbo and not sure if there is something I am missing in the formatting. A quick fix is adding an id to the application template but I can work on a better solution. 
- Testing has all been manually done.  What is the protocol for testing? 

_A start for now! Please let me know what can be improved._

Photos below:

Screenshot 2023-08-03 at 17 57 03. 
  
Screenshot 2023-08-03 at 17 57 13. 

Screenshot 2023-08-03 at 17 57 25

### DIFF
--- a/app/javascript/controllers/navbar_controller.js
+++ b/app/javascript/controllers/navbar_controller.js
@@ -1,0 +1,26 @@
+document.addEventListener('turbo:load', function () {
+  const hamburger = document.getElementById('hamburger')
+  if (hamburger) {
+    hamburger.addEventListener('click', function () {
+      const mobileMenu = document.getElementById('mobile-menu')
+      const content = document.getElementById('main')
+      const openIcon = document.getElementById('menu-open-icon')
+      const closeIcon = document.getElementById('menu-close-icon')
+
+      if (
+        mobileMenu.style.display === 'none' ||
+        mobileMenu.style.display === ''
+      ) {
+        mobileMenu.style.display = 'block'
+        content.style.marginTop = '200px' // adjust the value as needed
+        openIcon.style.display = 'none'
+        closeIcon.style.display = 'block'
+      } else {
+        mobileMenu.style.display = 'none'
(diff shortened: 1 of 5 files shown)


app/javascript/controllers/navbar_controller.js
```diff
@@ -0,0 +1,26 @@
+document.addEventListener('turbo:load', function () {
```
yshmarov, line 1: at this point this is just vanilla js stored in a folder for stimulus controllers.

why don't you use an actual stimulus controller? :)

awalsh3: Great point! I am new to stimulus but I _should_ follow the convention. I will work on it sometime this week. Thanks!

adrienpoly: hello @awalsh3 thanks for taking this issue. 

I would like to suggest an alternative path at the same time with less JS or maybe none. the details/summary html tags can be used to create dropdowns. This is What Basecamp uses in Hey as an example 

Here is an article on that subject but they are also several others available :

https://dev.to/ayushn21/details-summary-tags-html-s-best-kept-secret-4ak4

adrienpoly: When I mentioned Basecamp/Hey it was more for the technical part not really the way it looks. Your personal site looks clean I am sure you can make a good judgement on what looks best. I don't have a very strong option on this.

PR #49, opened 2023-08-05
remove webdrivers gem

### LINKED ISSUE #48: Remove webdrivers gem
Rubyvideo dev is built using Rails 7.1 alpha
Rails 7.1 will ship without webdrivers gem 
https://github.com/rails/rails/pull/48847

Let's remove that gem

haiderashfaq:
Resolves #48


adrienpoly:
Nice ! couldn't by simpler 😄

PR #50, opened 2023-08-14
add a test to cover the Turbo stream broadcast with Litestack

adrienpoly:
Adds a simple test that broadcast the `_about` partial of a speaker to ensure TurboStream broadcast is working with Litestack

## Issue

The current test works when I use the normal cable `test` adapter but fails when I use the `litestack` adapter. The cable.test.db is created but the channel never connects. It would be a great benefit of Litestack to be about to have the exact same configuration for test and production for Action cable infrastructure.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,10 +10,10 @@ gem "rails", "~> 7.1.0"
 gem "propshaft"
 
 # Use sqlite3 as the database for Active Record
-gem "sqlite3", "~> 1.4"
+# gem "sqlite3", "~> 1.4"
 
 # Use the Puma web server [https://github.com/puma/puma]
-gem "puma", ">= 5.0"
+gem "puma"
 
 # use jbuilder for the api
 gem "jbuilder"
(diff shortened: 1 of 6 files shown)


palkan:
Hey @adrienpoly,

Got across this issue while browsing litestack repo 🙂

On of the options is to switch to a litestack adapter only in system tests (and keep using `test` for unit tests, since it's required for them). See example here: https://docs.anycable.io/v1-2/rails/getting_started?id=development-and-test

palkan:
Oh, I see.

> expected to find css "turbo-cable-stream-source[connected]

It's related to https://github.com/oldmoe/litestack/pull/41

The `connected` callback never called and, thus, the `connected` attribute is not set: https://github.com/hotwired/turbo-rails/blob/097d8f90cf0c5ed24ac6b1a49cead73d49fa8ab5/app/javascript/turbo/cable_stream_source_element.js#L10

Surprisingly, it works in production, because the actual subscription is created at the pub/sub level, but the client never receives the confirmation (but receives actual updates).

PR #64, opened 2023-10-23
Recommended talks

adrienpoly:
This PR is preparation work to isolate the recommended talks into a dedicated controller.

The logic that is being built in #19 can be then added to the Talks#related_talks method

### DIFF
--- a/app/controllers/talks/recommendations_controller.rb
+++ b/app/controllers/talks/recommendations_controller.rb
@@ -1,0 +1,14 @@
+class Talks::RecommendationsController < ApplicationController
+  skip_before_action :authenticate_user!
+  before_action :set_talk, only: %i[index]
+
+  def index
+    @talks = @talk.related_talks
+  end
+
+  private
+
+  def set_talk
+    @talk = Talk.find_by(slug: params[:talk_slug])
+  end
+end
--- a/app/controllers/talks/recommendations_controller.rb
+++ b/app/controllers/talks/recommendations_controller.rb
@@ -3,6 +3,7 @@ class Talks::RecommendationsController < ApplicationController
   before_action :set_talk, only: %i[index]
 
   def index
+    redirect_to talk_path(@talk) unless turbo_frame_request?
     @talks = @talk.related_talks
   end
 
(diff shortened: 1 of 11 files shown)


app/models/talk.rb
```diff
@@ -95,4 +95,8 @@ def thumbnail_lg
   def thumbnail_xl
     self[:thumbnail_xl].presence || "https://i.ytimg.com/vi/#{video_id}/maxresdefault.jpg"
   end
+
+  def related_talks(limit: 6)
+    Talk.order("RANDOM()").excluding(self).limit(limit)
```
yshmarov, line 100: this is not really "related" 😉

a youtube video has "tags". 

if you find talks with the highest amount of matching tags, that would really make the results more "related"

app/views/talks/recommendations/index.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= turbo_frame_tag "recommended_talks" do %>
+  <div data-turbo-temporary data-controller="transition" class="hidden" data-transition-enter-after-value="0">
```
yshmarov, line 2: TIL [`data-turbo-temporary`](https://turbo.hotwired.dev/handbook/building#preparing-the-page-to-be-cached) 😯

app/models/talk.rb
adrienpoly: Yes for sure this is preparation work for #19, the vector base search should give us more relevant recommendations once we got it in place. 

The main goal of this PR was to split in a frame this list of recommended talks so that the talks#show is less coupled to the recommendation engine. 

As for the youtube video tags unfortunately I don't think lots of the channels I scrapped had some defined. RailsWorlds has some in the description that would need to be parsed. 
I want to have down the road some tags for the talks so that we can have a much better facet search experience. I thought they could either be crowdsourced and user-generated our check if I could generate some using OpenAI and the video transcript

app/controllers/talks/recommendations_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Talks::RecommendationsController < ApplicationController
+  skip_before_action :authenticate_user!
+  before_action :set_talk, only: %i[index]
+
+  def index
+    redirect_to talk_path(@talk) unless turbo_frame_request?
```
yshmarov, line 6: love to see this `unless turbo_frame_request?`

PR #86, opened 2024-07-18
Add Helvetic Ruby videos

marcoroth:
Add videos from talks at Helvetic Ruby from both editions using the script from #60.


db/seeds.rb
```diff
@@ -32,6 +33,8 @@
 
       event = Event.create!(name: event_data["title"], date: event_data["published_at"], organisation: organisation)
 
+      event.update!(slug: event_data["slug"])
+
```
marcoroth, line 37: The slug is needed here so the slugs from the `data/organisations.yml` and the `data/helveticruby/playlists.yml` files are actually respected.

adrienpoly:
great thank merging it will seed everything locally to test a deploy it after

PR #88, opened 2024-07-21
Adds Helsinki Ruby Brigade

matiaskorhonen:
Adds videos from the [Helsinki Ruby Brigade](https://rubybrigade.fi) meet-ups.

* All the videos are listed in one yaml file as we typically only have one or two talks per event so it doesn't seem to make sense to split them up into separate playlists
* Needs the slug fix from the Helvetic Ruby PR
    * #86 

Let me know if you'd like any changes…


adrienpoly:
great thanks for the addition

PR #90, opened 2024-07-31
Add Baltic Ruby 2024 videos

marcoroth:
CleanShot 2024-07-31 at 18 16 32@2x

### DIFF
--- a/app/models/youtube/video_metadata_baltic_ruby_2024.rb
+++ b/app/models/youtube/video_metadata_baltic_ruby_2024.rb
@@ -1,0 +1,75 @@
+module Youtube
+  class VideoMetadataBalticRuby2024
+    SPEAKERS_SECTION_SEPARATOR = " - "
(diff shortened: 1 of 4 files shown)


adrienpoly:
great ! Those cover are awesome 🤩

I ll wait 2 weeks before publishing them unless one of the organiser comes here  and ask for an immediate publication.  
I had this request from another event organiser. We link to the YouTube Video and play it from there so each view does count as real view count for Youtube, but they were some doubt that it could reduce the reach within the Youtube algorithm. Not a Youtube SEO expert so agreed to this organiser request. 

FYI on a side note, I added recently ruby conf 2022 and 23. What I ended up doing was taking the raw output and giving it to chat GPT to clean the metadata. It work very well. I think I'll update the instructions like that rather than having to build a dedicated metadata parser for each conference 
With mini it costs pennies

adrienpoly:
live

PR #92, opened 2024-08-01
Link to speaker page on the Home page

marcoroth:
This pull request makes the speaker avatars clickable on the home page.


adrienpoly:
great idea

PR #107, opened 2024-08-08
Add RubyKaigi videos

marcoroth:
CleanShot 2024-08-08 at 12 38 27


adrienpoly:
courious to see what the transcript and summary outcome looks like for the japanese talks

marcoroth:
Ah right! FWIW, I added a `language` field for all these talks so we could indicate that a talk is non-english somehow in the UI

adrienpoly:
Live

will need to find a better way to retrieve the transcript. I am currently using a hack. For other talk I could get a raw transcript for 95%. For Kaigi it is like 50% max

I would like to avoid going to Wisper speech to text solution to keep it simple. 

Someone suggested to use YouTube-dl to get them

PR #113, opened 2024-08-09
Add RubyConf TH videos

marcoroth:
CleanShot 2024-08-09 at 20 06 08

### DIFF
--- a/data/organisations.yml
+++ b/data/organisations.yml
@@ -123,3 +123,14 @@
   slug: euruko
   language: english
   youtube_channel_id: UCgXsZr2e0AgUJMZm5t03WVw
+- name: RubyConf TH
+  website: https://rubyconfth.com
+  twitter: rubyconfth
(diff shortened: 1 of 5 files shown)


adrienpoly:
Could you update the published_at date with the first day of the conference. This is what I am using for sorting the talks by dates. YouTube published date is only an approximation. Good enough if we don’t have the real date

marcoroth:
Updated, maybe we should rename that field then.  I understood `published_at` as "the date the video was published on", but something like `date` would probably be more accurate. 

I updated the dates to be exact to the respective day of the conference.

adrienpoly:
Great yeah I agree it is confusing 

maybe the date field could be optional if we have it we take it if not then we fallback to the published_at. This way we don’t have to update all existing files

marcoroth:
I'm happy to backfill them if we agree on something 🙌🏼

PR #118, opened 2024-08-12
Update talk from metadata tools

adrienpoly:
adds an admin filter to select talks by event
adds a new admin action to update talk metadata from the yml file


app/models/talk.rb
```diff
@@ -133,4 +136,18 @@ def related_talks(limit: 6)
   def transcript
     enhanced_transcript.presence || raw_transcript
   end
+
+  def update_from_yml_metadata
+    self.title = yml_metadata["title"]
+    self.description = yml_metadata["description"]
+    self.date = yml_metadata["published_at"]
```
marcoroth, line 143: You probbaly also want to double check the speakers here. I noticed that a bunch of speakers had some typos or other small details differently that could cause a mismatch

adrienpoly: Good call. I will have a look just to make sure we don’t create duplicate speakers when we correct a typo

marcoroth: Opened #120 which might help with this

adrienpoly:
@marcoroth updated this PR with the Static::Video api

much nicer

marcoroth:
Nice, sweet!

app/models/talk.rb
```diff
@@ -143,4 +147,15 @@ def related_talks(limit: 6)
   def transcript
     enhanced_transcript.presence || raw_transcript
   end
+
+  def update_from_yml_metadata
```
marcoroth, line 151: @adrienpoly should this be `update_from_yml_metadata!` or does the `performs` on L58 somehow figure that out by itself?

adrienpoly: indeed there is something wrong here will fix it on main thanks for the catch

adrienpoly: fixed https://github.com/adrienpoly/rubyvideo/commit/a1f1d5a8078e5a308425cfd31f9455ba63f05357

PR #127, opened 2024-08-15
Prevent duplicates

adrienpoly:
Ensure we don’t have duplicated topics for a talk


marcoroth:
Good point! Also feel free to remove this duplicate 😅 

https://github.com/adrienpoly/rubyvideo/blob/815025650b83f884f11c3b3d340f1e7dacb3b436/db/seeds.rb#L325-L326

PR #129, opened 2024-08-16
Status enum for topics

adrienpoly:
@marcoroth I am replacing the publish boolean by an enum so that we can reject topics once for all 

I ll see if it make sens to feed the list of rejected topics to the prompt


marcoroth:
Even better, sounds good to me 👍

PR #140, opened 2024-08-18
Add `language` column to `Talk`

marcoroth:
This pull request adds a `language` field to the `Talk` model, so we can populate it from the video metadata and persist it in the database.

We normalize the language using the `iso-639` gem and only persist the ISO-639 alpha2 code in the database.


adrienpoly:
great addition

marcoroth:
I added the Avo filter anyway, but reported the issue here https://github.com/avo-hq/avo/issues/3151

PR #150, opened 2024-08-20
default to en when language is nil or invalid

adrienpoly:
@marcoroth I think there where some kind conflict between the validation and normalize. Also added the `apply_to_nil: true` setting for the normalize function. as a result if the language is nil or unknow it always defaults to `en`

### DIFF
--- a/app/models/talk.rb
+++ b/app/models/talk.rb
@@ -60,7 +60,7 @@ class Talk < ApplicationRecord
 
   # normalization
   normalizes :language, apply_to_nil: true, with: ->(language) do
-    Language.find(language)&.alpha2 || Language::DEFAULT
+    language.present? ? Language.find(language)&.alpha2 : Language::DEFAULT
   end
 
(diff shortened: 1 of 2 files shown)


marcoroth:
Oh nice, I think `apply_to_nil` was the missing piece I was looking for!

Though, I think we should only default to `en` if it's nil, but not when it's unknown, in that case it should fail because you passed it an invalid value. So I think this would be "more" correct:

```diff
-   # attributes
-   attribute :language, default: Language::DEFAULT

   # normalization
-   normalizes :language, with: ->(language) {
+   normalizes :language, apply_to_nil: true, with: ->(language) {
     language.present? ? Language.find(language)&.alpha2 : Language::DEFAULT
   }
```

adrienpoly:
> Though, I think we should only default to `en` if it's nil, but not when it's unknown, in that case it should fail because you passed it an invalid value. 

Makes sens

PR #111, opened 2024-08-09
RubyConf data cleanup + RubyConf 2022 videos

marcoroth:
* Add missing talks for RubyConf Mini 2022
* Add talks from RubyConf 2022
* Added missing speakers
* Fix Speaker names
* Fix wrong/missing/incomplete talk titles
* Formatting

PR #151, opened 2024-08-20
Make sure all talks get seeded

marcoroth:
We had a few reasons why not all talks would be seeded.

* We lost some of the organizations in https://github.com/adrienpoly/rubyvideo/pull/108, I added them back here.
* Not all talks have titles/valid speaker names (this is fixed in https://github.com/adrienpoly/rubyvideo/pull/111, #104, #103, and #102)
* the handling for when talks didn't have a language caused it to set it to `nil` instead of the `Language::DEFAULT`, this is fixed in this PR by only setting if it it has a language present.

I extended the db seed test to also make sure that all talks from the static files get seeded.

Additionally, I updated the seed logic so we can re-seed and update the talks by doing so. I also removed a duplicate talk and rewrote part of the slug generation so we don't generate the same slug if two talks have the same title.

~~Some of this might conflict with https://github.com/adrienpoly/rubyvideo/pull/150.~~


marcoroth:
We are only missing the speakers for some lightning talks videos and the missing changes from the other PRs

adrienpoly:
ok perfect will try to review this promptly

marcoroth:
Okay cool. This branch now fixes all remaining missing information. With the changes from #102, #103, #104, #111 and #116 we should then be able to seed all existing talks! 🙌🏼

adrienpoly:
@marcoroth ok I think I merge all related PRs if you could rebase from main and let me know if there is anything I should do when merging

marcoroth:
okay cool, I added the missing info and fixed the tests. This should be ready to go now!

adrienpoly:
Looks great will run a few additional test tonight and merge it

marcoroth:
Awesome, sounds good! 🙌🏼

adrienpoly:
Thanks for this amazing work

PR #156, opened 2024-08-22
limit to 21 talks

### LINKED ISSUE #155: Show 21 talks per page
Currently we only show 20 talks per page leaving off one spot at the end of the list. I think it would make sense if we would show 21 talks per page so we can fill out all the 7 rows.

CleanShot 2024-08-21 at 01 37 02

adrienpoly:
close #155


marcoroth:
Why did it return 20 items if you had a 9 items in the call?

adrienpoly:
> Why did it return 20 items if you had a 9 items in the call?

I think the pagy api changed. The items: 9 was doing nothing and probably the default is 20

PR #160, opened 2024-08-23
Try out Oaken for seeds and test data

kaspth:
Paired with @marcoroth


adrienpoly:
thanks @kaspth and @marcoroth for this PR.
Oaken was something I wanted to try out but never really got the time for it. I like how this looks.

While chatting quickly with @marcoroth I understand that the production seed is the issue that remains to be solved. 
up to now I was using the seed task to update new content in production. I know it is not really a task that we are due to run in production but so far in our particular case it worked quite well. This being said I wouldn't be opposed to having a dedicated task for ingesting data in prod (and dev for setup) if this could make it easier 


PS sorry lots of conflit to solve. They where a large backlog of PR waiting to be merge....

marcoroth:
I've been wondering if we even should fixtures/test data in the first place. Would it make sense to just use the actual data in `data/` for the test and write the tests with that in mind?

kaspth:
heyo, sorry for being out of the loop on this one!

> I've been wondering if we even should use fixtures/test data in the first place. Would it make sense to just use the actual data in data/ for the tests and write the tests with that in mind?

@marcoroth this makes sense to me if that's easier.

> up to now I was using the seed task to update new content in production. I know it is not really a task that we are due to run in production but so far in our particular case it worked quite well.

@adrienpoly all good, I think having production specific seeds is going to be more common in the future — Oaken is trying to suggest this to people too. I'd just recommend running `Rails.application.load_seed` directly, which is what the rake `db:seed` task calls.

marcoroth:
I'm going to close this one. As we realized, it doesn't really make sense to have seeds in general in RubyEvents, as our "seeds" are the YAML files we have in `data/`. Thanks for pairing on this @kaspth! 🙏🏼

kaspth:
All good! I think relying on the data in data/ sounds right.

Can't believe it's already been almost two years since we looked at this 😅

PR #163, opened 2024-08-24
Add speaker tab to talk show page

marcoroth:
CleanShot 2024-10-22 at 20 11 21


adrienpoly:
while it makes sense for talks having lots of speaker, this tab doesn't really add any value for most talks as the majority of the talks have only one speaker.

I wonder if displaying the speakers bellow the title as we do on the card wouldn't be enough

CleanShot 2024-08-31 at 11 21 56@2x

marcoroth:
Yeah, maybe we could do it inline if there is one speaker or otherwise add the speaker tab

marcoroth:
I think it still makes sense to have the speakers tab for the case where there are a lot of speakers. But now I made it so that clicking on the "See all speakers" reveals the speakers tab with all the speakers.

CleanShot 2024-11-01 at 00 50 06

PR #167, opened 2024-08-26
Add Brighton Ruby 2023 videos

marcoroth:
~~This PR is based on #266.~~

| Organisation Show | Event Show |
| --- | --- |
| CleanShot 2024-10-30 at 23 40 07 | CleanShot 2024-10-30 at 23 38 55 |


adrienpoly:
Thanks for this great addition. 

I think we should ask @andycroll permission first. Organizers who are publishing their videos on Youbute defacto accept that their video can be embedded into another website as it is a native feature of Youtube. Also videos viewed from embedded players do count for the view stats on Youtube. 

For orgs having made the choice to host by themself the videos, I think we should get their explicit consent first. 

One option could be that we just pull the metadata + Thumbnail and link to the Org website for the player (to be decided by the org)

adrienpoly:
> Good point, I'm more than happy to extract the video provider to its own pull request, since it's technically two "features" in one pull request.
> 
> I was just testing the mp4 "provider" with the lastest Brighton Ruby edition, since I knew they were using their own video hosting.

I will send an email to Andy. Don’t change anything for now

h-lame:
As someone who hosted a conference series (http://rubymanor.org) and meetup (https://lrug.org) that both used mostly self-hosted content I'm very keen to see the mp4 provider go live, but interested in what you would need from us in terms of permissions to surface it?

I _think_ that we'd be ok with embeds if you have a nice player, but linking direct to the videos instead of embedding would also work.

marcoroth:
I think the whole premise of RubyVideo is the watch the talk as an embedded video on the RubyVideo site. So you get all the annotations and relevant context of where this talk took place. If you just link to the native MP4 browser player from the thumbnail on RubyVideo you loose a lot of the magic.

I'm going to extract the mp4 provider so we can use it to also embed other conferences.

PR #178, opened 2024-09-09
watch list

adrienpoly:
https://github.com/user-attachments/assets/340c29cb-4f31-4ea2-9ff0-e88eaf152424


marcoroth:
Nice, I really like this!

Small nudge, I feel like the heart icon is usually used for "Favorite" or similar, whereas for a watch list I have expected some like a bookmark or clock-like icon.

adrienpoly:
CleanShot 2024-09-09 at 17 54 44@2x

Bookmark style icon

adrienpoly:
merging it like that next part will be to manager user auth

PR #191, opened 2024-09-10
add github oauth

adrienpoly:
Sign in / up with Github

CleanShot 2024-09-10 at 02 44 03@2x


app/views/shared/_navbar.html.erb
```diff
@@ -8,18 +8,28 @@
...
             <%= icon "github", size: :lg %>
           <% end %>
         </li>
       </ul>
-      <%= render "shared/user_dropdown" %>
+      <% signed_in do %>
+        <%= render "shared/user_dropdown" %>
+      <% end %>
+      <% signed_out do %>
+        <% if false %>
```
adrienpoly, line 29: poor man Feature flag

PR #213, opened 2024-09-24
Add basic empty state for events index page

davidalejandroaguilar:
# Description
    
This PR adds an empty state to the events index page, so that when there are no events, instead of displaying a blank page, we display a message to the user.

For now, I'm just adding a basic `<p>` tag, and using no punctuation at the end. However, let me know if you have something else in mind.

## Why

I was clicking around the webpage to test performance ([like you guys asked](https://x.com/adrienpoly/status/1838235678859022747)), and when I saw the blank page I thought I had indeed broken something up. Turns out, there's just no events for some letters.

In order to not confuse users as to whether the page is still loading, an empty state message would help.

# Media

https://github.com/user-attachments/assets/8d03fd28-2406-47c5-8787-5e2045f337f8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,6 @@ doc/*
 
 service_account.json
 /.cache
+
+# asdf
+.tool-versions
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -535,6 +535,7 @@ GEM
 PLATFORMS
   aarch64-linux
   arm64-darwin-23
+  arm64-darwin-24
   x86_64-darwin
   x86_64-linux
 
--- a/app/views/events/index.html.erb
+++ b/app/views/events/index.html.erb
@@ -11,10 +11,14 @@
       <% end %>
   </div>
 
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -73,3 +73,6 @@ doc/*
 
 service_account.json
 /.cache
+
+# asdf
+.tool-versions
```
davidalejandroaguilar, line 78: For those using asdf like myself, and setting `asdf local ruby 3.3.5` sets a .tool-versions file. I'm adding it to the `.gitignore` file. Let me know if you think otherwise.

Gemfile.lock
```diff
@@ -535,6 +535,7 @@ GEM
 PLATFORMS
   aarch64-linux
   arm64-darwin-23
+  arm64-darwin-24
   x86_64-darwin
   x86_64-linux
```
davidalejandroaguilar, line 540: Automatically added, this should be fine.

adrienpoly:
thanks

PR #214, opened 2024-09-25
update setup with docker compose

adrienpoly:
I had several comments that getting started with Rubyvideo.dev was a bit challenging especially with the setup for Meilisearch.

This PR tries to simplify that with a docker-compose conf and an updated readme


jmarsh24:
Nice Job! I tested this PR and was able to get the codebase up and running!

PR #216, opened 2024-10-03
Explore all talks recorded at event card

### LINKED ISSUE #180: Show a "Explore all talks recorded at #{event.name}" on talk show page
We don't have to implement the whole design, but I think it's a nice idea to have a "Explore all talks recorded at #{event.name}" on the talk show page so you can easily explore which other talks happened at a conference.

CleanShot 2024-09-09 at 12 33 41

jmarsh24:
fixes #180

I left out the event city and country because the current seed data doesn't contain this information. I have added a link to the event at the bottom of every **talk#show** view and have styled it responsively to match the design in the issue.

There is also not an event image so I just included the talk thumbnail. (This should probably be fixed)

https://github.com/user-attachments/assets/ef746f3a-4124-4600-bf30-169726690f3f


app/views/talks/_talk.html.erb
```diff
@@ -69,3 +69,32 @@
...
+<%= link_to event_path(talk.event), class: "card bg-gray-200 w-full xl:flex xl:flex-row xl:px-6 xl:py-6 xl:gap-4 overflow-hidden" do %>
+  <div class="flex aspect-video xl:w-1/5 overflow-hidden">
+    <% event_talk = talk.event.talks.where.not(id: talk.id).sample || talk.event.talks.take %>
+    <%= image_tag event_talk.thumbnail_sm,
+          srcset: ["#{event_talk.thumbnail_lg} 2x"],
+          id: dom_id(event_talk),
+          height: 174,
+          width: 310,
+          alt: "explore all talks recorded at #{talk.event.name}",
+          style: "view-transition-name: #{dom_id(event_talk, :image)}",
```
adrienpoly, line 84: this should be remove otherwise it breaks view-transition. view transition name must be uniq on the page

adrienpoly (commented):
I think for now we should remove all images until we have a proper Event image 

@marcoroth WDYT as this was your design suggestion ?

jmarsh24:
@marcoroth @adrienpoly I think if this were up to me I think that the header above the list of talks from the same event should be a link. Instead of "RailsConf 2024" it could be "Explore all talks from RailsConf 2024" and the hover over could be a link (an image can be added afterward).

marcoroth:
> I think for now we should remove all images until we have a proper Event image

I think I agree with this. If we just have the name of the event, maybe the description and the speaker avatars we should be good to start.

jmarsh24:
I think this combined with #226, is already a good starting point.

marcoroth:
Agreed, and with https://github.com/adrienpoly/rubyvideo/pull/163 we should have the full picture for each talk too.

marcoroth:
Opened #300 to add artwork. With that we should be able to render a nice looking card!

marcoroth:
@jmarsh24 I brought back the card, which we can enhance with #300 once it's merged. But let's merge this in the meantime 🙌🏼 

Thank you!

PR #217, opened 2024-10-03
Fix Talk#Show speakers n+1

jmarsh24:
There is an n+1 query triggered by rendering the speakers of each talk because it's not eager loaded in the controller when rendering Speakers#show

| **Before** | **After** |
|------------|-----------|
| Before | After |


adrienpoly:
Thanks but I think the PR is missing the commit

jmarsh24:
@adrienpoly I added the missing commit 😇

jmarsh24:
@adrienpoly I think this one is also ready to merge with approval.

PR #218, opened 2024-10-03
Optimize talks/card partial render

jmarsh24:
| **Before** | **After** |
|------------|-----------|
| Before | After |


adrienpoly:
Thanks if you can update the image on the PR as they are broken
Also if you can fix the syntax, talk is not anymore available so should be removed from the locals and probably add a `as: talk` to be explicit for what the partial will receive

jmarsh24:
I forgot to push one last commit on this one.

PR #224, opened 2024-10-16
Fix sign out links

jmarsh24:
**Before**
`button_to` creates a form which causing an issue that the user can only logout if they click directly on the text. 

https://github.com/user-attachments/assets/4629bd20-444c-4b10-9c64-592e306914e1

**After**
Using `link_to` with `data-turbo-method: :delete` fixes this issue.

https://github.com/user-attachments/assets/e5b01f91-e2cd-4a3b-999d-4d69e8a7eb75

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,8 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 20
-          cache: yarn
+          cache: "yarn"
+          cache-dependency-path: "/dev/null"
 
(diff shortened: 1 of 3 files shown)


adrienpoly:
that is a good point I will suggest an alternative PR that preserve the button_to but fixes it with CSS. If it works I would prefere that we keep the ability to use button to in dropdowns

adrienpoly:
@jmarsh24 here is an alternative solution https://github.com/adrienpoly/rubyvideo/pull/286

PR #228, opened 2024-10-17
Add "Summarized using AI" badge

marcoroth:
I think it's important that AI generated content is also labelled as such, especially nowadays - which is why this pull request adds a "Summarized using AI" badge to the talk summary on the show page.

CleanShot 2024-10-17 at 13 42 01@2x

It gives some more details on hover in the tooltip:

CleanShot 2024-10-22 at 03 59 21@2x

The speaker can update the talk to use the badge or not:

CleanShot 2024-10-22 at 03 57 38@2x


adrienpoly:
2 comments:

1. a verified speaker has the ability to update manage the talk summary 
If he does so there should be a way to remove this badge 

2. using the primary color here seems to me a bit overkill. I think it should be more subtle

marcoroth:
Oh good points! Thanks for pointing those out!

1) Oh right. I guess we could add a boolean flag to the talk so we know if the speaker updated the summary or not, or is that overkill too?

2) Maybe yeah, I only took it to match the theme and because the secondary color didn't seem to fit into the design.

adrienpoly:
I like on conf talk how this is mentioned in a subtle way 
CleanShot 2024-10-18 at 08 01 51@2x

What do you think about adding the sparkle icon to the tab title

marcoroth:
Yeah, nice! I like that too. That was actually my first idea, but I wasn't able to add the SVG to the tab title given how the tabs are being setup with CSS variables

adrienpoly:
ahhh yes I see what you mean. For now it uses the js-less tab version of DaisyUI. To achieve that it probably require to add some JS and uses the version of the Daisy tab without a radio button https://daisyui.com/components/tab/#lifted

h-lame:
FWIW - I think this is a good idea. 

But, I also noticed on [my rubyconf talk 2023](https://www.rubyvideo.dev/talks/re-interpreting-data) that the transcript, which I think has been pulled from YouTube directly rather than AI summarised as part of the ingest, is also clearly an AI summary rather than an actual transcript.  I've no idea if that's a general problem - but maybe worth flagging too?  With the same caveats about it being removable if someone edits the transcript later.

marcoroth:
@h-lame yeah the transcript is pulled directly from YouTube. I think @adrienpoly did something to "enhance" the transcript (see https://github.com/adrienpoly/rubyvideo/blob/main/app/models/talk/transcript_commands.rb#L24-L88) but I'm not sure if that could mess with the script itself somehow in the process.

marcoroth:
I updated this pull request by adding a `summarized_using_ai` column to the database. It's set to true by default, but if a speaker decides to update their summary they can uncheck the checkmark in the talks#edit process to remove the badge.

adrienpoly:
> FWIW - I think this is a good idea.
> 
> But, I also noticed on [my rubyconf talk 2023](https://www.rubyvideo.dev/talks/re-interpreting-data) that the transcript, which I think has been pulled from YouTube directly rather than AI summarised as part of the ingest, is also clearly an AI summary rather than an actual transcript. I've no idea if that's a general problem - but maybe worth flagging too? With the same caveats about it being removable if someone edits the transcript later.

The transcript that I am able to get from Youtube is really of a bad quality. No sentences/punctuations, lots of miss used words. I am running those raw transcript through an AI prompt to enhance them a bit. I have seen some timings errors in those enhanced transcript but in general I feel the outcome is much more readable than the source. 

### Example from Youtube on your talk
CleanShot 2024-10-22 at 07 22 46@2x

In a perfect world probably we should use either a dedicated transcript service or use Wisper to ccreate our own transcript but this requires resources (dev ops/time) that I wasn't able to alocate yet 

I agree that wa can add a badge too on the transcript

marcoroth:
Going to merge this, we can still update the styling if we don't like how this works! :raised

PR #229, opened 2024-10-17
Add Rails World 2024 videos

### LINKED ISSUE #221: Latest Rails World Talks
The latest Rails World 2024 just occurred and the front page of rubyvideo.dev doesn't have it. The design and latest talk algorithm should be updated to catch trending videos like this.

https://www.youtube.com/watch?v=olxoNDBp6Rg
https://www.youtube.com/watch?v=zPBbHu-BKpQ
https://www.youtube.com/watch?v=-cEn_83zRFw

image

marcoroth:
This pull request adds the videos for the Rails World 2024 talks. To be merged on or after Oct 30th.

CleanShot 2024-10-17 at 13 15 51@2x

Resolves #221


jmarsh24:
What do you think of the idea that talks can be searchable upon release, but when you click here, you would actually be redirected to youtube until 2 weeks after (October 30th in this case)

That way rubyvideo.dev can be very up to date but still not subtract from youtube SEO.

marcoroth:
Maybe yeah, but then again, that seems like a lot of manual effort to maintain and keep up with these dates. 

But I also don't fully understand the argument for not having them available right away. I don't see how this would negatively impact the performance of the video on YouTube, I would have guessed it improves the performance because it not only comes from "organic" traffic from YouTube itself.

jmarsh24:
There wouldn't be anything manual about it. We can make it available to view on rubyvideo.dev 2 weeks after the youtube_video upload_date. 

It can just a link to the youtube video up until that point after that it would be a link to the resource on rubyvideo.dev

I think having the latest content without all the extra recommendations youtube gives you would be very valuable to the user.

adrienpoly:
> But I also don't fully understand the argument for not having them available right away. I don't see how this would negatively impact the performance of the video on YouTube, I would have guessed it improves the performance because it not only comes from "organic" traffic from YouTube itself.

I agree and mentioned similar arguments. Yet the fondation wanted to make sure nothing would penalised the launch of the videos and rather than playing with some unknown YT SEO algorithm they preferred to limit external risks.

TBH I don't think the traffic of RV would make any change at least as it is now

I didn't wanted to re open the discussion but having just link in the first 15/30 days would probably be beneficial for the vids as it creates backlinks

PR #232, opened 2024-10-17
Link to youtube for first two weeks

jmarsh24:
I have created a helper method which creates an external link to youtube if the date is less than 2 weeks ago. This allows rubyvideo to have the latest videos without affect any sort of SEO issues on youtube.

https://github.com/user-attachments/assets/fd6a52aa-7566-46a3-8793-4524d9e7a60c


app/helpers/talks_helper.rb
```diff
@@ -1,2 +1,21 @@
 module TalksHelper
+  def link_to_talk(talk = nil, options = {}, html_options = {}, &block)
+    raise ArgumentError, "must pass an instance of talk to this helper method" unless talk.present? && talk.is_a?(Talk)
+
+    url, target = if talk.date > 2.weeks.ago
```
marcoroth, line 5: The problem with this is that this date of the talk is meant the date when the talk was held, not when it was released.

jmarsh24: ahhh I missed that. 🤔 I guess we would have to expand the data/inject model to track this information.

marcoroth: Yeah sorry, I guess that's what I was alluding to in https://github.com/adrienpoly/rubyvideo/pull/229#issuecomment-2419259464

jmarsh24: We could get the information from youtube when users add more videos, and run a script to backfill the data model. 

If it will allow rubyvideo to have the latest videos, I think it's worth it.

adrienpoly:
My suggestion is to have a state on the talk model to either embbed the video (current behaviour) or display the thumbnail with a link to an external site. As an example for Brighton Ruby we decided to display the metadata and thumbnail but get a link to the original video. So this feature could be used for that use case too.

Then we can create in AVO a batch action to quickly update all videos from one state to the other. It doesn't have to be automatic at this point I think

jmarsh24:
Just to be sure I understand correctly I have created an outline summarizing your idea below.

**Work Summary**
- Add `published` boolean, default: :false to Talk
- Batch action in avo to update this published column
- add this arguement to the `link_to_talk` helper method. which feeds in the talk#published.

Is this what you are suggesting?

jmarsh24:
Here are some screenshots with the updated functionality. 

- I added an **embedded:boolean** on talk. By default talks will link to youtube until they are marked for embedding. 🛌🏼  
- I added two actions to embed and unembed the talks.
- I updated the link_to_talk helper method to use this boolean value.

https://github.com/user-attachments/assets/448a55c0-90b9-4718-9fee-8c95d5d80765

https://github.com/user-attachments/assets/04b1ce40-a291-465d-863e-4c4f0e974117

jmarsh24:
@marcoroth @adrienpoly I'm a bit lost on this one. The CI has failing tests but locally it's fine. I think it has something to do with Sqlite.

**Original Error**
image

I changed the migration number and then I got 
image

**Things I have tried**

- adding `bin/rails db:drop && bin/rails db:create && bin/rails db:migrate` to CI.
- Setting parallelized workers to 1 in CI
- Recreating a new migration
- Changing the migration number

All of these have failed and for some reason I get this missing attribute error. 😢

jmarsh24:
@adrienpoly Do you have any ideas on what might be causing this issue? 🤔

PR #234, opened 2024-10-17
Add GoRuCo videos

marcoroth:
CleanShot 2024-10-17 at 22 02 05@2x

CleanShot 2024-10-17 at 22 01 57@2x

CleanShot 2024-10-17 at 23 26 50@2x

### DIFF
--- a/data/goruco/goruco-2008/videos.yml
+++ b/data/goruco/goruco-2008/videos.yml
@@ -82,15 +82,15 @@
   raw_title: RejectConf / Lightning Talks by Various Presenters
   speakers:
     - Ryan Davis
-    - Johnson
-    - Delmont
+    # - Johnson # TODO: missing last name
+    # - Delmont # TODO: missing last name
     - Gregory Brown
     - Dan DiMaggio
     - Phil Matarese
     - Peter Jaros
     # - TODO # Mindstorms
     - Luke Miglia
-    - Wilson
+    # - Wilson # TODO: missing last name
   event_name: GoRuCo 2008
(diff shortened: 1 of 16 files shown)


marcoroth:
Sadly it seems like there are no recordings online from the 2007, 2009, and 2010 editions. 
Additionally the websites for 2012 and 2016 aren't up, so cannot determine the running order of the talks

data/goruco/goruco-2008/videos.yml
```diff
@@ -0,0 +1,100 @@
...
+    - Ryan Davis
+    - Johnson
+    - Delmont
+    - Gregory Brown
+    - Dan DiMaggio
+    - Phil Matarese
+    - Peter Jaros
+    # - TODO # Mindstorms
+    - Luke Miglia
+    - Wilson
```
adrienpoly, line 93: I understand that it is sometime complicated for lighting talk to get the full name of the speaker. 
When we just have the first name it means we will create a speaker with just the first name and this could create confusion 

what do you think to comment those speakers and add a comment `missing last name` maybe someone can fix that in the future?

adrienpoly (commented):
great I wasn't even aware of this conference
just a small suggestion for the lighting talk

data/goruco/goruco-2008/videos.yml
marcoroth: Yeah, that's fair 👍🏼

marcoroth:
Otherwise this should be ready as-is for the remaining years

PR #252, opened 2024-10-19
use basic search while fixing Meilisearch or dropping it

adrienpoly:
Having hard time with Meilsearch right now. I think most of it are dev ops related
This PR migrate to a very basic search by title 
I am also looking at Full text search like https://www.teloslabs.co/post/full-text-search-with-rails-and-sqlite

### DIFF
--- a/app/controllers/talks_controller.rb
+++ b/app/controllers/talks_controller.rb
@@ -7,8 +7,12 @@ class TalksController < ApplicationController
   # GET /talks
   def index
     if params[:q].present?
+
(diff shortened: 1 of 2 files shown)


jmarsh24:
What issue are you running into with meilisearch?

jmarsh24:
One thing to note with Meilisearch is that you need to make sure that there is enough RAM on the server. Sometimes it will fail silently unless you poke it with curl requests. This will usually show up by indexes failing.

PR #255, opened 2024-10-20
Stylize GitHub

marcoroth:
```diff
-Github
+GitHub
```


adrienpoly:
I am curious how does it works. 
I don't see any custom inflector in your PR

For a class named GitHub the filename should be git_hub.rb but here the filename remains github.

Is GitHub a default inflector in Rails?

adrienpoly:
Well there is an inflector not sur why I didn’t saw it when I first review the PR

PR #256, opened 2024-10-20
Add Language Emoji to talk card

marcoroth:
This pull request adds an emoji flag and language name on hover for non-english talks in the talk card.

https://github.com/user-attachments/assets/8085cbe7-1f47-48ed-ada1-4a76286b209d


app/views/talks/_card.html.erb
```diff
@@ -2,7 +2,14 @@
 <%# locals: (talk:, user_favorite_talks_ids: [], favoritable: false, back_to: nil, back_to_title: nil) -%>
 
 <div class="card card-compact bg-white shadow-xl w-full" id="<%= dom_id talk %>" style="<%= "view-transition-name: #{dom_id(talk, :talk_card)}" %>">
-  <%= link_to talk_path(talk, back_to: back_to, back_to_title: back_to_title), class: "flex aspect-video overflow-hidden" do %>
+  <%= link_to talk_path(talk, back_to: back_to, back_to_title: back_to_title), class: "flex aspect-video overflow-hidden relative" do %>
+    <% if talk.static_metadata.language.present? && talk.static_metadata.language.downcase != "english" %>
```
adrienpoly, line 6: Talk model has a langage attribute, any reason why we are not using it here?

marcoroth: hmm, there's was a reason why I didn't, but I'm blanking on why now. Let me change it 👍🏼

marcoroth: oh, I guess that's why...
CleanShot 2024-10-20 at 22 13 36

This doesn't make any sense

marcoroth: Updated 👍🏼

adrienpoly: I don't really like this with_essential_card_data but it was the only way to get decent perf for topic.talks very strange

PR #261, opened 2024-10-20
Add Speakerdeck column to Speaker

marcoroth:
This pull request adds a `speakerdeck` column to the `speakers` table. Maybe this allows us in the future to automate the matching of `slides_url` to decks on Speakerdeck.

| Profile Sidebar | Suggestion |
| --- | --- |
| CleanShot 2024-10-20 at 19 52 48 | CleanShot 2024-10-20 at 19 52 53 |

### DIFF
--- a/db/migrate/20241020174649_add_speakerdeck_to_speaker.rb
+++ b/db/migrate/20241020174649_add_speakerdeck_to_speaker.rb
@@ -1,0 +1,5 @@
+class AddSpeakerdeckToSpeaker < ActiveRecord::Migration[8.0]
+  def change
+    add_column :speakers, :speakerdeck, :string, default: "", null: false
+  end
+end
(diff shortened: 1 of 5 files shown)


db/migrate/20241020174649_add_speakerdeck_to_speaker.rb
```diff
@@ -0,0 +1,5 @@
+class AddSpeakerdeckToSpeaker < ActiveRecord::Migration[8.0]
+  def change
+    add_column :speakers, :speakerdeck, :string
```
adrienpoly, line 3: I know it it controversial but I prefer for most cases to default string to "" with null: false

adrienpoly:
great idea

db/migrate/20241020174649_add_speakerdeck_to_speaker.rb
marcoroth: 👍🏼

marcoroth:
Updated 🙌🏼

PR #263, opened 2024-10-20
Add `speakerdeck` to Avo Speaker resource

marcoroth:
Looks like I missed this in #261

### DIFF
--- a/app/avo/resources/speaker.rb
+++ b/app/avo/resources/speaker.rb
@@ -16,6 +16,7 @@ class Avo::Resources::Speaker < Avo::BaseResource
     field :name, as: :text, link_to_record: true, sortable: true
     field :twitter, as: :text
     field :github, as: :text
+    field :speakerdeck, as: :text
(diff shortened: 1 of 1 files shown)


app/avo/resources/speaker.rb
```diff
@@ -16,6 +16,7 @@ def fields
     field :name, as: :text, link_to_record: true, sortable: true
     field :twitter, as: :text
     field :github, as: :text
+    field :speakerdeck, as: :text
```
adrienpoly, line 19:
```suggestion
    field :speakerdeck, as: :text,  hide_on: :index
```

I think we should hide it from the index view

marcoroth: I was thinking about that and then thought: "oh we have all the other socials here too"

CleanShot 2024-10-20 at 23 24 56

adrienpoly: yeah it's just that now the actions (show/edit ...) are out of the viewport

will try it and update it if needed

PR #267, opened 2024-10-22
Add more speaker related charts in analytics

marcoroth:
CleanShot 2024-10-22 at 05 30 07@2x


adrienpoly:
Nice!


I am a bit suprise by the shape of the yearly speaking experience. Given that a majority of speakers have 1 talk I would have expected the first column to be higher, but I might be missing something

marcoroth:
> I am a bit suprise by the shape of the yearly speaking experience. Given that a majority of speakers have 1 talk I would have expected the first column to be higher, but I might be missing something

There's a good chance that the query is not quite right 😅

adrienpoly:
I propose that we comment this last query and merge it with the others so that we have time to work on it a bit more 

or maybe add a poor man feature flag and show it only in dev

PR #279, opened 2024-10-23
Add pronouns to `Speaker`

marcoroth:
This pull request adds the option for speakers to add their pronouns to their profile.

The value is set to `not_specified` by default and therefore doesn't show anything on the profile. If a speaker chooses the `dont_specify` option it will also not be shown.

Additionally the `custom` type allows to enter pronouns as a freeform field.

| Speakers#edit | Pronouns Select Options |
| ---- | ---- |
| CleanShot 2024-10-23 at 12 42 09 | CleanShot 2024-10-23 at 12 42 13 |

| Custom Pronouns | Speakers#show (in about) |
| --- | --- |
| CleanShot 2024-10-23 at 12 42 22 | CleanShot 2024-10-24 at 00 32 01 |

### DIFF
--- a/app/controllers/speakers_controller.rb
+++ b/app/controllers/speakers_controller.rb
@@ -60,8 +60,8 @@ class SpeakersController < ApplicationController
 
   def speaker_params
     {
-      anonymous: params.require(:speaker).permit(:github),
-      signed_in: params.require(:speaker).permit(:github),
+      anonymous: params.require(:speaker).permit(:github, :pronouns_type, :pronouns),
+      signed_in: params.require(:speaker).permit(:github, :pronouns_type, :pronouns),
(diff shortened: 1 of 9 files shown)


marcoroth:
Sadly the GitHub API doesn't expose the pronouns, so we can't fetch it through the `Speaker::EnhanceProfileJob` job.

adrienpoly:
WDYT of moving it up next to the name ?

something like that 
CleanShot 2024-10-23 at 17 52 54@2x

overall I would like to get rid of Github, Twitter, Website and just either list the content or replace those with an icon

marcoroth:
Nice! I like this suggestion, let me update the PR 👍🏼

PR #281, opened 2024-10-23
Add RubyNation 2017 videos

marcoroth:
CleanShot 2024-10-23 at 15 04 32

### DIFF
--- a/data/organisations.yml
+++ b/data/organisations.yml
@@ -420,3 +420,15 @@
   slug: ruby-on-ales
   language: english
   youtube_channel_id: UCWnPjmqvljcafA0z2U1fwKQ
+
+- name: RubyNation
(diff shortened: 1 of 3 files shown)


adrienpoly:
spent a bit of time investigating the test failure. It was I think a wrong format for published at

marcoroth:
Oh wow, I looked at this before and couln't figure out what's wrong. 

Looking at the diff I don't see it either now, but doesn't matter - but great it's working now! 🙌🏼

Thank you!

adrienpoly:
Yeah diff is more complex Than it should be. Tried a lot of things before finding 


They were one 
  published_at: "2017" with only the year

marcoroth:
ah I see, thanks! 🙌🏼

Issue #157, opened 2024-08-22
Rethink Lightning Talks


marcoroth:
The Lightning talks usually consist of multiple talks by different speakers about different topics. I think we should split up the lightning talks video into multiple individual "talks" on RubyVideo. A few reasons:

* Summaries are not gonna be accurate
* The Tags/Topics are not going to represent well what all the people talked about
* As a user you never see the lightning talk titles, so you don't know what you get.
* As a user you would never just click on a "Lightning Talks" talk on RubyVideo and say "let me watch all lightning talks"
 
But instead, if they were shown as individual talks you might end up clicking on one of them. Blue Ridge Ruby and Helvetic Ruby published individual videos for each of the lightning talks, but if an event doesn't do that, we could do it.

For the Euruko 2019 lightning talks we currently have something like:

```yaml
- title: Lightning Talks
  raw_title: Lightning Talks
  speakers:
    - Madison White # Super-powering your editor with Sorbet Typer
    - Miriam Tocino # Zerus & Ona
    - Julik Tarkhanov # NoSuchBucket
    - Benjamin Vetter # search_flip
    - Norma Miller # A "Splash" Course on Live Captioning
  event_name: Euruko 2019
  published_at: "2021-06-01"
  description: ""
  video_id: 2-1wlXkLt-I
```

But it would be better is if we had something like this instead:

```yaml
- title: Lightning Talks
  raw_title: Lightning Talks
  event_name: Euruko 2019
  published_at: "2021-06-01"
  description: ""
  video_id: 2-1wlXkLt-I
  talks:
    - title: Super-powering your editor with Sorbet Typer
      raw_title: Super-powering your editor with Sorbet Typer
      speakers: 
        - Madison White
      description: ""
      video_cue_start: 00:00
      video_cue_end: 05:00
      
    - title: Zerus & Ona
      raw_title: Zerus & Ona
      speakers: 
        - Miriam Tocino
      description: ""
      video_cue_start: 05:10
      video_cue_end: 10:00
      
    - title: NoSuchBucket
      raw_title: NoSuchBucket
      speakers: 
        - Julik Tarkhanov
      description: ""
      video_cue_start: 10:10
      video_cue_end: 15:00
      
    - title: search_flip
      raw_title: search_flip
      speakers: 
        - Benjamin Vetter
      description: ""
      video_cue_start: 15:10
      video_cue_end: 20:00
      
    - title: A "Splash" Course on Live Captioning
[...]

PR #287, opened 2024-10-23
Merge some speakers

h-lame:
I spotted some speaker issues with wrong attribution or duplicate versions, so this PR fixes a bunch.

TBH I just eyeballed the [speakers page](https://www.rubyvideo.dev/speakers) and spotted a few discrepancies.  I suspect there's a bunch more if you go digging (particularly with things like how people typically style their names vs longer or shorter versions).

Probably easiest to look commit by commit in the off chance you disagree with any particular changes (e.g. I have removed people from `speakers.yml` if I've merged them - maybe that's not right?).

### DIFF
--- a/data/euruko/euruko-2016/videos.yml
+++ b/data/euruko/euruko-2016/videos.yml
@@ -74,19 +74,29 @@
 - title: 'Lightning Talks: Day 1'
   raw_title: EuRuKo 2016 - Day 1 Lightning Talks
   speakers:
-    - Max Gorin # Making a Multiplayer Memoery Game with Elixir and Elm (0:29)
-    - Johanna Lang # Team Joda 2016 (Rails Girls Summer of Code 2016) (6:20)
-    - Dayana Mick # Team Joda 2016 (Rails Girls Summer of Code 2016) (6:20)
+    - Max Gorin # Making a Multiplayer Memory Game with Elixir and Elm (0:29)
+    - Johanna Lang # Team Joda 2016 (Rails Girls Summer of Code 2016) (6:20)
+    - Dayana Mick # Team Joda 2016 (Rails Girls Summer of Code 2016) (6:20)
     - Kinga Kalinowska # We Are Rubycats! (Rails Girls Summer of Code 2016) (12:11)
     - Izabela Komorek # We Are Rubycats! (Rails Girls Summer of Code 2016) (12:11)
-    - Melanie Keatley # Magic Solution to the Tech Gender Gap (18:22)
+    - Melanie Keatley # Magic Solution to the Tech Gender Gap (18:22)
     - Stephanie Nemeth # Magic Solution to the Tech Gender Gap (18:22)
     - Krzysztof (Christopher) Wawer # Validations (23:39)
     - Demir Zekić # The Importance of Teaching and Metoring (28:53)
     - Jan Lelis # Whirly - The Friendly Terminal Spinner (35:25)
   event_name: EuRuKo 2016
   published_at: '2016-09-23'
-  description: ''
+  description: |-
+    Lightning talks from Day 1 of EuRuKo
+
+    * "Making a Multiplayer Memory Game with Elixir and Elm" by Max Gorin (0:29)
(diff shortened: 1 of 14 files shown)


data/euruko/euruko-2016/videos.yml
```diff
@@ -74,19 +74,29 @@
...
+    - Max Gorin
+    - Johanna Lang
+    - Dayana Mick
+    - Kinga Kalinowska
+    - Izabela Komorek
+    - Melanie Keatley
+    - Stephanie Nemeth
+    - Krzysztof (Christopher) Wawer
+    - Demir Zekić
+    - Jan Lelis
```
marcoroth, line 86: Thanks for adding the lightning talks into the description!

Can we keep the comment here? For the case where we want to implement #157 it would be great to have all the lightning talks in the same comment format so we can parse/search&replace them in one go!

data/rocky-mountain-ruby/rocky-mountain-ruby-2012/videos.yml
```diff
@@ -65,7 +65,7 @@
     - Mike Gehard
     - Elain Marina
     - Jim Denton
-    - Tom Frei
+    - Thomas Frey
```
marcoroth, line 68: Nice catch!

data/goruco/goruco-2017/videos.yml
```diff
@@ -51,9 +51,9 @@
   video_id: 2C_IUMYzM7A
 
 - title: Developer Productivity Engineering
-  raw_title: 'GORUCO 2017: Developer Productivity Engineering by  Panayiotis Thomakos'
+  raw_title: 'GORUCO 2017: Developer Productivity Engineering by Pan Thomakos'
   speakers:
-    - Panayiotis Thomakos
+    - Pan Thomakos
```
marcoroth, line 56: We have the concept of "Canonical Speakers". I think we should setup one for "Panayiotis Thomakos" that sets the canonical to "Pan Thomakos" .

data/speakers.yml
```diff
@@ -5913,13 +5913,6 @@
   website: ""
   slug: james-duncan-davidson
   canonical_slug:
-- name: James Edward Gray
-  twitter: ""
-  github: JEG2
-  bio: ""
-  website: http://graysoftinc.com/
-  slug: james-edward-gray
-  canonical_slug:
```
marcoroth, line 5922: Let's keep this one too and set the `canonical_slug` to `james-edward-gray-ii` so future references point to the right record.

data/speakers.yml
```diff
@@ -10937,13 +10923,6 @@
   website: ablogaboutcode.com
   slug: pan-thomakos
   canonical_slug:
-- name: Panayiotis Thomakos
-  twitter: ""
-  github: ""
-  bio: ""
-  website: ""
-  slug: panayiotis-thomakos
-  canonical_slug:
```
marcoroth, line 10946: Let's keep this one too and set the `canonical_slug` to `pan-thomakos` so future references point to the right record.

data/speakers.yml
```diff
@@ -13915,20 +13885,6 @@
   website: http://tomdale.net
   slug: tom-dale
   canonical_slug:
-- name: Tom E Enebo
-  twitter: ""
-  github: ""
-  bio: ""
-  website: ""
-  slug: tom-e-enebo
-  canonical_slug:
```
marcoroth, line 13924: Let's keep this one too and set the `canonical_slug` to `thomas-e-enebo` so future references point to the right record.

marcoroth (commented):
Thanks @h-lame, very good detective work! 🕵🏼

h-lame:
Thanks for the feedback @marcoroth, I've rebased those changes into the original commits to keep the entries in speakers.yml where they are usuful redirect and to keep the comments in the Euruko lightning talks (can't think why I removed them in the first place when I left them in the day 2 talks 🤷 ).  While I was there I also added a few more merges, this time I think I have a better understanding of when to remove someone from speakers.yml (typos) and when to keep with a redirect (different styling of the name).

marcoroth (approved):
Awesome, thank you so much @h-lame! 🙏🏼

PR #299, opened 2024-10-25
Group talks on `speakers#show` page

marcoroth:
This pull request introduces the concept of grouped talks (similar to what was described in https://github.com/adrienpoly/rubyvideo/issues/236).

If we find grouped talks on a speaker profile we show a tab bar with "Grouped" and "Ungrouped" options, where the ungrouped option is the previous show page.

| Grouped | Ungrouped |
| --- | --- |
| CleanShot 2024-10-27 at 18 33 19 | CleanShot 2024-10-27 at 18 39 46 | 

if we don't find any grouped talks on a speakers profile we show just the list of talks as it was before:

| No Grouped Talks |
| --- | 
| CleanShot 2024-10-25 at 18 52 42 |

### DIFF
--- a/app/controllers/speakers_controller.rb
+++ b/app/controllers/speakers_controller.rb
@@ -26,8 +26,10 @@ class SpeakersController < ApplicationController
 
     @grouped_talks = @talks.group_by { |talk| talk.groupable_title.downcase }.map { |title, talks| [talks.first.groupable_title, talks] }.to_h
 
-    @ungrouped_talks = @grouped_talks.select { |title, talks| talks.count == 1 }.values.flatten
-    @grouped_talks = @grouped_talks.reject { |_title, talks| talks.count == 1 }.sort_by { |_title, talks| -talks.map(&:date).max.to_time.to_i }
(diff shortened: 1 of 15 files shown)


adrienpoly:
I like this idea to highlight grouped talks but I feel like the tab display is a bit overkilled and maybe not that obvious for the visitor.

I wonder if we could keep the talks into a single grid, reorder the talks so that grouped talks are adjacent and span them in the grid with some kind of border to show that they are similar talks

Or another alternative is to start introducing a sorting menu.
- most recent (default)
- least recent
- grouped

marcoroth:
I'm not sure I fellow. Are you just saying the UI for selecting the "view" is overkill or the functionally in general?

adrienpoly:
> I'm not sure I fellow. Are you just saying the UI for selecting the "view" is overkill or the functionally in general?

Yes having a tab for that might seems overkill to me. To me it is a kind of sorting so we should be able to add a date sorting order at some point to the UI. With this tab it become more complex I think

> I improved the grouping slighty comparer to what you see in the screenshots. It's grouping the non-grouped talks into their release year now too, which helps with navigating the talks.

I will try it live

marcoroth:
Updated the screenshots for Aaron's profile. You can click on them to view them in the full resolution.

PR #301, opened 2024-10-26
Add uses page content

silva96:
- Add uses page content
- Also, improved the readme by adding links to each project. 
 
I suggest you add affiliate links on Hetzner for example.

CleanShot 2024-10-26 at 10 37 31@2x

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -59,7 +59,7 @@ MeiliSearch::Rails.deactivate! do
 end
 
 # reindex all talk in MeiliSearch
-Talk.reindex! unless Rails.env.test? || Rails.env.production?
+# Talk.reindex! unless Rails.env.test? || Rails.env.production?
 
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -59,7 +59,7 @@
 end
 
 # reindex all talk in MeiliSearch
-Talk.reindex! unless Rails.env.test? || Rails.env.production?
+#Talk.reindex! unless Rails.env.test? || Rails.env.production?
```
silva96, line 62: When I setup the project, this failed because of the changes of 0b929c72

adrienpoly: good catch, we are i the middle of upgrading Meilisearch and Kamal 2 this is why we have a fallback to SQLITE search for now. Also for new comers it is simpler if the porject can run without a Meilisearch dependency.

PR #305, opened 2024-10-27
remove todo speaker

adrienpoly:


marcoroth (commented):
I think the reason why I added them not as a comment is because we need at least one speaker, otherwise the seeds wouldn't run.

PR #324, opened 2024-10-29
Add conference speakers page

### LINKED ISSUE #183: Conference Speaker List
This might be more relevant before/during a conference that it is after a conference, but I think there's some value in exploring all the speakers at a conference. 

Conference Show Speakers@2x

marcoroth:
This pull request implements a conference speakers page and enables the tab bar on the conference show page.

Resolves #183

CleanShot 2024-11-26 at 10 07 01


adrienpoly:
this is looking good to me, I ll try it live. 

I realize that we should add an attribute to the talk model to identify keynotes. This way we can have a real db query rather than iterating in ruby over all talks.

adrienpoly:
is this scroll up behaviour made on purpose? 

https://github.com/user-attachments/assets/e5a13e69-588c-46ff-8aba-c3d9e463849a

I find it a bit strange and would assume that navigating the tabs would preserve the scroll

marcoroth:
Nope, but I guess it's the first line I copied over in the view file.

marcoroth:
Solved it by wrapping the content into a turbo frame

marcoroth:
in the current state it's somehow breaking the frame content when navigating onto the events show page

marcoroth:
This now uses the `kind` column introduced in #334 to figure out the different types of speakers.

PR #326, opened 2024-10-29
enforce remote modals 

### LINKED ISSUE #315: Enforce modals for modal endpoints
A user has the possibility to navigate to the users#new endpoint and also open the modal which leads to a duplicate modal and page navigation. As a user, I would expect to only be able to view the session#new within the context of a modal.

CleanShot 2024-10-29 at 10 14 04@2x

adrienpoly:
close #315

adds a new concerns that works in pair with the remove modal concern
we can now specify that remove modals can only be viewed as remote modal by forcing certain route to be displayed only in turbo frames

```ruby
include RemoteModal
allowed_remote_modal_actions :new
force_frame_response only: %i[new]
```

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,6 +1,7 @@
 class SessionsController < ApplicationController
   include RemoteModal
   allowed_remote_modal_actions :new
+  force_frame_response only: %i[new]
 
   skip_before_action :authenticate_user!, only: %i[new create]
 
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,7 +1,6 @@
 class SessionsController < ApplicationController
   include RemoteModal
-  allowed_remote_modal_actions :new
-  force_frame_response only: %i[new]
+  respond_with_remote_modal only: [:new]
 
   skip_before_action :authenticate_user!, only: %i[new create]
 
--- a/test/controllers/sessions_controller_test.rb
+++ b/test/controllers/sessions_controller_test.rb
@@ -12,9 +12,16 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
     assert_response :success
   end
 
-  test "should get new" do
+  test "should get new in a remote modal" do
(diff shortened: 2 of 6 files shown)


test/controllers/sessions_controller_test.rb
```diff
@@ -13,7 +13,7 @@ class SessionsControllerTest < ActionDispatch::IntegrationTest
   end
 
   test "should get new" do
-    get sign_in_url
+    get sign_in_url, headers: {"Turbo-Frame" => "modal"}
```
jmarsh24, line 16: I think this test is no longer really doing what it says. This redirects to the root_path now which yields a success, but it's not the sessions#new page.

Is there a way to open the sessions#new modal after redirect to the home page?

app/controllers/sessions_controller.rb
```diff
@@ -1,6 +1,7 @@
 class SessionsController < ApplicationController
   include RemoteModal
   allowed_remote_modal_actions :new
+  force_frame_response only: %i[new]
```
jmarsh24, line 4: Why do the allowed_remote_modal_actions and force_frame_response behave different? One can pass keyword arguments directly and the other needs an array?

test/controllers/sessions_controller_test.rb
adrienpoly: When setting the headers with a turbo frame modal we do render the #new view. Maybe this test should go a bit further and ensure that we render this view. Not just an ok

app/controllers/sessions_controller.rb
adrienpoly: Yes this is a good point I can update the remote modal concern API

test/controllers/sessions_controller_test.rb
adrienpoly: I have updated the test suite to cover both cases

app/controllers/sessions_controller.rb
adrienpoly: I have refined a bit the remote modal api. I think it is now clearer 

now we only need to add

```ruby 
 respond_with_remote_modal only: [:new]
```

PR #329, opened 2024-10-30
fix: docker call in setup script

maciej-ka:
Warm greetings,

What used to be `docker-compose` is now `docker compose`.
Docker introduced that [change](https://docs.docker.com/compose/releases/migrate/) in 2020.

PR modifies setup script.
1. try to use `docker compose`
1. not found? try `docker-compose`
1. still not found? terminate with an unhandled error 🧨

In case `docker compose` is found but runs with error script will not attempt `docker-compose`.
It will terminate with an error, as before. I've checked this manually.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -8,6 +8,12 @@ def system!(*args)
   system(*args, exception: true)
 end
 
+def docker_compose!(options)
+  system!("docker compose #{options}")
+rescue Errno::ENOENT
+  system!("docker-compose #{options}")
+end
+
 FileUtils.chdir APP_ROOT do
   # This script is a way to set up or update your development environment automatically.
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
Interestingly enough, I still have both commands available for some reason. But this change is great because it doesn't require the old setup/executable to be present.

So, thank you @maciej-ka! 🙌🏼

maciej-ka:
Docker used to create an alias so that both commands worked.
However, my recent install missed it ... so perhaps it's no longer true?

Have a good day, Marco!

marcoroth:
Good to know, thanks for the context @maciej-ka! 🙏🏼

PR #331, opened 2024-10-31
Add Contributions page

marcoroth:
This page should guide people to contribute if they don't know where to start.

CleanShot 2024-11-26 at 21 21 58


app/controllers/contributions_controller.rb
```diff
@@ -0,0 +1,17 @@
+class ContributionsController < ApplicationController
+  skip_before_action :authenticate_user!, only: %i[index]
+
+  def index
+    speaker_ids_with_pending_github_suggestions = Suggestion.where("json_extract(content, '$.github') IS NOT NULL").where(suggestable_type: "Speaker").pluck(:suggestable_id)
+
+    speakers = Speaker.where(github: "").order(talks_count: :desc)
```
adrienpoly, line 7:
```suggestion
    speakers = Speaker.without_github.order(talks_count: :desc)
```

app/controllers/contributions_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+    @speakers_without_github_count = speakers.count
+    @speakers_without_github = speakers.where.not(id: speaker_ids_with_pending_github_suggestions).limit(11)
```
adrienpoly, line 9: I think we should try to randomize a bit the speakers we display here so that we don't end up stuck with only very difficult speakers to identify

app/views/contributions/index.html.erb
```diff
@@ -0,0 +1,70 @@
...
+    <h2 class="mb-4">Speakers without GitHub handles</h2>
+
+    <article class="prose mb-6">These speakers are currently missing a GitHub handle. By adding a GitHub handle to their profile, we can enhance their speaker profiles with an avatar and automatically retrieve additional information.</article>
+
+    <div id="speakers" class="grid sm:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4 gap-x-8 lg:gap-x-12 gap-y-2 min-w-full">
+      <% @speakers_without_github.each do |speaker| %>
+        <%= link_to edit_speaker_path(speaker), id: dom_id(speaker), class: "flex justify-between" do %>
+          <div class="flex items-center hover:bg-gray-200 p-2 rounded-lg">
+            <span><%= speaker.name %></span>
+          </div>
```
adrienpoly, line 32: I would suggest to open it in a remote modal 

adding data: {turbo_frame: "modal"} should do it

adrienpoly (commented):
really like this idea

app/views/contributions/index.html.erb
marcoroth: yes, nice! 🙌🏼

adrienpoly:
@marcoroth anything blocking this?

marcoroth:
The only thing that was blocking this was the way we display and show it, right now the layout isn't really too nice too look at

adrienpoly:
> The only thing that was blocking this was the way we display and show it, right now the layout isn't really too nice too look at

Ok I ll get a look at it

marcoroth:
I think this is good enough for a first pass. We can improve this page as we go 👍🏼

PR #332, opened 2024-10-31
Add `Open in Avo` and `Show on RubyVideo.dev` buttons

marcoroth:
I needed these two actions a bunch in the last few days, maybe it's useful for somebody else too!

CleanShot 2024-10-31 at 10 49 33

### DIFF
--- a/app/views/speakers/actions/_admin.html.erb
+++ b/app/views/speakers/actions/_admin.html.erb
@@ -14,7 +14,7 @@
     <% end %>
   <% end %>
 
-  <%= ui_button url: "/admin/resources/speakers/#{@speaker.slug}", target: :_blank, kind: :ghost, size: :sm do %>
+  <%= ui_button url: avo.resources_speaker_path(@speaker), target: :_blank, kind: :ghost, size: :sm do %>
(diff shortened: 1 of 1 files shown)


app/views/speakers/actions/_admin.html.erb
```diff
@@ -13,4 +13,14 @@
       <span>Fetch from GitHub</span>
     <% end %>
   <% end %>
+
+  <%= ui_button url: "/admin/resources/speakers/#{@speaker.slug}", target: :_blank, kind: :ghost, size: :sm do %>
```
adrienpoly, line 17: I think you could use the avo path helpers here
```suggestion
  <%= ui_button url: avo.resources_speaker_path(@speaker), target: :_blank, kind: :ghost, size: :sm do %>
```

adrienpoly (approved):
great

just one small suggestion

app/views/speakers/actions/_admin.html.erb
marcoroth: Ah nice, didn't know about this 🙌🏼

PR #333, opened 2024-10-31
Add NoRuKo 2020 videos and artwork

marcoroth:
I'm not sure if this should go into the "EuRuKo" organisation or not, since it's technically not a EuRuKo event. But then it kind of filled in the gap for the EuRuKo 2020 conference that didn't happen.

CleanShot 2024-10-31 at 12 35 24


marcoroth:
Going to merge this for now, we can still move the event if we feel it's not in the right org.

Issue #325, opened 2024-10-29
Add a `kind` attribute to the `Talk` model 


adrienpoly:
In the event model we have this query 

```ruby
def keynote_speakers
    talks.select { |talk|
      talk.title.start_with?("Keynote: ") ||
        talk.title.include?("Opening Keynote") ||
        talk.title.include?("Closing Keynote")
    }.flat_map(&:speakers)
  end
```

it would be great to replace it with a real db scope

marcoroth:
Maybe it should be a `talk_type` since we also have other types like "Lightning Talk", "Panel", "Workshop", "Fishbowl" and potentially others.

adrienpoly:
yes great idea 🎉

nicogaldamez:
Hey @marcoroth @adrienpoly ! Happy to help with this.

I like the idea of adding an enum to `Talk` to indicate if it's a Keynote, Lightning Talk, Standard, etc. 

I’ve got a couple of questions:

1) What do you think about naming the enum `kind`? `type` would be ideal, but since it’s a reserved word, we’d have to prefix it, ending up with something like `talk_type`. I usually use `kind` in these cases.
2) We'll need to assign the enum for existing talks.  How are you handling data migrations right now?
3) I’m thinking we could also add this logic in `update_from_yml_metadata` to set the talk kind directly there

marcoroth:
Hey @nicogaldamez, thanks, that would be awesome!

1. Yeah, I think `kind` is ideal since we already use in other models too 👍🏼 
2. Mostly we have been updating the YAML files in `data/` and then re-running the seeds.
3. Yeah, this is kinda what we have been doing, `update_from_yml_metadata!` is what we use to update the data in the database from the YAML files.

So either we can either teach the `update_from_yml_metadata!` what kind of talk it is based on the `title` attribute in the YAML file, or we backfill all the `kind`'s to the YAML files.

nicogaldamez:
Thanks @marcoroth for the quick response :)

>Mostly we have been updating the YAML files in data/ and then re-running the seeds.

How does that work in production? Are seeds run with every deploy?

marcoroth:
I think so, yes!

@adrienpoly has been deploying the changes, but that's how I designed the `update_from_yml_metadata` method to work.

marcoroth:
Resolved via #334

PR #334, opened 2024-10-31
Feature | Adds a kind attribute to talk model

nicogaldamez:
## Why?

* Issue: #325 
* We can identify the type of talk by its title—just by reading it, we can tell if it’s a keynote, panel, workshop, etc.

## What?

* I added a kind attribute to the `Talk` model, assigned automatically based on the talk title.
* All existing talks will be categorized as "standard" until seeds are rerun.
* Paired with @bruno-costanzo


app/models/talk.rb
```diff
@@ -60,8 +60,12 @@ class Talk < ApplicationRecord
   # delegates
   delegate :name, to: :event, prefix: true, allow_nil: true
 
+  # callbacks
+  before_validation :set_kind
```
marcoroth, line 64: Would it make sense to only run this when the title changed as well?
```suggestion
  before_validation :set_kind, if: :title_changed?
```

db/migrate/20241031112333_add_kind_to_talk.rb
```diff
@@ -0,0 +1,9 @@
+class AddKindToTalk < ActiveRecord::Migration[8.0]
+  def change
+    add_column :talks, :kind, :string, null: true
+
+    Talk.update_all kind: :standard
+
+    change_column_null :talks, :kind, false
```
marcoroth, line 7: Would it make sense to just set the default value to `:standard`?

```suggestion
    add_column :talks, :kind, :string, default: "standard", null: false
```

app/models/event.rb
```diff
@@ -66,11 +66,7 @@ def suggestion_summary
   end
 
   def keynote_speakers
-    talks.select { |talk|
-      talk.title.start_with?("Keynote: ") ||
-        talk.title.include?("Opening Keynote") ||
-        talk.title.include?("Closing Keynote")
-    }.flat_map(&:speakers)
+    speakers.merge(talks.keynote)
```
marcoroth, line 69: I'm surprised this works 😅 

Does it know how to merge speakers with talks and its speakers?

marcoroth (commented):
Thank you @nicogaldamez and @bruno-costanzo for taking a stab at this! 🙌🏼

app/models/talk.rb
```diff
@@ -286,4 +290,20 @@ def suggestion_summary
...
+      when /.*workshop:.*/i
+        :workshop
+      when /.*panel:.*/i
+        :panel
+      when /.*lightning\ talk: .*/i
+        :lightning_talk
+      else
+        :standard
+      end
+  end
```
adrienpoly, line 308: I wonder if we should keep a way to force the kind?

```ruby
Talk.create(title: "some title of a group talk", kind: "panel)
```

not sure if there is (yet) a real use case ?

nicogaldamez: @adrienpoly, we thought about it, and the only use case we can think of is if you add a kind in the YML files or create a talk directly from the console.

What do you about adding this to the callback:

```ruby
before_action :set_kind, if: -> { !kind_changed? }
```

Following @marcoroth’s suggestion to add title_changed?, we can include both conditions:

```ruby
before_action :set_kind, if: -> { title_changed? && !kind_changed? }
```

app/models/event.rb
nicogaldamez: It works because it's not merge from Hash or Array, it's [merge from ActiveRecord](https://apidock.com/rails/ActiveRecord/SpawnMethods/merge) ☺️.

I did that to re-use the scope provided by the enum. It's the same as doing this:

```ruby
speakers # this is the has many through association in event
  .where(talk: { kind: 'keynote' }
```

db/migrate/20241031112333_add_kind_to_talk.rb
nicogaldamez: I started by adding a default value at the database level, but ran into some issues with tests. I'll check if adding [these conditions in the callback](https://github.com/adrienpoly/rubyvideo/pull/334#discussion_r1827708596) solves the problem.

app/models/talk.rb
nicogaldamez: I'm running into an issue with using `title_changed` as a condition in the callback. When we run seeds, we need to update existing talks, but it never triggers `set_kind` because the title hasn't changed.

I think it’s better to keep just the !kind_changed? condition so we can skip the callback if a kind is provided.

Thoughts? @marcoroth @adrienpoly

adrienpoly: good for me

app/models/event.rb
marcoroth: Ah I see! I think I never used ActiveRecord's merge with two different models/table, thanks for the context

marcoroth: Wondering if we should promote it to a proper association, something like:

```ruby
 has_many :keynotes, -> { keynote }, class_name: "Talk"
 has_many :keynote_speakers, through: :keynotes, source: :speakers
```

app/models/talk.rb
```diff
@@ -333,4 +338,20 @@ def suggestion_summary
...
+      when /.*(keynote:|opening\ keynote|closing\ keynote).*/i
+        :keynote
+      when /.*workshop:.*/i
+        :workshop
+      when /.*panel:.*/i
+        :panel
+      when /.*lightning\ talk: .*/i
+        :lightning_talk
+      else
+        :standard
```
marcoroth, line 354: I'm wondering if we should use `:talk` instead of `:standard` so we could humanize the `kind` column when we are showing it the UI.

app/models/event.rb
nicogaldamez: @marcoroth I noticed you marked this as resolved. Do you still think it should be promoted to an association?

I'm fine with keeping it as a method, but I'm open to making it an association if you think it's better.

app/models/talk.rb
nicogaldamez: I like it!

app/models/event.rb
marcoroth: I was just thinking out loud, but didn't want to keep requesting things. We can refactor it later if we feel like it needs to be an association 😊

marcoroth (approved):
This looks good to me, thanks @nicogaldamez and @bruno-costanzo! 🙏🏼

Issue #338, opened 2024-11-01
Introduce the concept of "also know as" for events


marcoroth:
A lot of conferences have "multi" names for the conference. I think it would make sense if we maintain a list of "also know as" names for an event.

For example:

* RubyConf AU
* Ruby Conf AU
* RubyConf Australia
* Ruby Conf Australia

Other events have abbreviations:

* GORUCO
* GoRuCo
* Gotham Ruby Conference
* Gotham Ruby Conf
* Gotham Ruby

Similarly with the how an event might be stylized:

* wroclove.rb
* wrocloverb
* wroc_love.rb
* Wroc_love.rb

Other examples:
* Los Angeles RubyConf
* LA RubyConf

Rocky Mountain Ruby:
* Rocky Mountain Ruby
* Rocky Mountain Ruby Conf
* Rocky Mountain Ruby Confernece
* RMR
* rockymtnruby

Red Dot Ruby:
* Red Dot Ruby
* Red Dot Ruby Conf
* Red Dot Ruby Conference
* RedDotRubyConf
* RedDotRubyConference
* RDRC

PR #340, opened 2024-11-01
Brand Gotham Ruby Conference as "GORUCO"

flavorjones:
Brand Gotham Ruby Conference as "GORUCO" and not "GoRuCo"

Although it's appeared both ways, as a co-organizer of the event I can vouch that the preferred branding is shoutcaps.

### DIFF
--- a/data/goruco/goruco-2008/videos.yml
+++ b/data/goruco/goruco-2008/videos.yml
@@ -5,7 +5,7 @@
   raw_title: "Story Driven Development: The Next Generation of Rails Functional Testing by Bryan Helmkamp"
   speakers:
     - Bryan Helmkamp
-  event_name: GoRuCo 2008
+  event_name: GORUCO 2008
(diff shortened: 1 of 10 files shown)


marcoroth (approved):
Awesome, thanks! 🙏🏼 

I've been struggling with this the last few days as I found that a lot of conferences don't have a single way of how they reference to their own conference name. Part of why I just opened #338 today 😅

marcoroth:
Actually, did you also want to change the name in `data/organisations.yml`? 
https://github.com/adrienpoly/rubyvideo/blob/79f06f72b3703cd1511bc5fa24abb55b978bc0cd/data/organisations.yml#L293

flavorjones:
@marcoroth I think leaving the full name of the conference there is OK!

marcoroth:
Sweet, thanks! 🙌🏼

PR #351, opened 2024-11-02
Add utm params for external player

marcoroth:
I think it makes sense that people can measure the traffic coming from RubyVideo.dev


adrienpoly:
Great idea

PR #360, opened 2024-11-03
upgrade to Kamal 2

adrienpoly:


config/environments/production.rb
```diff
@@ -19,7 +19,7 @@
   config.public_file_server.headers = {"cache-control" => "public, max-age=#{1.year.to_i}"}
 
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
-  config.asset_host = "https://www.rubyvideo.dev"
+  config.asset_host = "https://staging.rubyvideo.dev"
```
adrienpoly: No you are right I need to change that, it was part of my testing

marcoroth, line 22: Is this Right?

PR #363, opened 2024-11-03
Add EMEA on Rails 2021 artwork

marcoroth:
| Organisation Show | Event Show |
| --- | --- |
| CleanShot 2024-11-03 at 14 57 57 | CleanShot 2024-11-03 at 14 58 08 |

/cc @hummusonrails


marcoroth:
btw, I also noticed we are missing some talks. Or I guess the workshops weren't recorded, but it seems that Ufuk's talk is missing.

image

hummusonrails:
> btw, I also noticed we are missing some talks. Or I guess the workshops weren't recorded, but it seems that Ufuk's talk is missing.
> 
Yes, not everthing recorded correctly and we lost some of the recorded content. :(

marcoroth:
Ah too bad... 

I added some notes about it in the YAML file so we have the context in the future. 

Thank you!

PR #365, opened 2024-11-03
Upgrade Rails to `8.0.0.rc2`

marcoroth:
Upgrades Rails to `8.0.0.rc2`. I ran the `rails app:update` command and took over the changes that made sense to me.


bin/dev
```diff
@@ -1,8 +1,2 @@
-#!/usr/bin/env sh
-
-if ! gem list foreman -i --silent; then
-  echo "Installing foreman..."
-  gem install foreman
-fi
-
-exec foreman start -f Procfile.dev "$@"
+#!/usr/bin/env ruby
+exec "./bin/rails", "server", *ARGV
```
adrienpoly, line 2: I think we should keep forman here ?

marcoroth: Hmm yeah. Not sure why this changed.

adrienpoly: I think I saw an issue on that not sure why this was changed

adrienpoly: https://github.com/rails/rails/issues/53496

config/puma.rb
```diff
@@ -33,6 +33,9 @@
 # Allow puma to be restarted by `bin/rails restart` command.
 plugin :tmp_restart
 
+# Run the Solid Queue supervisor inside of Puma for single-server deployments
+plugin :solid_queue if ENV["SOLID_QUEUE_IN_PUMA"]
+
```
adrienpoly, line 38: There has been back and forth discussion in SolidQueue on whether it should run as a puma pluging or separatly. 

At first I was using the puma pluging but then as it was about to be removed (finally not) I removed it. now there is a dedicated process in Kamal for the job https://github.com/adrienpoly/rubyvideo/blob/b13149b5177486b2bd37eb049db8e0e4802ffc70/config/deploy.yml#L12-L15

and in  procfile.dev we also start solid queue

marcoroth: Isn't this only running the plugin when the ENV variable is present?

adrienpoly: yes ok let's keep it like that

bin/dev
adrienpoly, line 2:
```suggestion
#!/usr/bin/env sh

if ! gem list foreman -i --silent; then
  echo "Installing foreman..."
  gem install foreman
fi

exec foreman start -f Procfile.dev "$@"
```

PR #370, opened 2024-11-04
scroll into view active talk

adrienpoly:
https://github.com/user-attachments/assets/4f6fbec8-1c7e-4eaf-8670-3cc1decc9872

### DIFF
--- a/app/javascript/controllers/scroll_position_controller.js
+++ b/app/javascript/controllers/scroll_position_controller.js
@@ -1,0 +1,25 @@
+import { Controller } from '@hotwired/stimulus'
+
+// Connects to data-controller="scroll-position"
+export default class extends Controller {
+  static targets = ['active']
+
+  connect () {
+    this.centerCurrentTalk()
+  }
+
+  centerCurrentTalk () {
+    if (!this.hasActiveTarget) return
+
+    const activeCard = this.activeTarget
+
+    const container = this.element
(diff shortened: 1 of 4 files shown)


adrienpoly:
should play nicely with #352

app/javascript/controllers/scroll_position_controller.js
```diff
@@ -0,0 +1,25 @@
...
+    const activeCard = this.activeTarget
+
+    const container = this.element
+    const containerHeight = container.clientHeight
+    const cardPosition = activeCard.offsetTop
+    const cardHeight = activeCard.clientHeight
+
+    // Calculate scroll position to center the card
+    const scrollPosition = cardPosition - (containerHeight / 2) + (cardHeight / 2)
+    container.scrollTop = Math.max(0, scrollPosition)
```
marcoroth, line 23: Thoughts? 🙈 

```suggestion
    this.activeTarget.scrollIntoView({ block: "center" })
```

marcoroth: Ah, I guess this doesn't work in all cases.

adrienpoly: not sure why when I first tried I had some issue, I tried again and it seems to work fine. I apply this change in took the opportunity to rename the controller scroll-into-view to make it more generic

PR #375, opened 2024-11-05
Add talk card play hover

marcoroth:
https://github.com/user-attachments/assets/81367651-8763-4155-9108-8b2f51535564


marcoroth:
For talks in other languages it also reveals the language on hover:


https://github.com/user-attachments/assets/8339c72c-89e9-42fe-a850-bf7d814cb42c

PR #379, opened 2024-11-06
add staging manual deploy step

adrienpoly:


adrienpoly:
it is manual, but should allow us to trigger manually a deploy to staging from a branch

Issue #364, opened 2024-11-03
Ability to mark talk as seen


marcoroth:
If you've seen a talk it would be cool to be able to "mark it as seen". When signed in we could visually show to the user that they've already seen this talk. 

Maybe we can even ask the user if they've seen it in-person at the event or online.

nicogaldamez:
Happy to help with this!

What do you think about having a button next to edit button:
<img width="854" alt="image" src="https://github.com/user-attachments/assets/8ec099ba-aac2-427d-a6f8-530468deed52">
![image](https://github.com/user-attachments/assets/9da35214-d37f-4a16-8571-51d67bbda7a6)

adrienpoly:
I think we need two new buttons. Bookmark (add to favorite) and mark as watched. The location looks good to me but I feel like those two buttons should be grouped somewhere and we might be lacking a bit of space here. or we have to move the edit button ?

marcoroth:
I guess depending on how we (re-)arrange it we can look how YouTube is doing it:

![CleanShot 2024-11-05 at 23 25 07](https://github.com/user-attachments/assets/29006303-0876-431b-a8c1-e044a4be512b)

nicogaldamez:
Sounds good! How about we go ahead with adding the "Mark as watched" button now, then handle the Bookmark button and layout adjustments in a separate PR?

"Mark as Watched / Unmark as Watched" texts sound good to you?

adrienpoly:
> Sounds good! How about we go ahead with adding the "Mark as watched" button now, then handle the Bookmark button and layout adjustments in a separate PR?

Yes sure just go with one button first 

> "Mark as Watched / Unmark as Watched" texts sound good to you?

I feel it is a bit long but I cannot find anything real better so for me let's go like that

marcoroth:
Here's an idea on how we could keep it short:

![Talk Show Interaction Buttons@2x](https://github.com/user-attachments/assets/ff4e96c8-ee46-426b-9e37-837e651f3204)

marcoroth:
Resolved via #383

PR #383, opened 2024-11-07
Mark talk as Watched/Unwatched

nicogaldamez:
## Why?

* Issue: #364 
* Mark talk as watched/unwatched

## What?

* I added a button below the video player to mark the talk as watched or unwatched
* The button is only visible to signed-in users.

image
image


nicogaldamez:
@marcoroth @adrienpoly any idea why CI is failing?

image

adrienpoly:
This is the test that run the entire seed, it only runs in the CI as it is fairly long this is why you probably don't see this error in dev.

To go back to the error by itself I had a similar case in the past. I think we might have to either remove the constraint or make it deferrable 

More info here : 
https://github.com/rails/rails/pull/49376

app/views/talks/_talk.html.erb
```diff
@@ -43,14 +43,36 @@
...
-            <%= heroicon :pencil, size: :sm %>
-            <span>Edit</span>
+        <div class="flex gap-1">
+          <% if signed_in? %>
+            <% if talk.watched? %>
+              <div class="tooltip tooltip-top" data-tip="Mark talk as unwatched">
+                <%= button_to [talk, :watched_talk],
+                      class: "flex gap-2 items-center rounded-full border bg-gray-100 hover:bg-gray-200/50 border-gray-200 font-regular px-3 py-1 text-sm",
+                      method: :delete do %>
+                  <%= heroicon :x_mark, size: :sm %>
```
marcoroth, line 56: I somehow feel like the `x` is weird here, since you "unwatch" is usually not a typical action.

marcoroth (approved):
This is great, thanks @nicogaldamez! 🙌🏼 

I think I only have one minor thing about the icon for the "unwatch" action, but other than that this looks great to me!

app/views/talks/_talk.html.erb
```diff
@@ -43,14 +43,36 @@
...
+                  <span class="text-nowrap">Unwatched</span>
```
marcoroth, line 57: Also, maybe just "Unwatch" here for the text. Because "Unwatched" could be seen that it's currently in the "Unwatched" state which technically means the exact opposite 😅

app/views/talks/_talk.html.erb
nicogaldamez: @marcoroth  what do you think of something like this?
Kapture 2024-11-11 at 07 41 42

marcoroth: I personally like this a lot better, thanks for exploring 🙌🏼

marcoroth: Thinking about it again, I wonder if the text should even change when the state changes.

What if you have an empty circle for the "unwatched" state that gets checked off if you mark it as "watched" but the text always says "Watched"?

nicogaldamez: @marcoroth I first looked for an empty circle on the [heroicons list](https://heroicons.com) but couldn’t find one, so I ended up going with `minus-circle`.

Here we have the two versions.
Kapture 2024-11-11 at 08 09 12

Kapture 2024-11-11 at 08 13 48

I'm not sure which one works better. What do you think?

adrienpoly:
I like the way it is shaping, good job!

I ran this PR locally an I am getting this error when I click a second time the button, not sure if there is something with my db or what???

CleanShot 2024-11-14 at 21 59 28@2x

nicogaldamez:
@adrienpoly that shouldn't happen. When you mark a talk as watched, the button should change to "Unwatched", and it shouldn't try to mark it as watched again. Anyway, I replaced the `create!` with a `find_or_create_by!` to avoid this issue.

Let me know if you still have any problems.

marcoroth:
Thanks for working on this @nicogaldamez. I slightly tweaked the styles of the button. Before I was slightly confused in which state a video is.  I hope this improves the usability aspect a bit and hope you don't mind. 🙌🏼 

CleanShot 2024-11-20 at 02 51 47

nicogaldamez:
@marcoroth I love it 🤩

PR #384, opened 2024-11-07
Implement basic talks filter based on language and kind

marcoroth:
https://github.com/user-attachments/assets/df76062a-42aa-4de2-96f9-bdcb026abec7

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -14,13 +14,7 @@ class EventsController < ApplicationController
   def show
     set_meta_tags(@event)
 
-    event_talks = @event.talks
-    if params[:q].present?
-      talks = event_talks.pagy_search(params[:q])
-      @pagy, @talks = pagy_meilisearch(talks, limit: 21)
-    else
-      @pagy, @talks = pagy(event_talks.with_essential_card_data.order(date: :desc), limit: 21)
-    end
+    @talks = @event.talks.with_essential_card_data.order(date: :desc)
   end
 
   # GET /events/1/edit
--- a/app/javascript/controllers/index.js
+++ b/app/javascript/controllers/index.js
@@ -22,6 +22,9 @@ application.register("scroll-position", ScrollPositionController)
 import SplideController from "./splide_controller"
 application.register("splide", SplideController)
 
+import TalksFilterController from "./talks_filter_controller"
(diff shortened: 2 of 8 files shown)


adrienpoly:
I like it very much 

One concern  is that it introduce a new custom kind of button. 
I now from experience that it is very easy to end up with UI having 10s of variation of buttons and color very quickly. 
I understand that our set of buttons might not be perfect as they are but I would prefer that we keep consistent sets of buttons across all screens. This will make it easier to improve our overall design down the road

CleanShot 2024-11-08 at 09 59 46@2x

To get something closer to what you had, I made those changes

```diff
themes: [
      {
        rubyvideoLight: {
          ...defaultTheme,
          '--btn-text-case': 'none',
          primary: '#DC143C',
          'primary-content': '#ffffff',
          secondary: '#7A4EC2',
          'secondary-content': '#ffffff',
          accent: '#1DA1F2',
          'accent-content': '#ffffff',
          neutral: '#261B23',
          'neutral-content': '#ffffff',
          'base-100': '#F8F9FA',
+          '--rounded-btn': '999px'
          // 'base-content': '#2F2F2F'
          // 'base-100': '#ffffff',
          // info: '#3abff8',
          // success: '#36d399',
          // warning: '#fbbd23',
          // error: '#f87272'
        }
      }
```

```erb
<% classes = "btn btn-sm btn-outline" %>
<% active_classes = "btn-primary btn-active" %>
```
then in a subsequent PR we could explore what should be our default button for tool bars, outlined with a gray border like you proposed looked great

adrienpoly:
Given #433, I wonder if we should introduce some js filters vs having more global talks filters solution.
We could imagine to have in the events#show page a turbo frame with `src=/talks?event=event-slug`. This would somehow enable filtering/sorting in all talks grid views (speakers, topics, events)

marcoroth:
I'll rebase this one as soon as have #468 merged in 👍🏼

adrienpoly:
#468 is merged so you can rebase this one. 

I think we can go ahead with it like that now but down the road I think the filters can be a backend functionality. 

given #433 and the exploratory work in #440 I will try to see if we could have a way to easily embed in a page a grid of talks with some filtering capability so that filters can be available everywhere (topics, event, speakers) and independent of the pagination.

but once again that can be done after this PR is already a great improvement

PR #397, opened 2024-11-10
Upgrading to yarn 4.x

guillaumebriday:
- Upgrading from yarn 1.22 to yarn 4.x (berry)
- Using the `.node-version` on the CI
- Using corepack to automatically install yarn in the Dockerfile
- The diff is huge because of the new `yarn.lock`


marcoroth:
Hey @guillaumebriday, thanks for this pull request. I'm not sure this makes sense, personally I'd rather switch back to pnpm or npm if we move off of Yarn 1.x.

PR #398, opened 2024-11-10
Declare assets as immutable

guillaumebriday:
- See: https://twitter.com/guillaumebriday/status/1851979909058515089
- and https://github.com/rails/rails/pull/53067


adrienpoly:
Thanks I didn't know about this directive. I'll read a bit about it before merging

adrienpoly:
Thanks this looks indeed interesting

PR #399, opened 2024-11-10
Cleaning deploy files

guillaumebriday:
- Removing useless comments
- Removing useless environment variables

### DIFF
--- a/config/deploy.staging.yml
+++ b/config/deploy.staging.yml
@@ -8,6 +8,7 @@ image: adrienpoly/rubyvideo_staging
 servers:
   web:
     - 5.75.232.156
+
   job:
     hosts:
       - 5.75.232.156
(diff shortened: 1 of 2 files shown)


config/deploy.staging.yml
```diff
@@ -44,67 +41,20 @@ builder:
   secrets:
     - RAILS_MASTER_KEY
 
-#
 env:
   clear:
-    RUBY_YJIT_ENABLE: 1
```
guillaumebriday, line 50: yjit is now enable by default

config/deploy.staging.yml
```diff
@@ -44,67 +41,20 @@ builder:
...
-    RAILS_SERVE_STATIC_FILES: true
```
guillaumebriday, line 51: Never used, and static files are now served by default: https://edgeguides.rubyonrails.org/configuring.html#config-public-file-server-enabled

adrienpoly:
thanks for the clean up

Issue #391, opened 2024-11-09
Reduce CI test time


adrienpoly:
The test job now takes about 3 minutes to run. It is ok but given the size of the app it feels a lot. Mot of the time is spent in the seed test. Those tests are important as they ensure seed will run smoothly but as the data is growing the test suite will become slower and slower.

One solution I could see is that we have a list of event that we consider done and that we don't seed over and over every time. and if needed we have the action in the admin to do so.

Happy to hear other suggestions

marcoroth:
I think we just have to improve the way we seed. There's probably a lot we can improve and optimize.

marcoroth:
We might be able to close this issue now after merging #392, depending on how you feel @adrienpoly

adrienpoly:
yes much better now. I  might in the future extract the seed test to its own workflow but let's close this now

PR #401, opened 2024-11-10
extract smoke test

adrienpoly:
extract the seed smoke test in its own workflow. I think this should make test/seed/lint all run under 1min

relates to #391


adrienpoly:
sweet
CleanShot 2024-11-10 at 16 00 36@2x

adrienpoly:
it was more or less 2min I think
https://github.com/adrienpoly/rubyvideo/actions/runs/11765630432

marcoroth:
Ah I guess the Web UI was just cutting it short and said "1m" instead of "1m 50s".

PR #407, opened 2024-11-12
Global search bar

adrienpoly:
This PR adds a global search bar available with cmd+k hotkeys

https://github.com/user-attachments/assets/bad7c1a6-582c-495e-9a3a-38f3eb93d5e8

The way it is build allows to perform multiple async queries in parallel to different Resources. From a UX point of view maybe 3 resource is too much?? Open to discuss this

### DIFF
--- a/app/views/shared/_navbar.html.erb
+++ b/app/views/shared/_navbar.html.erb
@@ -1,6 +1,6 @@
 <div class="navbar relative z-50 bg-transparent">
   <div class="container grid grid-cols-2 lg:grid-cols-3 justify-between items-center">
-    <div class="items-center gap-6 hidden md:flex">
+    <div class="items-center gap-6 hidden lg:flex">
       <ul class="desktop-menu">
         <%= render "shared/navbar/link", link_title: "Talks", path: talks_path %>
         <%= render "shared/navbar/link", link_title: "Speakers", path: speakers_path %>
@@ -16,7 +16,7 @@
     </div>
 
     <div class="flex justify-end items-center">
-      <ul class="desktop-menu items-center">
+      <ul class="hidden lg:flex items-center gap-4">
         <%= render "shared/navbar/search_bar" %>
 
         <% signed_in do %>
@@ -38,7 +38,7 @@
         <% end %>
       </ul>
 
-      <div class="mobile-menu gap-2">
+      <div class="flex lg:hidden gap-2">
         <%= render "shared/navbar/search_bar" %>
         <%= render "shared/user_mobile_dropdown" %>
       </div>
--- a/app/views/shared/_navbar.html.erb
+++ b/app/views/shared/_navbar.html.erb
@@ -9,14 +9,16 @@
       </ul>
     </div>
 
-    <div class="flex md:hidden lg:flex justify-start lg:justify-center items-center">
+    <div class="flex justify-start lg:justify-center items-center">
       <%= link_to root_path do %>
         <%= image_tag image_path("logo.png"), class: "size-10 hover:opacity-80" %>
       <% end %>
     </div>
 
     <div class="flex justify-end items-center">
-      <ul class="desktop-menu">
+      <ul class="desktop-menu items-center">
+        <%= render "shared/navbar/search_bar" %>
+
         <% signed_in do %>
           <div class="flex items-center gap-2">
             <%= render "shared/navbar/link", link_title: "Watch List", path: watch_lists_path %>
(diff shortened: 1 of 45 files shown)


app/views/shared/_navbar.html.erb
```diff
@@ -27,12 +29,6 @@
           </div>
         <% end %>
 
-        <li class="flex content-center">
-          <%= link_to "https://github.com/adrienpoly/rubyvideo", target: "_blank", alt: "github repository for the Ruby video project" do %>
-            <%= icon "github", size: :lg %>
-          <% end %>
-        </li>
-
         <% signed_in do %>
```
adrienpoly, line 32: I think we could place it in the footer now

app/views/shared/_spotlight_search.html.erb
```diff
@@ -0,0 +1,27 @@
+<% main_resource = request.path.split("/").last.presence_in(%w[talks speakers events]) || "talks" %>
```
adrienpoly, line 1: depending on the location the `enter` key will navigate to different resource

When on the speaker page, type `da` + enter sends you to `/speaker?s=da`

open to discuss this UX if not clear

adrienpoly:
I pushed a live version to test it to the staging env at https://staging.rubyvideo.dev/

app/views/shared/_spotlight_search.html.erb
```diff
@@ -0,0 +1,36 @@
...
+<% resources = %w[talks speakers events] %>
+
+
+<%= ui_modal position: :top, close_button: false, data: {action: "keydown.meta+k@window->modal#open spotlight@window->modal#open"}, class: "spotlight"  do %>
```
nicogaldamez, line 5: I'm a Mac user, so I can't test it. How would a Windows/Linux user use the hotkey?

app/views/spotlight/talks/index.turbo_stream.erb
```diff
@@ -0,0 +1,20 @@
+
+<%= turbo_stream.update "talks_search_results", method: "morph" do %>
+  <div class="flex items-center flex-wrap gap-2 py-4">
+    <h2>Talks</h2>
+    <% if @talks.size.positive? %>
+      <span class="text-sm text-gray-500">(<%= @talks.size %> of <%= @talks_count %>)</span>
+      <%= link_to talks_path(s: search_query), class: "link text-sm text-gray-500", target: "_top" do %>
```
nicogaldamez, line 7: What do you think about displaying the search query on the talks list after the user clicks See Results? For example, I searched for 'postgres,' and I can see the highlighted text, but maybe showing the query explicitly would give more context to the user. Do you think that's unnecessary?

image

The same for speakers and events. Here I searched for "Amanda":
image

app/views/shared/navbar/_search_bar.html.erb
```diff
@@ -0,0 +1,8 @@
+
+<label class="input input-bordered input-sm flex items-center gap-2">
+  <input type="text" class="grow" placeholder="Search" autocomplete="off" data-controller="event" data-action="click->event#dispatchEvent" data-event-name-value="spotlight">
```
nicogaldamez, line 3: I noticed that if I open the modal and press ESC, this input still has focus, and I can't search from there. Maybe it could be a button styled to look like an input instead of being a real input.

app/views/shared/_spotlight_search.html.erb
adrienpoly: I am not a specialist but I think the `meta` key is cross platform https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey

on a related topic in a perfect world the hint for the hotkey should be plaform specific. Not sure I want to go that fare in this PR 😄 

CleanShot 2024-11-15 at 16 02 06@2x

app/views/spotlight/talks/index.turbo_stream.erb
adrienpoly: yes very good idea, this is a rather common UX and fairly easy to implement

app/views/shared/navbar/_search_bar.html.erb
adrienpoly: I will replace this by just a simple icon. It will look better in mobile too

app/views/spotlight/talks/index.turbo_stream.erb
adrienpoly: CleanShot 2024-11-17 at 22 17 49@2x

adrienpoly: I have updated the title like that ☝️

adrienpoly:
They are plenty of refinements possible, by example : 
- add topics search 
- make events/speakers/topics endpoint fuzzy searchable with FT5


but I think this is already a major improvements to search so let's merge it and build upon that

PR #412, opened 2024-11-17
remove hotwire progress

adrienpoly:
not a big fan of the hotwire progress bar


marcoroth (commented):
I personally like the progress bar. At least it gives you an indication that something is happening.

PR #413, opened 2024-11-18
fix css for avatar in tab

### LINKED ISSUE #411: In talk show view, clicking speaker tab sends tab content to bottom of page
When I click the speaker tab here:
image

The content is sent to the bottom of the page:
image

This seems to be caused by the `.avatar` class in this container:

image

When I deselect `.avatar>div: aspect-ratio: 1 / 1` the layout is fixed.

Safari 17.6 MacOS

adrienpoly:
close #411

We had a bug when displaying an avatar inside a tab in Safari 
Setting the img height to auto rather than 100% seems to do the trick

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -2,6 +2,7 @@
 @import "tailwindcss/components";
 @import "tailwindcss/utilities";
 
+@import "components/avatar.css";
 @import "components/diff.css";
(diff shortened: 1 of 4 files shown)


adrienpoly:
after further investigation I found this issue https://github.com/saadeghi/daisyui/issues/2988 and the suggested fix seems to correct the problem on safari

PR #421, opened 2024-11-20
navigate talks in turbo frame

adrienpoly:
Display talk in a turbo frame so that we can navigate from one event talk to the other instantly without loosing track

https://github.com/user-attachments/assets/1daeebdd-9a1a-4255-8482-99d048c0363a

### DIFF
--- a/app/views/talks/_card_horizontal.html.erb
+++ b/app/views/talks/_card_horizontal.html.erb
@@ -1,21 +1,25 @@
-<%# locals: (talk:, current_talk: nil, compact: false, watched_talks_ids:) -%>
+<%# locals: (talk:, current_talk: nil, compact: false, turbo_frame: nil, watched_talks_ids:) -%>
 
 <% active = talk == current_talk %>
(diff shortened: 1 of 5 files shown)


app/views/talks/_card_horizontal.html.erb
```diff
@@ -1,21 +1,25 @@
...
-        <%= fa("circle-check", size: :sm, class: "fill-white") %>
-      </button>
-    <% end %>
+    <div class="absolute inset-0 bg-black/50 justify-center items-center rounded hidden group-[.active]:flex">
+      <%= fa("play", size: :sm, class: "fill-white") %>
+    </div>
+
+    <button data-tip="Talk Watched" class="tooltip tooltip-right absolute inset-0 bg-black/50 justify-center items-center rounded hidden group-[.watched]:flex group-[.active]:hidden">
+      <%= fa("circle-check", size: :sm, class: "fill-white") %>
+    </button>
```
marcoroth, line 22: Does this account for the case where you mark a talk as watched so that it will add the checkmark to the cards in the sidebar?

adrienpoly: currently it does not update the side bar. I will address that in a follow up PR

Issue #404, opened 2024-11-10
Tabs on Speaker Show Page Show wrap on mobile


marcoroth:
Right now the tabs on the speaker show page don't wrap and thus let the whole page content overflow and expand because of it.

adrienpoly:
In TailwindUI Tabs in mobile become select

![IMG_9586](https://github.com/user-attachments/assets/70cade80-00a1-4444-a794-b1b193093979)


This other alternative is to make the tabs header scrollable

marcoroth:
I'm surprised daisyUI doesn't handle it out of the box.

marcoroth:
Here for example:

![image](https://github.com/user-attachments/assets/fc42976d-99c2-448d-9c7c-3ccae92bfa23)

https://www.rubyvideo.dev/talks/a-brewer-s-guide-to-filtering-out-complexity-and-churn-rocky-mountain-ruby-2024

nicogaldamez:
Thanks, @marcoroth, that helps a lot!

Along with the TailwindUI select option, here are two more alternatives based on GitHub's UI approach.

1. How about a list of collapsible items (accordion-style)? Imagine something like this, but instead of navigating to another page, it expands in place.
![image](https://github.com/user-attachments/assets/779c597c-5e09-416a-9fa2-6bb8785a4f5a)

2. Using horizontally scrollable tabs. We may want to add a background to make it clear to users that they can scroll to see more
![image](https://github.com/user-attachments/assets/9453bb1d-2d12-4faf-866d-e9a5adb6e0ed)

adrienpoly:
I like the collapsible option very much but in our case given the length of the transcript tab it might not be really suitable. 

The select option from TailwindUI is the most responsive but I am not a big fan of the UX (open to change my mind)

I think scrollable tab would be a good candidate here 

my 2 cents

adrienpoly:
following to this change https://github.com/adrienpoly/rubyvideo/pull/413 the tabs do wrap now on mobile. Maybe not optimum but at least better than creating an overflow. 

given also the changes in #423 at least now the talks#show page should not overflow anymore 

We might still want to make here the tabs scrollable to prevent the wrap?

marcoroth:
Looks like this was resolved via https://github.com/rubyevents/rubyevents/pull/772

PR #423, opened 2024-11-20
show less avatars in mobile to prevent overflow

adrienpoly:
this section would overflow on mobile thus the entire page overflows as seen in #404

### DIFF
--- a/app/views/talks/_talk.html.erb
+++ b/app/views/talks/_talk.html.erb
@@ -234,8 +234,8 @@
     <div class="avatar-group -space-x-3 rtl:space-x-reverse mt-4">
       <% speakers_with_avatars = talk.event.speakers.where.not(github: "").sample(12) %>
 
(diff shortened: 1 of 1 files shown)


app/views/talks/_talk.html.erb
```diff
@@ -234,8 +234,8 @@
     <div class="avatar-group -space-x-3 rtl:space-x-reverse mt-4">
       <% speakers_with_avatars = talk.event.speakers.where.not(github: "").sample(12) %>
 
-      <% speakers_with_avatars.each do |speaker| %>
-        <div class="avatar bg-white border-2">
+      <% speakers_with_avatars.each_with_index do |speaker, index| %>
+        <div class="<%= class_names("avatar bg-white border-2", "hidden lg:block" => index > 8) %>">
```
marcoroth, line 238: I think this technically breaks the `+[number]` count on mobile, since you don't add the missing four to the count.

PR #427, opened 2024-11-20
set event and organisation association as strict_loading true and remove n+1

adrienpoly:
This PR set event and organisation association as strict_loading to prevent futur eN+1 on those associations

The idea would be to gradually make associations more strict to prevent perf regressions down the road

This is a first small step in that direction


app/models/talk.rb
```diff
@@ -206,7 +206,7 @@ def thumbnail(size = :thumbnail_lg)
       end
     end
 
-    if event && (asset = Rails.application.assets.load_path.find(event.poster_image_path)) && !youtube?
+    if !youtube? && event && (asset = Rails.application.assets.load_path.find(event.poster_image_path))
```
marcoroth, line 209: Is this for delaying the event lookup and fetch?

adrienpoly: Yes I noticed that we very frequently query the event but most of the time we don’t need to. If we are in a YouTube talk thé we can move directly to the next part without making this extra query 

this method is a bit complex so let me know if I missed something

marcoroth: That looks good, was just wondering because it looked unrelated at first. But I think this is good 👍🏼

PR #433, opened 2024-11-22
add event, speaker, kind, topics talks filter and asc/desc order_by

adrienpoly:
It doesn't do much at this point but it give the possibility to manually filter the talks results from the url

**Examples:**

- `/talks?kind=keynote&order_by=date_asc`
- `/talks?speaker=marco-roth&order_by=date_desc&topic=hotwire`

This is the foundation to build filters in addition to the search


marcoroth (approved):
This looks good to me!

PR #435, opened 2024-11-23
Add a `scheduled` video provider

marcoroth:
This pull request implements a `scheduled` video provider that can be used to index talks for upcoming conferences early. A `scheduled` talk can be changed later to either `not_recorded` or `youtube`/`mp4` by adding the recording of the talk.

| Talk Card |
| --- | 
|  CleanShot 2024-11-23 at 07 02 42 |

| Talk Show |
| --- |
| CleanShot 2024-11-23 at 07 01 41 |


adrienpoly:
I like the idea of those scheduled talks. I wonder though if we should display them in the /talks index view?
Maybe we should display them only under the event show page?

marcoroth:
@adrienpoly yeah good point! I think we can exclude them from the talk/index yes

PR #436, opened 2024-11-23
Add more Talk kinds and improve the matching of existing kinds

marcoroth:
We have a lot more different kinds of "talks" in our current dataset. This pull request adds a few more kinds to better represent our dataset:

Follow up on #334

### DIFF
--- a/app/models/talk.rb
+++ b/app/models/talk.rb
@@ -82,7 +82,7 @@ class Talk < ApplicationRecord
   # enums
   enum :video_provider, %w[youtube mp4 scheduled not_recorded].index_by(&:itself)
   enum :kind,
-    %w[talk keynote lightning_talk micro_talk panel workshop gameshow fishbowl podcast q_and_a discussion fireside_chat
+    %w[talk keynote lightning_talk panel workshop gameshow podcast q_and_a discussion fireside_chat
       award].index_by(&:itself)
 
   # attributes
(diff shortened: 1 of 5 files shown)


app/models/talk.rb
```diff
@@ -356,18 +358,33 @@ def suggestion_summary
...
+    when /.*(fishbowl:|fishbowl\ discussion:).*/i
+      :fishbowl
+    when /.*(podcast:|podcast\ recording:|live\ podcast:).*/i
+      :podcast
+    when /.*(q&a|q&a:|ama|q&a\ with|ruby\ committers\ vs\ the\ world|ruby\ committers\ and\ the\ world).*/i
+      :q_and_a
+    when /.*(discussion:|discussion).*/i
+      :discussion
+    when /.*(fireside\ chat:|fireside\ chat).*/i
+      :fireside_chat
```
adrienpoly, line 392: fireside_chat isn't it a kind of q_and_a?

same for panel, fishbowl, discussion aren't they a similar kind at the end but with different naming in conferences?

marcoroth: Yeah, they are close to each other, but I guess the formats have different nuances which make them unique.

marcoroth: But maybe we can group them somehow that makes sense?

adrienpoly: I wonder if someone would search specifically for a fishbowl ?

marcoroth: I combined them 👍🏼

PR #438, opened 2024-11-23
Set correct publication dates for Rocky Mountain Ruby 2011 and 2013

adrienpoly:
I could not retrieve the exacte schedule so the order is not perfect

relates to : https://github.com/adrienpoly/rubyvideo/issues/376


marcoroth:
https://web.archive.org/web/20220706071320/https://confreaks.tv/events/rockymtnruby2011
https://web.archive.org/web/20110927200359/http://rockymtnruby.com/

marcoroth (approved):
Sadly couldn't figure out the exact order either, but at least the talks from the 2011 edition should have the right talk dates now.

PR #440, opened 2024-11-23
load topics talks in a turbo frame

adrienpoly:
for now I would like to test this only for admin to play a bit with it to be sure they are no side effects.


marcoroth:
Does this really make a difference? A talk usually only has 4-5 topics, no?

adrienpoly:
It is not for displaying the topics of a talk but to display the talks of a topic. 
like for this page https://www.rubyvideo.dev/topics/abstract-syntax-tree-ast

Given the filter introduce in #433 we can now display within a turbo frame the talks of a topics and we could extend that to speaker#show page and event#show page. That would DRY quite a bit the rendering of talks card.

adrienpoly:
ok I wen't much further :
- added infinite scroll to the talks index view to remove pagination
- all talks grid view in Events#show, Speakers#show, Topics#show are now displayed using the same turbo frame from the talks#index view

The benefits I see is to dry a bit that part of the views and when we start adding filters, sorting etc that can be added globaly to all views

Issue #439, opened 2024-11-23
Extract User Avatars into Component


marcoroth:
We have a bunch of representations of user avatars, different sizes, some with white borders, some with name initials fallbacks, some with neither. I think it would make sense to extract a ViewComponent for the avatars so we have one representation we can use throughout the app.

Ideally we could even fetch the profile image from somewhere else (like LinkedIn, X, Bsky) in addition to just GitHub and maybe incorporate https://github.com/adrienpoly/rubyvideo/issues/330.

aronwolf90:
Hi, if you want I can help here. My only problem is, that almost every place I looked, the avatar has a different code and also looks differently. Of course, I can move all of them into one components with a lot of if/else but not sure if this is what you want.

marcoroth:
Hey @aronwolf90, yeah part of the motivation of this issue is to unify the look and code for the avatars that we use throughout the codebase.

Ideally, it would take the best of them all. Including the initials fallback if an image isn't able to load.

But it would be very much appreciated if you want to take a stab!

aronwolf90:
Here is the PR https://github.com/rubyevents/rubyevents/pull/930

PR #463, opened 2024-11-28
Introduce `Speaker#avatar_url` and rework avatar rendering

marcoroth:
This pull request introduces `Speaker#avatar_url` which now defaults to the Bsky profile picture and falls back to GitHub or otherwise the initials. This guarantees that we always have a profile picture to show, for any speaker.

Ideally with #439 we can simplify this setup even more and can get rid of the UI Avatars API by just rendering the initials in HTML, and as fallbacks.

But for now this is a good trade-off.

Dependent on #462.

### DIFF
--- a/app/models/speaker.rb
+++ b/app/models/speaker.rb
@@ -124,10 +124,39 @@ class Speaker < ApplicationRecord
     super
   end
 
+  def avatar_url(...)
+    bsky_avatar_url(...) || github_avatar_url(...) || fallback_avatar_url(...)
+  end
+
+  def avatar_rank
+    return 1 if bsky_avatar_url.present?
+    return 2 if github_avatar_url.present?
+
+    3
+  end
+
+  def custom_avatar?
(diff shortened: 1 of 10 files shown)


adrienpoly:
> This pull request introduces `Speaker#avatar_url` which now defaults to the Bsky profile picture and falls back to GitHub or otherwise the initials. This guarantees that we always have a profile picture to show, for any speaker.

Any particular reason to default to Bsky rather than Github? I would have kept Github as the main. I think most of the speakers have a Github account and while Bsky is getting some traction it is still fairly new and I am unsure if it will still be this new shiny thing 6 months from now.

marcoroth:
I defaulted it to Bsky because I noticed that a lot of people don't use real profile pictures (or super old ones) on GitHub while they have more up-to-date profile pictures on his new-ish platform.

I'd love to do the same for the other socials so we can have a more robust story for avatars.

marcoroth:
But that being said, I'm also happy to revert it back and make it GitHub first again.

PR #466, opened 2024-11-28
Extract enhancer jobs to a speaker associated object

adrienpoly:
I like quite a lot the [associated object](https://github.com/kaspth/active_record-associated_object) pattern combined with the [performs gem](https://github.com/kaspth/active_job-performs). 

This is a WIP that extract the Bsky enhancer job to an Speaker::ProfileEnhancer associated object 

I personnaly like how we can :
1. remove the job files
2. quickly define methods that can be both called synchronously or asynchronously 
3. get a much nicer API (IMHO) `speaker.profile_enhancer.enhance_all_later!`

not tested yet and the Github part is missing just for now a POC to see how it could look like


marcoroth (commented):
Yes, this is great! I like the two gems as well and think this is a great way to refactor this 🙌🏼

kaspth:
Hell yes, thanks for using these! ❤️

adrienpoly:
merging this so that we can go further with #495 and #496

PR #75, opened 2024-02-25
Add playback rate control button

b1anca:
Related to https://github.com/adrienpoly/rubyvideo/issues/74

Add playback rate selection to the video player, allowing to choose between 1x, 1.25x, 1.5x, 1.75x, and 2x speeds

https://github.com/adrienpoly/rubyvideo/assets/28804464/a656df99-6f54-42ed-b569-977b8461ace8

PR #467, opened 2024-11-29
Add open in YouTube functionality

### LINKED ISSUE #410: YouTube Sign-in required in some conditions, no link to source to workaround
From my work VPN, I'm unable to watch embedded YouTube videos without being presented with the "Sign-in to confirm you are not a bot" error.

image

Over at Real Youtube, my account is currently signed in and I am able to watch videos, but there isn't an easy way to get there from the video index.

I propose adding a link to the hosted source material to allow users to give a workaround to resolve this type of issue. It also adds transparency with video hosting attribution.

aamfahim:
Closes #410 
Added button on the video player to open YouTube video in a new tab.

https://github.com/user-attachments/assets/d8cf02a1-deb7-4430-9479-a8cec273814d


app/assets/stylesheets/components/video.css
```diff
@@ -17,4 +17,9 @@
   .v-bigPlay {
     opacity: 0.5 !important;
   }
+  
+  .v-openInYouTube.v-controlButton svg {
+    width: 26px;
```
adrienpoly, line 22: I think we should try to avoid px base sizes. For now if I zoom in with my browser the other buttons are getting bigger t. This new button should have the same behaviour

app/javascript/controllers/video_player_controller.js
```diff
@@ -1,6 +1,7 @@
 import { Controller } from '@hotwired/stimulus'
 import Vlitejs from 'vlitejs'
 import VlitejsYoutube from 'vlitejs/providers/youtube.js'
+import youtubeSvg from '../../assets/images/icons/fontawesome/youtube-brands-solid.svg?raw'
```
adrienpoly, line 4: I wonder if importing assets like that will work in production when the asset is digested? Never done it with Vite/Propshaft in the past. If needed I can try to deploy this branch on our staging environement

adrienpoly (commented):
Thanks, initially I want thinking to have some kind of button outside of the player but I like this idea to put it inside the toolbar.

There is one issue though is that the toolbar is only visible once you click on the video to play it. So this might be creating a strange UX as you need to start the video to open in in a new tabs. Then go back to the previous tab to stop the video and then go back t youtube tab to play it.

app/javascript/controllers/video_player_controller.js
marcoroth: That should be fine, since the JavaScript is built (and the SVG inlined) at compile-time.

app/assets/stylesheets/components/video.css
aamfahim: Hmm, I do agree with avoiding base px sizes. But I use Edge and Firefox browsers and on both the svg dimensions stay consistent. I did not see any changes on the icon so I left it as it. I'll update it to not use px size.

aamfahim:
>start the video to open in in a new tabs. Then go back to the previous tab to stop the video and then go back t youtube tab to play it

You are right about the fact that the video keeps playing after opening the new tab. Maybe we can pause the video when the new tab is open?

aamfahim:
> You are right about the fact that the video keeps playing after opening the new tab. Maybe we can pause the video when the new tab is open?

@adrienpoly let me know what you think.

https://github.com/user-attachments/assets/978a5061-4b6c-4a19-ba5e-7e9c3ab2d9dc

marcoroth:
I think it's quite hidden if it's just inside the playing video. I wonder if it would make sense to pop it out and make it another button next to the other ones, something like:

CleanShot 2024-11-29 at 19 45 10

aamfahim:
@marcoroth Great Idea actually! Let me know what you think.

https://github.com/user-attachments/assets/efb0dccb-37a2-4978-adba-50a1d919fcfe

app/javascript/controllers/video_player_controller.js
```diff
@@ -61,4 +64,26 @@ export default class extends Controller {
       this.player.seekTo(time)
     }
   }
+
+  createOpenInYouTube () {
+    const videoContainer = document.querySelector('[data-video-player-target="player"]')
```
marcoroth, line 69:
```suggestion
    const videoContainer = this.playerTarget
```

aamfahim: updated with https://github.com/adrienpoly/rubyvideo/pull/467/commits/dbd930f61fd7078cad72793e82adcee2bfd1d89c

app/javascript/controllers/video_player_controller.js
```diff
@@ -61,4 +64,26 @@ export default class extends Controller {
...
+
+    const videoPlayerIcon = document.createElement('button')
+    videoPlayerIcon.className = 'v-openInYouTube v-controlButton'
+    videoPlayerIcon.innerHTML = youtubeSvg
+    videoPlayerIcon.addEventListener('click', handleYouTubePlay)
+
+    const externalPlayButton = document.querySelector('.v-playOnYoutubeButton')
+    if (externalPlayButton) {
+      externalPlayButton.addEventListener('click', handleYouTubePlay)
+    }
```
marcoroth, line 85: Sorry, I have one more thing:

I was just wondering, is there a reason why this couldn't be a regular `<a>` tag? 

```suggestion
    const videoId = this.playerTarget.Container.getAttribute('data-youtube-id')
  
    const videoPlayerIcon = document.createElement('a')
    videoPlayerIcon.className = 'v-openInYouTube v-controlButton'
    videoPlayerIcon.innerHTML = youtubeSvg
    videoPlayerIcon.href = `https://www.youtube.com/watch?v=${videoId}`
    videoPlayerIcon.target = '_blank'
    videoPlayerIcon.dataset.action = "click->video-player#pause"
```

Similarly in the `app/views/talks/_talk.html.erb`, then it wouldn't need to have this dislocated behavior of attaching the onlick callback.

aamfahim: when I was doing the implementation I checked #75 and via inspect element I found that the volume and full screen are implemented as a buttons so that's why I chose the current implementation.

aamfahim: > Similarly in the `app/views/talks/_talk.html.erb`, then it wouldn't need to have this dislocated behavior of attaching the onlick callback.

not sure what you mean here

marcoroth: I think for the other controls it would make sense for them to be button tags, but for a link opening a new website we should be using a tag that's meant for that. 

Doing it as a button also breaks the command+click behaviour for clicks.

So if it possible I would love to have this a a regular `<a>` tag

aamfahim: Of course I like the implementation better as well. Better readability wise too.

However, one issue I'm noticing is that the icon is no longer consistent with the other icons based on screen size. Maybe we can wrap the button with <a> and put the icon inside? I think it's the best of both worlds.

marcoroth: Sure, if that works that's fine too 👍🏼

aamfahim: I cant really test `cmd+click` behavior on my device. Please let me know if you can test the behavior on your side, if it doesn't work I'll update the implementation.

app/javascript/controllers/video_player_controller.js
```diff
@@ -61,4 +64,24 @@ export default class extends Controller {
...
+  playOnYoutube()
+  {
+    const videoId = this.playerTarget.dataset.youtubeId
+    window.open(`https://www.youtube.com/watch?v=${videoId}`, '_blank')
+    this.player.pause()
+  }
+
+  createYouTubeIcon () {
+    const wrappedAnchorElement = document.createElement('a')
+    wrappedAnchorElement.dataset.action = "click->video-player#playOnYoutube"
```
marcoroth, line 77: If you just put the `href` on the `<a>` tag itself it such more predictable and also easier code-wise imo:

```suggestion
  pause() {
    this.player.pause()
  }

  createYouTubeIcon () {
    const videoId = this.playerTarget.dataset.youtubeId
    const wrappedAnchorElement = document.createElement('a')
    wrappedAnchorElement.dataset.action = "click->video-player#playOnYoutube"
    wrappedAnchorElement.href = `https://www.youtube.com/watch?v=${videoId}`
    wrappedAnchorElement.target = "_blank"
```

app/views/talks/_talk.html.erb
```diff
@@ -41,6 +41,15 @@
         </div>
 
         <div class="flex gap-1">
+          <% if talk.video_provider == "youtube" %>
+            <div class="tooltip tooltip-top v-playOnYoutubeButton" data-tip="Watch on YouTube" data-action="click->video-player#playOnYoutube">
```
marcoroth, line 45: Similarly here:
```suggestion
            <a class="tooltip tooltip-top v-playOnYoutubeButton" data-tip="Watch on YouTube"  href="<%= "youtube_url" %> target="_blank">
```

marcoroth:
@aamfahim we almost got it, sorry for the back and forth! 🙌🏼

aamfahim:
@marcoroth all good 😄. I didn't catch you wanted `app/views/talks/_talk.html.erb` to be replaced with <a> as well. Should be good now!

marcoroth (approved):
Awesome, thank you @aamfahim! 🙌🏼

PR #468, opened 2024-11-29
Component page + basic Design system

adrienpoly:
A cheap design system page

It can be helpfull when creating a component to test it in isolation. I didn't wanted to bring something like Lookbook at this point to remain light 

CleanShot 2024-11-29 at 08 58 29@2x


adrienpoly:
in addition to the component page I have started to work a bit at improving our buttons by:
1. reducing the amount of variants
2. adjusting the style

So we would have 5 kinds only primary, secondary, ghost, link and pill 

`pill` buttons would be dedicated to filters and toolbars and replace our current custom buttons 

CleanShot 2024-12-03 at 08 41 39@2x

adrienpoly:
a preview is live on staging 

This idea is to get a much shorter list of buttons with less variations

CleanShot 2024-12-09 at 00 36 37@2x

This PR also:
- set the foundation for a tab component with the tabs Stimulus controller
- adds a basic tooltip component that wraps tippy
- adds an external_link_to helper
- allow to pass options to the remote modal

PR #486, opened 2024-12-04
Sort Talks by ranking and Events by year in Spotlight Search

### LINKED ISSUE #219: Feature Request: Sort Talks Menu search by publication year
First, I'd like to express my appreciation for rubyvideo.dev. It's an invaluable resource for the Ruby community, providing easy access to conference talks.

When using the [search](https://www.rubyvideo.dev/talks?q=hotwire) function to find videos on specific topics, I've noticed that the results are not sorted by the year of publication. This makes it challenging to quickly find the most recent talks on a given subject.

Could we implement a default sorting mechanism that displays the most recent videos first in the search results? This would help users access the latest information

aamfahim:
Resolves #219 

Not sure how it was before but after #407 I noticed on the search modal the talks being shown are over a decade old. I think users would prefer to see the newer talks first.

@marcoroth Also mentioned the same issue with the events as well. So I fixed that as well.

### Talks Before

https://github.com/user-attachments/assets/cac36c6b-0138-4615-a5fd-68c53f9b54d5

### Talks After

https://github.com/user-attachments/assets/7337dbbb-0618-4a20-b4a4-b1a57d4f64c3

### Events Before

https://github.com/user-attachments/assets/fe02caba-c769-4d82-ab1d-36f02e3ff9bc

### Events After

https://github.com/user-attachments/assets/c0a7fb90-733d-4315-a768-c64240df51fe

### DIFF
--- a/app/controllers/spotlight/events_controller.rb
+++ b/app/controllers/spotlight/events_controller.rb
@@ -3,7 +3,7 @@ class Spotlight::EventsController < ApplicationController
   skip_before_action :authenticate_user!
 
   def index
-    @events = Event.includes(:organisation).canonical
+    @events = Event.includes(:organisation).canonical.order(date: :desc)
     @events = @events.ft_search(search_query) if search_query.present?
     @events_count = @events.count
     @events = @events.limit(5)
--- a/app/controllers/spotlight/talks_controller.rb
+++ b/app/controllers/spotlight/talks_controller.rb
@@ -3,7 +3,7 @@ class Spotlight::TalksController < ApplicationController
   skip_before_action :authenticate_user!
 
   def index
-    @talks = Talk.with_essential_card_data
+    @talks = Talk.with_essential_card_data.order(date: :desc)
     @talks = @talks.ft_search(search_query) if search_query.present?
     @talks_count = @talks.size
     @talks = @talks.limit(5)
--- a/app/controllers/spotlight/talks_controller.rb
+++ b/app/controllers/spotlight/talks_controller.rb
@@ -3,8 +3,8 @@ class Spotlight::TalksController < ApplicationController
   skip_before_action :authenticate_user!
 
   def index
-    @talks = Talk.with_essential_card_data.order(date: :desc)
-    @talks = @talks.ft_search(search_query) if search_query.present?
+    @talks = Talk.with_essential_card_data
+    @talks = @talks.ft_search(search_query).ranked if search_query.present?
     @talks_count = @talks.size
     @talks = @talks.limit(5)
     respond_to do |format|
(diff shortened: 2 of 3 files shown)


marcoroth (commented):
Nice one, I also noticed this today and I this makes sense. The same for the events too!

Thanks for this @aamfahim! 🙏🏼

aamfahim:
Thanks @marcoroth. I was just recording before and after for the PR lol. I'll update the events as well!

adrienpoly:
I agree with the implementation for Event.

However, I have some reservations about the Talk model. Currently, the Talk model uses a fuzzy full-text search on the title, description, and speaker names via the Talk::Searchable concern. This concern includes a ranked scope that sorts results based on relevance, assigning more weight to search terms found in the title compared to the description.

That said, I notice that the ranked scope isn’t being applied in the current implementation. I recommend incorporating it to improve result accuracy. Here's an example:

```ruby
@talks = @talks.ft_search(search_query).ranked if search_query.present?
```

Using a date-based ordering might lead to issues. For example, it could prioritize newer talks where the keyword appears only once in the description over older talks where the keyword appears in the title. This could result in less relevant results being ranked higher, which may not be ideal for the user's search experience.


### suggestions to test

for the same bm25 score talks are sorted by dates
```ruby
scope :ranked, -> do
  order(Arel.sql("bm25(talks_search_index, 10.0, 1.0, 5.0) ASC, talks.date DESC"))
end
```


We weight the bm25 score with a number to give more weight for recent talks, the `0.0001` value might needs to be adjusted

```ruby
scope :ranked, -> do
  order(Arel.sql("bm25(talks_search_index, 10.0, 1.0, 5.0) + (strftime('%s', 'now') - strftime('%s', talks.date)) * 0.0001 ASC"))
end
```

aamfahim:
@adrienpoly I agree that it makes more sense to show relevant talks (aka ranked) rather than just the most recent 5 talks.

I've tested the first two suggested scopes. Here are my findings

```ruby
scope :ranked, -> do
  order(Arel.sql("bm25(talks_search_index, 10.0, 1.0, 5.0) ASC, talks.date DESC"))
end
```

With this modification, results aren't much difference from the current implementation BUT it is working, when searching for `active record` there is a difference. So incase of same bm25 score, date is being used as the tie breaker.

```ruby
scope :ranked, -> do
  order(Arel.sql("bm25(talks_search_index, 10.0, 1.0, 5.0) + (strftime('%s', 'now') - strftime('%s', talks.date)) * 0.0001 ASC"))
end
```
With this modification, I'm only getting the most recent 5.

Key words tested: `testing`, `ruby`, `active record`.

adrienpoly:
@aamfahim I played a bit with the weight in #497 hopefully this can help you refine the best ranking strategy

PR #493, opened 2024-12-06
Extract a `retries` helper for performs to clarify intent

kaspth:
The key detail here is the `retry_on StandardError` and that it really means retry on most exceptions.

It seems like callers shouldn't be privy to that implementation detail.

Having a `retries` method that we can pass an attempts argument into, let's our jobs be a bit simpler.

Spotted from #466.


adrienpoly:
oh this is great I love it 👍 will port it to other projets

PR #494, opened 2024-12-06
Swap `queue_as :default` to `:low`

kaspth:
This is really meant to target https://github.com/adrienpoly/rubyvideo/pull/493 but I can't do that from my fork.

Most of our jobs aren't time-critical, and it seems like have to remember to mark most jobs as going to the low queue.

It seems like we'd be better off swapping the default queue to low, and then jobs with higher priority have to mark themselves that.

Note: `AnalyzeTalkTopicsJob` still enqueues with `queue_as :default`, not sure if that should be changed.


adrienpoly:
> Note: `AnalyzeTalkTopicsJob` still enqueues with `queue_as :default`, not sure if that should be changed.

you are right currently nothing is time-critical great to set the default to low. AnalyzeTalkTopicsJob is also a job launched manually from the admin so nothing critical here. will update that later on

PR #495, opened 2024-12-06
Speaker::Profiles: reimplement GitHub profile enhancement + Minisky wrapper

kaspth:
I played around with a bunch of different things here to show what I'd do — and we may not want all of it.

Enhances https://github.com/adrienpoly/rubyvideo/pull/466


app/clients/github/user_client.rb
```diff
@@ -1,5 +1,14 @@
 module GitHub
   class UserClient < GitHub::Client
+    def from_matching(name:)
+      if users = search(name, per_page: 5, page: 1).items
+        names = name.downcase.split(/[ -]/).to_set
+
+        users.lazy.map { profile(_1.login) }.
+          find { names.subset?(_1.name&.downcase&.split(/[ -]/).to_set) }
+      end
```
kaspth, line 9: I haven't really stress tested this reimplementation. It worked with `GitHub::UserClient.new.from_matching name: "Kasper Timm Hansen"` and only issued one request for my profile.

app/models/blue_sky.rb
```diff
@@ -0,0 +1,16 @@
+BlueSky = DelegateClass(Minisky) do
```
kaspth, line 1: Sometimes there's issues with Zeitwerk's reloading and doing something like this but I don't remember what it is exactly. I haven't yet tested it with the app running.

app/models/speaker/profiles.rb
```diff
@@ -0,0 +1,27 @@
+class Speaker::Profiles < ActiveRecord::AssociatedObject
+  performs do
+    retry_on StandardError, attempts: 3, wait: :polynomially_longer # TODO: replace with `retries: 3`.
+    limits_concurrency key: -> { _1.id }
+  end
+
+  def enhance_all_later
+    enhance_with_github_later
+    enhance_with_bsky_later
+  end
```
kaspth, line 10: I know not everybody likes callbacks, but I'm showing what we could also consider doing with ActiveRecord::AssociatedObject:

```suggestion
  # has_object :profiles, after_save: true
  def after_save
    enhance_with_github_later if speaker.github_previously_changed?
    enhance_with_bsky_later   if speaker.bsky_previously_changed?
  end
```

app/models/blue_sky.rb
marcoroth, line 1: TIL about `DelegateClass` 🤯

marcoroth (commented):
I like where this going, thanks for exploring @kaspth!

If I see right, I think it's missing the recent changes from https://github.com/adrienpoly/rubyvideo/pull/479 which might slightly change the approach to some of these changes.

app/controllers/speakers/enhance_controller.rb
```diff
@@ -2,7 +2,7 @@ class Speakers::EnhanceController < ApplicationController
   def update
     @speaker = Speaker.find_by(slug: params[:slug])
 
-    @speaker.profile_enhancer.enhance_all_later!
+    @speaker.profiles.enhance_all_later
```
adrienpoly, line 5: I like this naming!!!

adrienpoly:
ah sorry @kaspth since I merged #466 it closed automatically this branch and I cannot reopen it for whatever reason 
would you mind rebase main and reopen this PR ?

adrienpoly:
ok I managed to reopen it. if you could rebased it from main that would be great

app/models/blue_sky.rb
marcoroth, line 1: I feel like this file should go into `app/clients/` if we want to extract the client.

adrienpoly: I agree and my next question is do we really need this gem ? Is it a simple http get ? Or Something exotic ?

if it is a simple get then we can use the application client

kaspth: I've moved this into `app/clients`. I think we can look at removing `Minisky` in a separate PR.

marcoroth (approved):
Awesome, thank you @kaspth! 🙏🏼

PR #496, opened 2024-12-06
Extract a Talk::TopicAnalysis associated object

kaspth:
Goes onto #466.


kaspth:
@adrienpoly got this one rebased on main.

adrienpoly:
Thanks for this proposal as part of #509 I have extracted all AI related task to an `agents` associated object including the topic extraction. so I am going to close this one. 

Thanks for showing me the way with the `profiles` naming I was trying to follow the recommended pattern ([i]er, -[i]ed and -ion) but sometime plurals makes a lot of sens!

kaspth:
> Thanks for showing me the way with the `profiles` naming I was trying to follow the recommended pattern ([i]er, -[i]ed and -ion) but sometime plurals makes a lot of sens!

I need to update the naming recommendations, I just realized I've seen it in @garrettdimon's post https://garrettdimon.com/journal/posts/organizing-rails-code-with-activerecord-associated-objects via `entitlements` and `seats`.

PR #497, opened 2024-12-06
Ranking test

adrienpoly:
NOT to merge 

This is an exploration PR to follow #486

It displays the bm25 ranking in the card so that we can understand the effect of the weight for the date

adjusting the date weight to DATE_WEIGHT = 0.000000001 seemed to gave good results 

as an example searching for hotwire http://localhost:3000/talks?s=hotwire
give this results

CleanShot 2024-12-06 at 23 58 25@2x

we have all the talk sorted by date with hotwire in the title

then page two after older talks with hotwire in the title we start to see talks without Hotwire in the title (probably found from the description) but from earlier date. So we keep a good relevancy on the title math with some weight for the date 

CleanShot 2024-12-06 at 23 59 45@2x

This PR is not really intended to be merged but more as food for thought for finishing #486

poke @aamfahim

### DIFF
--- a/app/controllers/spotlight/events_controller.rb
+++ b/app/controllers/spotlight/events_controller.rb
@@ -6,7 +6,7 @@ class Spotlight::EventsController < ApplicationController
     @events = Event.includes(:organisation).canonical
     @events = @events.ft_search(search_query) if search_query.present?
     @events_count = @events.count
-    @events = @events.limit(5)
+    @events = @events.limit(8)
     respond_to do |format|
       format.turbo_stream
     end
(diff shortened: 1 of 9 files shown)


aamfahim:
hmm, I went as far as `0.000001` during my testing but didn't really see much difference, but I guess I was missing a couple decimal points lol. But I'm seeing better results with `0.000000001`.

adrienpoly:
> hmm, I went as far as `0.000001` during my testing but didn't really see much difference, but I guess I was missing a couple decimal points lol. But I'm seeing better results with `0.000000001`.

yes this is why I displayed the values tu understand the effect of this weight

PR #498, opened 2024-12-07
Add thumbnails for Child Talks via `ThumbnailExtractor` and `Downloader`

marcoroth:
This pull request adds a `Talk::ThumbnailExtractor` and `Talk::Downloader` which allows a meta talk YouTube video to be downloaded so we can extract thumbnails for each of the segments for the `child_talks` of a Meta Talk.

The video only has to be downloaded once so the person who is committing the new talks can run the `rails download_missing_meta_talks` rake task to download the videos for which the thumbnails are missing and then run the `rails extract_thumbnails` rake task to actually extract the thumbnails.

Now the CI (seed_smoke_test) will fail if somebody forgets to extract the thumbnails, since we now run the `rails verify_thumbnails` rake task in CI to make sure we always have all required thumbnails.

CleanShot 2024-12-07 at 04 39 16


adrienpoly:
Sweet !

PR #509, opened 2024-12-11
Extract talk agents

adrienpoly:
This PR extract all AI related task (improve_transcript, summarize, analyse_topics) to an `agents` associated_object.

In addition prompts are extracted to their own class responsible to build the prompts parameters so that the `Agents` class remains lean


app/models/talk.rb
```diff
@@ -102,17 +101,20 @@ class Talk < ApplicationRecord
...
   # jobs
   performs :update_from_yml_metadata!
+  performs :fetch_and_update_raw_transcript!, retries: 3
+
+  # associated objects
+  # As of now it must be after performs otherwise bin/rails verify_thumbnails will fail
+  # with the error The Talk::Agents associated object referenced from Talk doesn't exist
+  has_object :agents
+  has_object :downloader
+  has_object :thumbnails
```
adrienpoly, line 112: @kaspth I ran into an issue here with associated object

prior to this commit has_object for downloader and thumbnails where defined before the `performs` 
When I added the `agents`  object I had this issue for a rake task


```bash
> bin/rails verify_thumbnails
bin/rails aborted!
The Talk::Agents associated object referenced from Talk doesn't exist
/Users/adrien/code/projects/rubyvideo/app/models/talk.rb:102:in `<class:Talk>'
/Users/adrien/code/projects/rubyvideo/app/models/talk.rb:53:in `<main>'
/Users/adrien/code/projects/rubyvideo/Rakefile:13:in `block in <main>'

Caused by:
NoMethodError: undefined method `agents' for an instance of Integer (NoMethodError)

        value.public_send(...)
             ^^^^^^^^^^^^
/Users/adrien/code/projects/rubyvideo/app/models/talk/agents.rb:2:in `<class:Agents>'
/Users/adrien/code/projects/rubyvideo/app/models/talk/agents.rb:1:in `<main>'
/Users/adrien/code/projects/rubyvideo/app/models/talk.rb:102:in `<class:Talk>'
/Users/adrien/code/projects/rubyvideo/app/models/talk.rb:53:in `<main>'
/Users/adrien/code/projects/rubyvideo/Rakefile:13:in `block in <main>'
Tasks: TOP => verify_thumbnails
(See full trace by running task with --trace)
```

The only solution I found was to move down the has object definition AFTER the `performs` 

is there something we can do to avoid that ?

poke @marcoroth I know you ran into a similar issue I didn't quite understand what was your solution at that time

marcoroth: @kaspth tought me you can extend your Associated Object with `ActiveJob::Performs` for now. Like we did here:

https://github.com/adrienpoly/rubyvideo/blob/244bdc0b593967799525420eb44628b4aee85772/app/models/speaker/profiles.rb#L2

adrienpoly: perfect thanks, I added that and it works perfectly

kaspth: Yeah, I need to fix this in the railtie

PR #518, opened 2024-12-13
update ruby 3.4.0-rc1

adrienpoly:


adrienpoly:
requires https://github.com/ruby/setup-ruby/pull/675

PR #521, opened 2024-12-13
Upgrade ActiveRecord::AssociatedObject to 0.8.3

kaspth:
So we can cut the `extend ActiveJob::Performs` calls and have the gem do it for us.


kaspth:
I also mentioned RubyVideo's usage in the README https://github.com/kaspth/active_record-associated_object?tab=readme-ov-file#rubyvideodev

adrienpoly:
great thanks for refining the DX

PR #525, opened 2024-12-15
Discard talk

### LINKED ISSUE #372: add soft delete for talks
A speaker might want to remove or hide some talks he made in the past.

We should add a way to discard some talks

**A discarded talk:**
- does not appear anymore on the speaker profile page
- does appear in the event index view as grayedout
- remains in the event statistics (talks_count)
- remains in the yearly talk statistics
- doesn't have a #show page
- #show page is removed from the sitemap

 
For now this is an admin feature only, in the future we could make it self service for verified users but that is outside of the scope of this issue

adrienpoly:
this is a first stab at #372

while implementing the feature and in particular trying the discard the talks from Zzak I realised that since a talk can have multiple speaker (https://www.rubyvideo.dev/talks/building-cli-apps-for-everyone) if a talk is discarded then it will be hidden for all speakers of that talk. 

I wonder if we should keep it like that or discard the SpeakerTalk relationship?

close #372

### DIFF
--- a/db/migrate/20241215081736_add_discarded_at_to_talks.rb
+++ b/db/migrate/20241215081736_add_discarded_at_to_talks.rb
@@ -1,0 +1,6 @@
+class AddDiscardedAtToTalks < ActiveRecord::Migration[8.0]
+  def change
+    add_column :talks, :discarded_at, :datetime
+    add_index :talks, :discarded_at
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -157,3 +157,5 @@ gem "omniauth-rails_csrf_protection"
 
 # silence Ruby 3.4 warnings
 gem "ostruct"
+
+gem "discard", "~> 1.4"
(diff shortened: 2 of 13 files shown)


db/migrate/20241215081736_add_discarded_at_to_talks.rb
```diff
@@ -0,0 +1,6 @@
+class AddDiscardedAtToTalks < ActiveRecord::Migration[8.0]
+  def change
+    add_column :talks, :discarded_at, :datetime
```
marcoroth, line 3: I feel like we should call this something like `hidden_at` rather than `discarded_at`, since this feature is more about "hiding" a talk from showing up.

adrienpoly: discarded_at is the attribute used by the [gem discard](https://github.com/jhawthorn/discard) but we could easily do without it with a simple concern. 

For me the main question remains do we hide the talk or hide the SpeakerTalk relation 
If we hide the relation then some things we wanted to do becomes a bit more tricky 
- does appear in the event index view as grayedout (only if all SpeakerTalk are hidden)
- doesn't have a #show page (only if all SpeakerTalk are hidden)

PR #527, opened 2024-12-16
Add location to paris-rb events

Spone:
... or maybe we should stick with "Paris, France" even if less accurate?

### DIFF
--- a/data/paris-rb/playlists.yml
+++ b/data/paris-rb/playlists.yml
@@ -13,7 +13,7 @@
 
 - id: PLjyiiigeVQV-rqM4LD2SUbUfxXLuta1yU
   title: Paris.rb Conf 2020
-  location: "Villejuif, France"
+  location: "Paris, France"
   description: Talks recorded in February 2020 during https://2020.rubyparis.org in Paris.
   published_at: "2020-03-20"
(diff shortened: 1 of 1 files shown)


adrienpoly:
Thanks, 

I would vote for Paris, not sure Villejuif really rings a bell to anyone beside french 😄 

maybe @marcoroth has an opinion on that since he added lots of location for other event.

marcoroth:
I think I agree that Paris makes more sense, since it's also in the conference name 🙈

data/paris-rb/playlists.yml
```diff
@@ -1,6 +1,7 @@
 ---
 - id: PLjyiiigeVQV_mHcSnJnw_1LR4nzGrdF4l
   title: Paris.rb Conf 2018
+  location: "Villejuif, France"
```
Spone, line 4:
```suggestion
  location: "Paris, France"
```

data/paris-rb/playlists.yml
```diff
@@ -12,6 +13,7 @@
 
 - id: PLjyiiigeVQV-rqM4LD2SUbUfxXLuta1yU
   title: Paris.rb Conf 2020
+  location: "Villejuif, France"
```
Spone, line 16:
```suggestion
  location: "Paris, France"
```

Issue #507, opened 2024-12-10
Filter for Talks page


marcoroth:
Either we should have a `Past Talks` and `Upcoming Talks` page or we should have a filter on the `/talks` page, since the first page only show scheduled talks now 😅 

![CleanShot 2024-12-10 at 01 37 50](https://github.com/user-attachments/assets/94173113-17d6-47c5-be29-ac8d53c1c7f7)

adrienpoly:
yes I noticed that and I found it a bit annoying. Technically the controller already [support filters](https://github.com/adrienpoly/rubyvideo/pull/433) it would be fairly easy to add the video_provider. The missing piece is the UI for all those filters. I have started to look at a few options from other site. This is clearly something that I would like to push forward 

For this particular case we could have a video_provider (name?) filter with a default value that filters out the upcoming talks

justinallenmarsh:
@marcoroth @adrienpoly I came up with a filter for this. There is a scope added by #530 which already filters upcoming out.

https://share.cleanshot.com/4flshVxH

Do we want to go with a solution that allows users to filter by what they want specifically? maybe "Published" can be default.

adrienpoly:
It looks great thanks. I would go with the published by default.
The components takes quite a bit of space but right now it is not a big issue. Further down the road we might want to add more filters and we might need more space. Maybe we would have to move to a select at some point ? But let's keep it like that for now.

How does it looks on mobile?

justinallenmarsh:
Here is a link of how it works on mobile.

**Mobile**
https://share.cleanshot.com/kfrgc1sr

**Desktop**
https://share.cleanshot.com/BZctgK6J

I also think it takes up a lot of space, but it fits at the moment like this. I think if we add more filters and select then we can create a modal window.

I changed the wording to align a little bit more with the data model and also used "All" instead of "All Talks" to conserve space.

You can find the PR here. #927

PR #530, opened 2024-12-16
adds a watchable scope

adrienpoly:
relates to #507

This PR adds a Talk scope `watchable` for the video_provider that have watchable content.
This PR defaults the `/talks` view to this scope.
This scope can be bypassed if we pass the params all=true. We would need to see how we convert this into human readable filters and sorting UI

### DIFF
--- a/app/models/talk.rb
+++ b/app/models/talk.rb
@@ -182,7 +182,7 @@ class Talk < ApplicationRecord
   scope :for_topic, ->(topic_slug) { joins(:topics).where(topics: {slug: topic_slug}) }
   scope :for_speaker, ->(speaker_slug) { joins(:speakers).where(speakers: {slug: speaker_slug}) }
   scope :for_event, ->(event_slug) { joins(:event).where(events: {slug: event_slug}) }
-  scope :watchable, -> { where(video_provider: [:youtube, :mp4, :vimeo, :parent]) }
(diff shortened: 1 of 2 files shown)


app/models/talk.rb
```diff
@@ -182,6 +182,7 @@ def formatted_kind
   scope :for_topic, ->(topic_slug) { joins(:topics).where(topics: {slug: topic_slug}) }
   scope :for_speaker, ->(speaker_slug) { joins(:speakers).where(speakers: {slug: speaker_slug}) }
   scope :for_event, ->(event_slug) { joins(:event).where(events: {slug: event_slug}) }
+  scope :watchable, -> { where(video_provider: [:youtube, :mp4, :vimeo, :parent]) }
```
marcoroth, line 185: Something to be aware of: the provider of a `parent` talk can still be `scheduled`, `not_published` or `not_recorded` which is technically not watchable. 

Ideally it would be the union of:
```ruby
Talk.where(video_provider: [:youtube, :mp4, :vimeo])
Talk.joins(:parent_talk).where(parent_talks_talks: { video_provider: [:youtube, :mp4, :vimeo] })
```

But I guess for the sake of simplicity let's just do:
```suggestion
  scope :watchable, -> { where(video_provider: [:youtube, :mp4, :vimeo]) }
```

That way you only get the ones which are really "watchable", which also includes all child talks.

PR #537, opened 2024-12-19
Optimize contribution page

adrienpoly:
This PR extract the contribution content to lazy loaded partials

While the traffic on this route is not huge the response time is really too much with about 2.5s
CleanShot 2024-12-19 at 08 21 02@2x

With this PR the initial page load should be instant and when navigating tabs load only the required content 
We can even cache this content if needed (mostly when they are large collections of cards to render

## Todo :

- [ ] Overdue scheduled talks
- [ ] Talks without speakers
- [ ] Not published talks
- [ ] Add missing events
- [ ] Events without schedule

### DIFF
--- a/.erb_lint.yml
+++ b/.erb_lint.yml
@@ -18,6 +18,7 @@ linters:
       - app/views/contributions/_talks_without_slides.html.erb
       - app/views/contributions/_events_without_location.html.erb
       - app/views/contributions/_events_without_dates.html.erb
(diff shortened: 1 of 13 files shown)


marcoroth (commented):
Awesome, thanks @adrienpoly! I was meaning to do this but you beat me to it 😅

adrienpoly:
merging this first part for now I will address the other partials in another PR

PR #539, opened 2024-12-22
Add an icon with a tooltip about future talks

### LINKED ISSUE #534: Add `with scheduled` label
Does rubyvideo need an update by adding a `with scheduled` label (or something like that) on the Open Analytics page for current and future years? :slightly_smiling_face: 

IMHO, without such label it looks like a bug on the graph :slightly_smiling_face: 

If rubyvideo needs this, I will prepare a PR :slightly_smiling_face: 

image

kalashnikovisme:
- [x] Added an icon with a tooltip about scheduled future talks

Closes #534 

cc @adrianthedev @marcoroth


marcoroth (approved):
Thanks @kalashnikovisme! 🙌🏼

PR #540, opened 2024-12-22
Refactor Talk indexing into a Talk::Index record.

kaspth:
This does take some extra work since Active Record can't detect `rowid` is the primary key.

But with some awareness around that, it's pretty smooth sailing and we can entirely remove a lot of the custom SQL with standard Active Record.

### DIFF
--- a/app/models/talk/index.rb
+++ b/app/models/talk/index.rb
@@ -1,0 +1,37 @@
+class Talk::Index < ApplicationRecord
+  self.table_name = :talks_search_index
+  self.primary_key = :rowid
+
+  attribute :rowid, :integer
+  alias_attribute :id, :rowid
+
+  belongs_to :talk, foreign_key: :rowid
+
+  # SQLite/Active Record doesn't pick up the virtual table primary key by default.
+  # So on direct queries we need to do `Talk::Index.with_rowid.first`, if we want
+  # to have `rowid`/`id` values populated in the result set.
+  scope :with_rowid, -> { select("#{table_name}.rowid, #{table_name}.*") }
+
+  def self.search(query)
+    query = query&.gsub(/[^[:word:]]/, " ") || "" # remove non-word characters
+    query = query.split.map { |word| "#{word}*" }.join(" ") # wildcard search
+    where("#{table_name} match ?", query)
+  end
+
+  def self.snippets(**)
+    COLUMNS_WITH_OFFSETS.each_key.reduce(all) { |relation, column| relation.snippet(column, **) }
+  end
+
+  def self.snippet(column, tag: "mark", omission: "…", limit: 32)
+    select("snippet(#{table_name}, #{COLUMNS_WITH_OFFSETS.fetch(column)}, '<#{tag}>', '</#{tag}>', '#{omission}', #{limit}) AS #{column}_snippet")
+  end
+
+  def reindex
+    update! ALL_COLUMNS.index_with { talk.public_send _1 }
+  end
+
+  private
+
+  ALL_COLUMNS = %i[id title summary speaker_names]
+  COLUMNS_WITH_OFFSETS = ALL_COLUMNS.without(:id).each.with_index.to_h
+end
--- a/app/models/talk/searchable.rb
+++ b/app/models/talk/searchable.rb
@@ -4,40 +4,29 @@ module Talk::Searchable
   DATE_WEIGHT = 0.000000001
 
   included do
-    scope :ft_search, ->(query) do
-      query = query&.gsub(/[^[:word:]]/, " ") || "" # remove non-word characters
-      query = query.split.map { |word| "#{word}*" }.join(" ") # wildcard search
-      joins("join talks_search_index idx on talks.id = idx.rowid")
-        .where("talks_search_index match ?", query)
-    end
+    has_one :index, -> { with_rowid }, foreign_key: :rowid, dependent: :destroy
+
+    scope :ft_search, ->(query) { joins(:index).merge(Talk::Index.search(query)) }
 
     scope :with_snippets, ->(**options) do
-      select("talks.*")
-        .select_snippet("title", 0, **options)
-        .select_snippet("summary", 1, **options)
-        .select_snippet("speaker_names", 2, **options)
+      select("talks.*").merge(Talk::Index.snippets(**options))
     end
 
     scope :ranked, -> do
       select("talks.*,
           bm25(talks_search_index, 10.0, 1.0, 5.0) +
           (strftime('%s', 'now') - strftime('%s', talks.date)) * #{DATE_WEIGHT} AS combined_score")
-        .order(Arel.sql("combined_score ASC"))
+        .order(combined_score: :asc)
     end
 
     after_create_commit :create_in_index
     after_update_commit :update_in_index
-    after_destroy_commit :remove_from_index
   end
 
   class_methods do
-    def rebuild_search_index
-      connection.execute("DELETE FROM talks_search_index")
+    def reindex_all
+      Index.delete_all
       Talk.find_each(&:create_in_index)
-    end
-
-    def select_snippet(column, offset, tag: "mark", omission: "…", limit: 32)
-      select("snippet(talks_search_index, #{offset}, '<#{tag}>', '</#{tag}>', '#{omission}', #{limit}) AS #{column}_snippet")
     end
   end
 
@@ -46,20 +35,10 @@ module Talk::Searchable
   end
 
   def create_in_index
-    execute_sql_with_binds "insert into talks_search_index(rowid, title, summary, speaker_names) values (?, ?, ?, ?)", id, title, summary, speaker_names
+    build_index.reindex
   end
 
   def update_in_index
-    execute_sql_with_binds "update talks_search_index set title = ?, summary = ?, speaker_names = ? where rowid = ?", title, summary, speaker_names, id
-  end
-
-  def remove_from_index
-    execute_sql_with_binds "delete from talks_search_index where rowid = ?", id
-  end
-
-  private
-
-  def execute_sql_with_binds(*statement)
-    self.class.connection.execute self.class.sanitize_sql(statement)
+    index.reindex
   end
 end


app/models/talk/searchable.rb
```diff
@@ -4,62 +4,41 @@ module Talk::Searchable
...
       select("talks.*,
           bm25(talks_search_index, 10.0, 1.0, 5.0) +
           (strftime('%s', 'now') - strftime('%s', talks.date)) * #{DATE_WEIGHT} AS combined_score")
-        .order(Arel.sql("combined_score ASC"))
+        .order(combined_score: :asc)
     end
 
     after_create_commit :create_in_index
     after_update_commit :update_in_index
-    after_destroy_commit :remove_from_index
```
kaspth, line 30: We get this for free now via `dependent: :destroy`!

kaspth:
I don't get what's going on in the tests.

In development the query looks correct and finds the talk:

```
rubyvideo(dev)> Talk.ft_search "Hotwire Cookbook"
  Talk Load (2.3ms)  SELECT "talks".* FROM "talks" INNER JOIN "talks_search_index" ON "talks_search_index"."rowid" = "talks"."id" WHERE (talks_search_index match 'Hotwire* Cookbook*') /* loading for pp */ LIMIT 11 /*application='Rubyvideo'*/
```

But in tests the query looks different and doesn't find the talk:

```
SELECT "talks".* FROM "talks" INNER JOIN "talks_search_index" ON "talks_search_index"."rowid" = "talks"."id" WHERE (talks_search_index match ?)  [[nil, "Hotwire* Cookbook*"]]
```

The `[[nil, "Hotwire* Cookbook*"]]` structure seems to indicate a bind parameter and potentially a prepared statement, but we don't seem to have that turned on/off anywhere. Maybe it's a Rails default somehow?

Swapping to `where("#{table_name} match :query", query:)` also doesn't work. Hm, I'm confused.

kaspth:
Ok, if I put `prepared_statements: false` in `database.yml`, I get the same query as in development:

```
SELECT "talks".* FROM "talks" INNER JOIN "talks_search_index" ON "talks_search_index"."rowid" = "talks"."id" WHERE (talks_search_index match 'Hotwire* Cookbook*')
```

But it still doesn't find anything.

kaspth:
Ah, something seems to be up when we try to insert the record, hm. It's missing the rowid here.

```
INSERT INTO "talks_search_index" ("title", "summary", "speaker_names") VALUES ('yo', '', 'Yauhen Karatkou') RETURNING "rowid" /*application='Rubyvideo'*/
```

kaspth:
Ok, wow, maybe this is more trouble than it's worth. We have to drop back into raw SQL to get Active Record to put in the `rowid` when inserting.

adrienpoly:
I like how it all feels really ActiveRecord.

kaspth:
I'm hoping to get back to this, because I'd like to wrap it up.

app/models/talk/index.rb
```diff
@@ -0,0 +1,27 @@
...
+    index_columns.each_key.reduce(all) { |relation, column| relation.snippet(column, **) }
+  end
+
+  def self.snippet(column, tag: "mark", omission: "…", limit: 32)
+    offset = index_columns.fetch(column)
+    select("snippet(#{table_name}, #{offset}, '<#{tag}>', '</#{tag}>', '#{omission}', #{limit}) AS #{column}_snippet")
+  end
+
+  def reindex
+    update! id: talk.id, title: talk.title, summary: talk.summary, speaker_names: talk.speaker_names
```
kaspth, line 25: We're taking advantage of Active Record automatically generating an update or insert statement based on `new_record?`.

Internally, `update!` calls `create_or_update` which then calls `new_record? ? _create_record(&block) : _update_record(&block)`.

app/models/active_record/sqlite/index.rb
```diff
@@ -0,0 +1,23 @@
+module ActiveRecord::SQLite::Index
+  extend ActiveSupport::Concern
+
+  included do
+    self.ignored_columns = [table_name.to_sym, :rank] # Rails parses our virtual table with these extra non-attributes.
```
kaspth, line 5: A SQLite `fts5` table like this is built on a `rowid` primary key and several columns to index. Active Record now understands these types of virtual tables but it doesn't recognize the `rowid` at all, not even as a column. (It's probably because the column isn't spelled out anywhere in the schema statements so Active Record has no column to parse there). 

So this module encapsulates all the work we need to have Active Record recognize the `rowid` column as a column and as the primary key.

app/models/active_record/sqlite/index.rb
```diff
@@ -0,0 +1,23 @@
...
+    self.primary_key = :rowid
+
+    # Active Record doesn't pick up the `rowid` primary key column.
+    # So we have have explicitly declare this scope to have `rowid` populated in the result set.
+    default_scope { select("#{table_name}.rowid, #{table_name}.*") }
+  end
+
+  private
+
+  def attributes_for_create(attribute_names)
```
kaspth, line 18: We don't need to override `attributes_for_update`, because the `rowid` is included in the `query_constraints_hash` method, that Active Record calls in `_update_record` or wherever it was. The details are a little fuzzy to me now because of all the details I had to do.

app/models/talk/searchable.rb
```diff
@@ -4,62 +4,37 @@ module Talk::Searchable
...
   class_methods do
-    def rebuild_search_index
-      connection.execute("DELETE FROM talks_search_index")
-      Talk.find_each(&:create_in_index)
-    end
-
-    def select_snippet(column, offset, tag: "mark", omission: "…", limit: 32)
-      select("snippet(talks_search_index, #{offset}, '<#{tag}>', '</#{tag}>', '#{omission}', #{limit}) AS #{column}_snippet")
+    def reindex_all
+      Talk::Index.delete_all
```
kaspth, line 27: @adrienpoly reading this again I'm confused as to why we need to delete the existing ones beforehand? Do you remember?

app/models/talk/index.rb
kaspth: This also means we now take advantage of Active Record skipping needless `update` queries when there's no change to the attributes.

app/models/active_record/sqlite/index.rb
kaspth: Oh yeah, and then Active Record also parses the table name and something called `rank` as valid columns, so we need to ignore those since they're internal to the `fts5` implementation.

kaspth:
@adrienpoly @marcoroth think this is ready now. The `ActiveRecord::SQLite::Index` module I added should make this easier to port to the `Speaker`'s index table.

app/models/talk/searchable.rb
adrienpoly: no I don't remember. Maybe I did that initially to just make sure we get something clean knowing that our data is verylimieted so doing a full rebuild is not really costly

kaspth: @adrienpoly ah got it. Yeah, I think with this change we could let Active Record decide and spare some needless queries in case there's no update needed.

kaspth:
```suggestion
```

adrienpoly:
Thanks I will try it on my db and play a bit with it

adrienpoly (approved):
@kaspth I tested it locally and it works great. I pushed a small extra test as I was playing with it. 

Overall I like how we are getting closer to the ActiveRecord native API. It is clearly code that I wouldn't have been able to write. At first I was a bit concerned about maintenance in the future. but now that it is here it seems fairly easy to extend it to other models so I am less concerned about maintenance. 

Looking at ActiveRecord::SQLite::Index code I do see a few hacky things with AR I hope this won't bite us down the road

@marcoroth WDYT ?

Thanks again Kasper for this

test/models/talk_test.rb
```diff
@@ -204,6 +204,24 @@ class TalkTest < ActiveSupport::TestCase
...
+    talk.speakers.create!(name: "Kasper Timm Hansen")
+
+    assert_equal [talk], Talk.ft_search("sqlite full text search") # title
+    assert_equal [talk], Talk.ft_search("ActiveRecord backed virtual table") # summary
+    assert_equal [talk], Talk.ft_search("Kasper Timm Hansen") # speaker
+
+    talk.destroy
+    assert Talk.ft_search("sqlite full text search").empty?
+    assert Talk.ft_search("ActiveRecord backed virtual table").empty?
+    assert Talk.ft_search("Kasper Timm Hansen").empty?
```
kaspth, line 224: We could also do:

```suggestion
    assert_empty Talk.ft_search("sqlite full text search")
    assert_empty Talk.ft_search("ActiveRecord backed virtual table")
    assert_empty Talk.ft_search("Kasper Timm Hansen")
```

kaspth:
> @kaspth I tested it locally and it works great. I pushed a small extra test as I was playing with it.

Perfect!

> Overall I like how we are getting closer to the ActiveRecord native API. It is clearly code that I wouldn't have been able to write. At first I was a bit concerned about maintenance in the future. but now that it is here it seems fairly easy to extend it to other models so I am less concerned about maintenance.

Yeah, I've definitely been worried about this too. I think overall, the savings we're getting from using more standard Active Record and the hoops we have to jump through to get that are less concerning than the raw SQL hoops we were doing before.

And since that hoop logic is condensed in `ActiveRecord::SQLite::Index` it can be shared across models. Or even upstreamed into Rails. I've been wondering if @fractaledmind has thoughts on that part.

I'm also planning on doing a write up about how this works for future reference. I'm quite proud of having figured this out really.

> Looking at ActiveRecord::SQLite::Index code I do see a few hacky things with AR I hope this won't bite us down the road

For sure! I think `attributes_for_create` is the hackiest part and that seems pretty stable to me, although it is technically private API.

marcoroth:
Looks good to me!

fractaledmind:
I'll give this a thorough review tomorrow. Interested to dive in tho

adrienpoly:
> I'll give this a thorough review tomorrow. Interested to dive in tho

hey @fractaledmind I am sure you have been busy with the launch of your course. I wonder if you had time to give it a look. I am interested in merging it but I would love to get your thoughts on this too.

kaspth:
@adrienpoly hey, anything missing on this besides wanting the extra set of eyes?

adrienpoly:
Thanks for this PR. Very interesting code. Let’s ship it 🚀

kaspth:
I'm working on a fix for the test failure on main.

kaspth:
Got it: https://github.com/adrienpoly/rubyvideo/pull/650

Issue #483, opened 2024-12-03
Show duration of talk in talk card


marcoroth:
It would be great to know what the duration of a talk is before clicking on one. 

For the child talks we know the duration because we know about the segments, but for regular talks we could fetch the duration from the YouTube API using a fetch job, similarly to how we handle the transcript and summaries. 

![CleanShot 2024-12-03 at 21 06 07](https://github.com/user-attachments/assets/32581a9a-a099-455f-8643-ef66e2d82339)

nicogaldamez:
I made some progress on this, but I'm not sure where to set the duration (fetched from YouTube) for the talks. Should I include it in the seeds? 

I added `duration_in_seconds` to the Talk model, and it's displayed like this:

![image](https://github.com/user-attachments/assets/bafaee8f-a29e-4dcc-8ed4-744fe26530a3)

You can see the WIP in https://github.com/adrienpoly/rubyvideo/pull/543

PR #543, opened 2024-12-30
Adds duration to talks

nicogaldamez:
### Why
- Issue: #483 
- Users didn’t know how long a talk was until they clicked the Play button.

### What

I added an AVO action to fetch the duration of talks hosted on YouTube. The duration is now displayed on the thumbnail.

image


marcoroth (commented):
Hey @nicogaldamez, thanks for exploring this! I think this is great! 

Re: your question when/where to fetch it: I think we can add another Avo Action to fetch the durations manually, like we do with the [Transcript](https://github.com/adrienpoly/rubyvideo/blob/fe17fcd61c5113ba4dae4baeb0d6673432a96a53/app/avo/actions/transcript.rb) and [Summary](https://github.com/adrienpoly/rubyvideo/blob/fe17fcd61c5113ba4dae4baeb0d6673432a96a53/app/avo/actions/summarize.rb).

adrienpoly:
> Hey @nicogaldamez, thanks for exploring this! I think this is great!
> 
> Re: your question when/where to fetch it: I think we can add another Avo Action to fetch the durations manually, like we do with the [Transcript](https://github.com/adrienpoly/rubyvideo/blob/fe17fcd61c5113ba4dae4baeb0d6673432a96a53/app/avo/actions/transcript.rb) and [Summary](https://github.com/adrienpoly/rubyvideo/blob/fe17fcd61c5113ba4dae4baeb0d6673432a96a53/app/avo/actions/summarize.rb).

+1 for the action

nicogaldamez:
Thanks, @marcoroth! I'll add an Avo Action for this.

I'm sorry if I'm asking something silly, but I'm not familiar with Avo: handling this with an Avo Action implies that an admin user has to manually trigger this action?

marcoroth:
> I'm sorry if I'm asking something silly, but I'm not familiar with Avo

No worries, definitely not a silly questions, thanks for checking!

> handling this with an Avo Action implies that an admin user has to manually trigger this action?

Yes correct, this is how we currently also handle it for the transcripts and summaries. As an admin (you can also update your local user by setting `admin: true` on your `User`) to access Avo at: http://localhost:3000/admin.

marcoroth:
@nicogaldamez is this one ready to go from your side?

nicogaldamez:
Yes, @marcoroth! It's ready. I resolved the conflicts.

marcoroth (approved):
Awesome, thank you so much @nicogaldamez!

I added the duration fields to the Avo resource so we can so the duration more easily in Avo 🙌🏼

PR #548, opened 2025-01-02
remove amara help use caption link

### LINKED ISSUE #535: Remove amara.org from talk descriptions
A lot of talks have an amara.org link in their talks description. I think it doesn't really make sense to have them, especially since the service doesn't seem to be active anymore - at least not for the videos we have.

Example:
```
    Help us caption & translate this video!

    http://amara.org/v/[id]/
```

I think we should remove them.

adrienpoly:
close #535

we will need to update prod accordingly


marcoroth:
> we will need to update prod accordingly

I think the `seed` should take care of that

PR #553, opened 2025-01-04
introduce Rollup to aggregate time series for charts

### LINKED ISSUE #551: pre compute dashboard queries
The dashboard queries are very slow. The good news is that they are computed only once per day and then cached

CleanShot 2025-01-03 at 11 50 23@2x

It would be nice to pre compute them on a regular basis

maybe this could be a good solution https://github.com/ankane/rollup

adrienpoly:
close #551

This PR adds the Rollup gem to create aggregated series for the analytics displays

Currently a POC on the visits and pag view (daily and monthly)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -146,6 +146,7 @@ gem "country_select"
 gem "avo"
 gem "frozen_record", "~> 0.27.2"
 gem "diffy"
+gem "rollups"
 
 # Use OmniAuth to support multi-provider authentication [https://github.com/omniauth/omniauth]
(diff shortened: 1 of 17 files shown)


adrienpoly:
The current Rollup gem has some limitations with Sqlite for dimensions (Postgresql feature only). I have therefore extracted a large part of the gem and started to make it work for Sqlite. It felt a bit too much at this point to open a PR upstream on the gem. I want to first test it fully and maybe in the future try to upstream the Sqlite support

Issue #506, opened 2024-12-10
Fallback for YouTube XL thumbnails 


marcoroth:
Some videos on youtube don't have the XL thumbnail, which then leads to us rendering the generic YouTube thumbnail. I think it be better if we could render the event fallback thumbnail instead.

Though, I'm not sure how we'd be able to detect that. Maybe with an API request?

For example:

![CleanShot 2024-12-10 at 01 33 13](https://github.com/user-attachments/assets/8dd21f31-2c01-4b0a-af37-a91237310716)

https://www.rubyvideo.dev/events/rubyconf-au-2024
https://www.rubyvideo.dev/talks/data-unleashed-a-developer-s-perspective-on-navigating-the-architecture-maze
https://www.rubyvideo.dev/talks/mapping-concepts-into-code

PR #555, opened 2025-01-07
Fallback for YouTube thumbnails

nicogaldamez:
### Why
- Issue #506 
- Some videos on YouTube don't have the XL thumbnail, which then leads to us rendering the generic YouTube thumbnail

### What
Adds a method to check if the thumbnail is the generic YouTube thumbnail. It determines this by checking the image size.
image


adrienpoly:
Thank you for investigating this!

While I haven't tested the solution live, my understanding is that for each talk card, we are querying the YouTube domain to verify the thumbnail image. Unfortunately, this approach might not be viable in production since it could severely impact performance.

Given that only a small subset of talks have incorrect thumbnails, it doesn’t make sense to penalize all talk cards with an additional query.

When I briefly explored the issue, I considered a frontend solution. However, since YouTube URLs return a valid image regardless of correctness, there's no error response we can leverage to display a fallback image easily.

Here’s my current suggestion:
We could create a background task to perform a similar check as your current method but store the result in the model. This way, each talk’s thumbnail is verified only once. For talks without thumbnails, we could periodically rerun the task to see if YouTube has updated their image.

nicogaldamez:
Thanks for your response, @adrienpoly !

Have you thought about using cache to handle this?

adrienpoly:
> Thanks for your response, @adrienpoly !
> 
> Have you thought about using cache to handle this?

I started to look into caching the talk/card yes but this card is not that simple and it does require a bit a rework to improve it's cachability 

if your suggestion is to only cache the thumbnail check I think that would create a n+1 and also have a performance impact for the talk index views

marcoroth:
I was wondering if we can combine this we the API call we would be doing with https://github.com/adrienpoly/rubyvideo/pull/543

adrienpoly:
There are also the view_count and like that we could fetch from the API. This could justify to crawl regularly the YT api and add the thumbnail check if possible 

Part of it was added here #22 but we never really went further than just adding the API client

nicogaldamez:
>I was wondering if we can combine this we the API call we would be doing with https://github.com/adrienpoly/rubyvideo/pull/543 to check if the video has the appropriate thumbnails. Or is there no way to figure that out through the API?

Yes, we can use a HEAD request as it is in the PR

>There are also the view_count and like that we could fetch from the API. This could justify to crawl regularly the YT api and add the thumbnail check if possible

I like this! It will enable adding a filter or sort order to view the more popular talks

marcoroth:
Pushed an updated implementation in 85526900.

Instead of checking thumbnails synchronously on every request, this now runs as a recurring background job that validates YouTube thumbnails weekly. The job checks for the best available thumbnail, preferring 16:9 aspect ratio over 4:3 (which have black bars), and updates both `thumbnail_xl` and `thumbnail_lg` accordingly.

Added a `youtube_thumbnail_checked_at` timestamp so we only recheck thumbnails every 3 months. The job also uses [`ActiveJob::Continuable`](https://edgeapi.rubyonrails.org/classes/ActiveJob/Continuation.html) so it can resume from where it left off if interrupted.

PR #556, opened 2025-01-07
uses ahoy js to prevents bot tracking

adrienpoly:
We recently had lots of fake traffic creating a massive amount of visits 
Ahoy implemements a bot detection feature but backend side it is not always reliable. Using the front end package is [recommended in such cases](https://github.com/ankane/ahoy?tab=readme-ov-file#visits)

let's see if it filters more bot!!!

### DIFF
--- a/app/javascript/entrypoints/application.js
+++ b/app/javascript/entrypoints/application.js
@@ -9,3 +9,10 @@ import '@hotwired/turbo-rails'
 // const channels = import.meta.globEager('./**/*_channel.js')
 
 import '~/controllers'
+
+import ahoy from 'ahoy.js'
+
+ahoy.trackView()
+ahoy.trackClicks('a, button, input[type=submit]')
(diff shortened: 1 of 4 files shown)


app/javascript/entrypoints/application.js
```diff
@@ -9,3 +9,10 @@ import '@hotwired/turbo-rails'
 // const channels = import.meta.globEager('./**/*_channel.js')
 
 import '~/controllers'
+
+import ahoy from 'ahoy.js'
+
+ahoy.trackView()
```
marcoroth, line 15: Do we want to add this into a `turbo:load` event listener so it tracks it in the right way? Or did you do it like this on purpose?

adrienpoly: this is a good point, I used the install recommended by the documentation 
I will look into the source and do a few testing

marcoroth:
Closing this for now

PR #571, opened 2025-01-15
simpler cache key for talk page

### LINKED ISSUE #566: Increase Cache Hit Performance
In the talk show view, the current user is being used in the cache key. I believe this is only used for caching whether a talk is cached or not. This increases the stored cache significantly as a _talk has to be cached for every user who views it. Instead it should use a boolean value to indicate if the talk is watched or not in order for the partial to be cached and shared amongst users.

https://github.com/adrienpoly/rubyvideo/blob/20b2b1a57b6fd823f3c16f6484d877b58e2b56e0/app/views/talks/show.html.erb#L17

adrienpoly:
close #566

poke @jmarsh24

### DIFF
--- a/app/models/watched_talk.rb
+++ b/app/models/watched_talk.rb
@@ -15,6 +15,6 @@
 #  index_watched_talks_on_user_id              (user_id)
 #
 class WatchedTalk < ApplicationRecord
-  belongs_to :user, default: -> { Current.user }
+  belongs_to :user, default: -> { Current.user }, touch: true
   belongs_to :talk
 end
--- a/app/views/talks/show.html.erb
+++ b/app/views/talks/show.html.erb
@@ -13,14 +13,8 @@
         <% end %>
       </div>
 
-      <% if signed_in? %>
-        <% cache_key = [@talk, Current.user, @talk.watched_talks.where(user: Current.user)] %>
-      <% else %>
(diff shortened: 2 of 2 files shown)


jmarsh24 (commented):
There is an array of watched talks, could we instead just cache a true or false value for watched instead of caching the talk for a user. That way the cache is shared amongst all the users instead of per user. 🧐

adrienpoly:
yeah good idea, I look into that this being said I need to check if the user is not somehow required for something else in the _talk partial

app/views/talks/_talk.html.erb
```diff
@@ -1,4 +1,4 @@
-<%# locals: (talk:) -%>
+<%# locals: (talk:, watched:, signed_in:) -%>
```
jmarsh24, line 1: @adrienpoly I believe the watched boolean alone is sufficient for the cache key, as a talk cannot be marked as watched if the user is not signed in. Including signed_in seems redundant in this case.

adrienpoly: if the user is not signed in we currently don't display the button at all. So I still need that information unless we change that and always display the watched button. 

That would be a more significant change as for unsigned user clicking on that button should display the signin/up modale.

adrienpoly:
closing it for now

PR #587, opened 2025-01-19
fix missing watched talks

### LINKED ISSUE #586: Talk Cards Are over-cached?
Marking a video as "watched" doesn't show that in the talk card anymore, like they did in https://github.com/adrienpoly/rubyvideo/pull/418

adrienpoly:
close #586

### DIFF
--- a/app/views/events/show.html.erb
+++ b/app/views/events/show.html.erb
@@ -12,6 +12,7 @@
                 locals: {
                   favoritable: true,
                   user_favorite_talks_ids: @user_favorite_talks_ids,
+                  watched_talks_ids: user_watched_talks_ids,
                   back_to: back_to_from_request,
                   back_to_title: @event.name
                 },
--- a/app/views/talks/index.html.erb
+++ b/app/views/talks/index.html.erb
@@ -17,6 +17,7 @@
             locals: {
               favoritable: true,
               user_favorite_talks_ids: @user_favorite_talks_ids,
+              watched_talks_ids: user_watched_talks_ids,
               back_to: back_to_from_request
             } %>
     </div>


marcoroth (commented):
Thanks @adrienpoly!

I guess we are missing it in some more places too. 

Like: https://github.com/adrienpoly/rubyvideo/blob/83d473fede4616712b34bc1eee9f2f8d3ea6f0a5/app/views/events/show.html.erb#L14

marcoroth:
Does it make sense to move that call into the partial itself? Instead of passing it to all partial calls?

adrienpoly:
I considered moving the logic into the partial, but this might complicate caching in the future. Keeping the props explicit helps identify dynamic elements when constructing the cache key, reducing the risk of missing dependencies.

My proposal is to simplify the card partial by converting the watched state into a boolean (watched). This way, caching the partial would only depend on watched (along with other props), eliminating the need to pass user and user_watch_talk_ids into the cache key.

However, Rails currently lacks built-in support for dynamically evaluating locals when rendering a collection. ([stale proposal](https://discuss.rubyonrails.org/t/feature-proposal-allow-procs-lambdas-in-locals-for-render-calls-on-collections-to-support-dependent-variables/76141))

so we could not do this

```ruby
<%= render partial: "talks/card",
                collection: @talks,
                as: :talk, 
                locals: (talk) -> {
                  favorite: @user_favorite_talks_ids.include?(talk.id),
                  watched: user_watched_talks_ids.include(talk.id)
                } %>
```

We would need to go back to a regular iteration 

```ruby
     <% @talks.each do |talk| %>
        <%= render partial: "talks/card",
                  locals: {
                    talk: talk,
                    favorite: @user_favorite_talks_ids.include?(talk.id),
                    watched: user_watched_talks_ids.include?(talk.id)
                  } %>
      <% end %>
```

PR #600, opened 2025-01-24
Social media profiles to conferences

gorangorun:


marcoroth (commented):
This is great, thanks for exploring @gorangorun! I like where this is headed 🙌🏼 

I just have one nit, naming wise. Somehow I think `Socialable` feels off, but I have no great alternative. Just dropping this here for inspiration: maybe `SocialProfileable`, `Connectable`, or just `HasSocialProfiles`?

It might also just be fine leaving it as-is.

app/models/social_profile.rb
```diff
@@ -0,0 +1,47 @@
...
+class SocialProfile < ApplicationRecord
+  belongs_to :sociable, polymorphic: true
+
+  enum :provider, {
+    github: 0,
+    twitter: 1,
+    linkedin: 2,
+    bsky: 3,
+    mastadon: 4
+  },
```
marcoroth, line 26:
```suggestion
 PROVIDERS = %i[github twitter linkedin bsky mastodon]

  enum :provider, PROVIDERS.zip(PROVIDERS),
```

db/migrate/20250123120534_create_social_profiles.rb
```diff
@@ -0,0 +1,10 @@
+class CreateSocialProfiles < ActiveRecord::Migration[8.0]
+  def change
+    create_table :social_profiles do |t|
+      t.string :value
+      t.integer :provider
```
marcoroth, line 5: I think we might want to add `null: false` to these columns as well as migrating the `provider` to a string so we can see the actual provider value in the database too: 
```suggestion
      t.string :value, null: false
      t.string :provider, null: false
```

app/avo/resources/speaker.rb
```diff
@@ -4,7 +4,7 @@ class Avo::Resources::Speaker < Avo::BaseResource
     if id.is_a?(Array)
       query.where(slug: id)
     else
-      query.find_by(slug: id)
+      query.find_by(slug: id) || query.find_by(id:)
```
adrienpoly, line 7: I wonder why this why needed?

db/migrate/20250123120534_create_social_profiles.rb
gorangorun: Integer gives flexibility if provider names change for some reason, otherwise we will need to update them in db. Anyway, no issue with migrating to string.

app/avo/resources/speaker.rb
gorangorun: Speaker lookup is done based on `:slug`, but when creating association the lookup is done based on id. I'm new to Avo, so couldn't find a better solution. Feel free to suggest a better approach.

adrienpoly: ok thanks. I understand now. Will look if we can have an alternative. this is a minor point

db/migrate/20250123120534_create_social_profiles.rb
adrienpoly: in general in the code base we use string for enums. 

and define them like that 

```ruby
enum :video_provider, %w[youtube mp4 vimeo scheduled not_published not_recorded parent children].index_by(&:itself)
```

marcoroth:
Hey @gorangorun, would you mind to resolve the merge conflicts on this pull request?

marcoroth:
I'm going to close this one, as we are moving more towards having all our data in the YAML files in the `data/` folder. Thanks for exploring this! 🙏🏼

Issue #307, opened 2024-10-27
Add support to specify GitHub handles in seeds to differentiate speakers


marcoroth:
Right now if two speakers have the exact same name we have no way of handling them as individual speakers.

Example: I believe that these two speakers aren't the same person, but share the exact name:

https://www.rubyvideo.dev/speakers/katie-miller

Maybe we can have an annotation in the data files that would allow use to differentiate between the speakers, something like:

```yaml
  speakers:
    - Katie Miller <github:phedinkus>
```

marcoroth:
There are also two "Tom Stuart"'s

* https://www.tomstuart.co.uk
* https://tomstu.art

Ironically, both spoke at Ru3y Manor: https://www.rubyvideo.dev/events/ruby-manor-3/speakers

ChaelCodes:
How difficult would it be to use slug instead of github handle? Slug has to be unique anyway, right?

PR #603, opened 2025-01-25
prevent blank speaker name

adrienpoly:
and add a uniquness validation for the speaker name


marcoroth:
Looks good to me! We might need to revisit this when we look at implementing this one: https://github.com/adrienpoly/rubyvideo/issues/307

adrienpoly:
> Looks good to me! We might need to revisit this when we look at implementing this one: #307

Good point will scope the uniqueness to the combo name/handle

marcoroth:
I don't think scoping it to both the name/handle now makes sense, since that pretty much make it how it was before.

PR #616, opened 2025-01-30
Add `published_at`/`announced_at` columns to `Talk` and use `date` for talk date

### LINKED ISSUE #134: Rename Talk `published_at` to `date` or similar
> I understood `published_at` as "the date the video was published on", but something like `date` would probably be more accurate. 

_Originally posted by @marcoroth in https://github.com/adrienpoly/rubyvideo/issues/113#issuecomment-2293704029_

### LINKED ISSUE #470: Talk model: `date` and `published_at` dates
Current we use the `date` and `published_at` dates on the talk model (and in the `videos.yml` files) interchangeably. 

Ideally we could support both, where `date` is the date on which the talk was held on and `published_at` (probably should actually be `published_on`) on which the video/recording got published.

A `published_on` date could be useful to should in an RSS feed for sorting, or on the landing page to feature "the most recently released recordings", etc.

marcoroth:
This pull request adds a `published_at` and `announced_at` columns to the `Talk` model so we can use the `date` column for the actual Talk date.

This pull request also updates all the YAML files to reorganize the two date fields and adds the missing date respectively.

I'm hoping that we can backfill the remaining missing `published_at` dates from YouTube via the API.

Resolves #134 
Resolves #470


adrienpoly (approved):
great, this will be much more precise than just the current date attribute 

not really sure why some test are failing probably something with the fixtures???

marcoroth:
Yeah, it was mostly because of the new validation that requires a date to be present for a talk.

PR #635, opened 2025-02-04
Add Ruby in London Meetup and artwork

marcoroth:
CleanShot 2025-02-04 at 16 59 00@2x


adrienpoly:
they have a very classy logo!

marcoroth:
It's super simple and straightforward, and it also looks great on the site too! They have a series of events, all with similar logos: https://oliverbernard.com/community

PR #641, opened 2025-02-23
migrate staging server to an AMD64 arch

adrienpoly:
Migrating ton an AMD 64 arch server should speed up the deploy phase as the target arch will be the same as the runner

once everything is ok on staging I intend to do the same for the production server


adrienpoly:
Full build with cold cache takes 5min (vs 20min before)
this is  help also getting a more reliable deploy task I think
CleanShot 2025-02-23 at 09 31 27@2x

PR #647, opened 2025-03-10
Add Vienna.rb March 2025 Videos

hschne:
Videos of the talks from Vienna.rb on 6th of March have been published, adding them here. 

I tried to follow the guidelines [here](https://github.com/adrienpoly/rubyvideo/blob/main/docs/contributing.md), but that completely overwrote the existing data from last Decembers meetup. I ended up manually filling in a bunch of data. It wasn't too bad, I just hope I didn't break something somewhere. 

I also created a playlists for some old talks (see [here](https://www.youtube.com/playlist?list=PLJVqPQvGhwyyfrneIB4MaAjNoxDNaVkZw) ), not sure what the best way to backfill that kind of data is. Any guidance is much appreciated :bow:

### DIFF
--- a/data/vienna-rb/playlists.yml
+++ b/data/vienna-rb/playlists.yml
@@ -1,16 +1,15 @@
-- id: PLJVqPQvGhwyzrZ0fsm00wsLz7JoQHOpHV
-  title: 'Vienna.rb #63 - The Ruby Spring Meetup'
+- id: vienna-rb-meetup
+  title: Vienna.rb Meetup
   location: "Vienna, Austria"
-  description: Talks from the Ruby Meetup on the 6th of March 2025 at Platogo.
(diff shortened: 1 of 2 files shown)


marcoroth (approved):
Hey @hschne, thanks for the pull request.

Yeah, sadly the structure isn't the best one yet. The "playlist.yml" file is just called like that from when they were used as playlists. Now, it would be more appropriate to call them "conference" or "event-series". We wanted to rename them a long time ago, but it's something need to tackle soon.

As for this pull request, I just reverted the changes in the playlist.yml file to match the other meetups, so they don't get out of sync.

For the videos, I just backfilled the `id` key so we have something to match against when upserting the talks in production.

For the future: I'm not sure what the ideal structure is in terms of playlists, but I think what you have now is ideal, where each "event" has it's own playlist with the talks inside it.

PR #648, opened 2025-03-12
Change staging vps and setup litestream as an accessory

adrienpoly:
This PR updates to a new kind of VPS on Hetzner after trying the CX32 Intel based in replacement of the ARM I noticed that the performances were really poor.

CleanShot 2025-03-12 at 05 39 12@2x

This PR updates to a new VPS category using AMD chips 

CleanShot 2025-03-12 at 05 40 56@2x

## Litestream

initially I had Litestream setup using the Litestream Ruby gem. It is easy to setup but I ran into a few edge cases. I also realized that such a backup system is probably a better fit outside of the App. If something crash the app might not be responsive and we might have problem restoring back the db

This PR sets Litestream in a Kamal Accessory


adrienpoly:
A deploy to Staging on a cold cache takes about 3m 50s https://github.com/adrienpoly/rubyvideo/actions/runs/13867486045

The second run doesn't really seems to be faster so I suspect we don't fully uses the docker cache yet?

This is much better than with the ARM arch where (when it works) on a cold cache it was about 20min

PR #655, opened 2025-03-25
Kamal remote builder

adrienpoly:
inspired by work done by @leonvogt, this PR adds a remote builder for Kamal.

The current Arch for the production VPS is ARM. It is slow and unreliable to build and deploy from the CI that runs on AMD arch. I have an extra VPS on Hetzner on ARM that is setup for KAMAL (has docker installed). This PR setup a remote builder for Kamal using this VPS

From the test I did on a warm cache we should be able to deploy in about 1min30s

### DIFF
--- a/config/deploy.yml
+++ b/config/deploy.yml
@@ -41,6 +41,7 @@ builder:
     image: rubyvideo-build-cache
   secrets:
     - RAILS_MASTER_KEY
+  remote: ssh://root@5.75.232.156
 
 env:
   clear:


adrienpoly:
merging it so that we can start using it. @leonvogt, @marcoroth feel free to comment if needed

justinallenmarsh:
@adrienpoly Do you mind sharing how you setup the remote arm server to build on? I tried setting it up in the past and kept running out of space because the images were not getting cleaned up.

adrienpoly:
@jmarsh24 to my knowledge I havn't done anyting crazy. It is just another VPS where Kamal is installed and I setup the ssh key for it.

We are experiencing some deploy/build failure on a regular basis. I was thinking this had to do with my VPS that should be fully reinstalled and cleaned from other projects but beside that I havn't noticed (maybe I didn't really checked) that the VPS is running out of space

PR #669, opened 2025-04-07
Hide `all` button when all speakers is showing up

enderahmetyurt:
https://github.com/user-attachments/assets/9b63ef32-2ee9-46e5-9bf0-29d4ff912dfc


adrienpoly (approved):
I like this idea thanks

Issue #671, opened 2025-04-09
Add an indication that a talk has slides attached


marcoroth:
We have a `slides_url` column on `Talk` that we show as a tab on the talks#show page. It would be nice to indicate that a talk has slides by adding an icon to the talk card.

https://fontawesome.com/icons/presentation-screen

![Image](https://github.com/user-attachments/assets/7275e734-c572-4c4f-aac7-61c59bc91cf0)

cporrast:
@marcoroth submitted a [PR request](https://github.com/rubyevents/rubyevents/pull/673)

PR #673, opened 2025-04-11
Added icon to talk card when slides are available

cporrast:
image
image

Added the icon when the slides_url is not blank

Resolves https://github.com/rubyevents/rubyevents/issues/671


marcoroth (approved):
Thank you @cporrast!

I slightly moved the location of the icon, since we already have the "talk status" in the top left corner.

PR #682, opened 2025-04-25
use an association for meta talk speakers 

### LINKED ISSUE #679: speakers association is defined twice in Talk model
We have two definition for the speakers association on the talk model

The first is the association 

The second is the manual definition that supports meta talk

```ruby
  def speakers
    return super unless meta_talk

    super.to_a.union(child_talks.flat_map(&:speakers).uniq)
  end
```

This causes 2 issues
1. we cannot really eager load this association
2. in Avo when we visit a meta talk it crashes as we return an array not an active record relation

It would be great if we could combine this into a real association

adrienpoly:
close #679

In both case now we return an association. It fixes the crash I was seeing in Avo when visiting a meta talk page.


marcoroth (approved):
Thanks for fixing this, this looks good to me!

PR #685, opened 2025-04-27
Improve transcript prompt

adrienpoly:
This PR adds several improvements to the AI prompts and Agents
- uses by default OpenAI 4.1-nano and mini that offers a 1M context (good for long transcript) and better performances
- I have reworked a bit the prompt
- added an LlmRequest model to store the request and response. This model provides a way to track with which modle and prompt the talk was processed

### DIFF
--- a/app/models/llm/client.rb
+++ b/app/models/llm/client.rb
@@ -1,0 +1,13 @@
+module Llm
+  class Client
+    def initialize(provider_client = OpenAI::Client.new)
+      @provider_client = provider_client
+    end
+
+    def chat(parameters:, resource:, task_name:)
+      LlmRequest.find_or_create_by_request!(parameters, resource: resource, task_name: task_name) do
+        @provider_client.chat(parameters: parameters)
+      end
+    end
+  end
+end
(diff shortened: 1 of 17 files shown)


app/models/llm/client.rb
```diff
@@ -0,0 +1,13 @@
+module Llm
```
marcoroth, line 1: Should we capitalize this constant? 
```suggestion
module LLM
```

app/models/llm_request.rb
```diff
@@ -0,0 +1,51 @@
...
+#  updated_at    :datetime         not null
+#  resource_id   :integer          not null, indexed => [resource_type]
+#
+# Indexes
+#
+#  index_llm_requests_on_request_hash  (request_hash) UNIQUE
+#  index_llm_requests_on_resource      (resource_type,resource_id)
+#  index_llm_requests_on_task_name     (task_name)
+#
+class LlmRequest < ApplicationRecord
```
marcoroth, line 22: Also wondering if this would be better as `LLM::Request` to match the `LLM::Client` constant

PR #686, opened 2025-04-29
SF Ruby: Update February, and add March and April events

irinanazarova:
This is a raw attempt to update the data. 
photos and open mic are to be addressed separately


marcoroth (approved):
Hey Irina, thank you!

I touched up the videos.yml file a bit and added the thumbnails for the different sections!

PR #687, opened 2025-04-29
Add Kaigi on Rails 2020-2023 videos

okuramasafumi:
I followed the instruction at `docs/contributing.md`, but there are some issues left:

- `language` fields are missing, most of them are Japanese but not all
- some `date` fields are missing
- Somehow two keynotes are wrongly places
- Some speaker names are now in Japanese, not possible to make it English automatically
- `slides_url` fields are missing, although we know it as organizers

We have our own YAML file to manage some of these data, so is it a good idea to use it as a data source?


okuramasafumi:
@marcoroth smoke test still fails, and I guess that's because `speakers.yml` cannot contain data with Japanese such as "べーた". Could this be fixed?

marcoroth (approved):
Thanks @okuramasafumi, I cleaned up the data a bit and made the smoke test passing.

CleanShot 2025-05-14 at 15 34 22@2x

PR #698, opened 2025-05-10
Extract Email Verifications & Password Resets to Associated Objects

kaspth:
Right now we've got `password_reset_tokens` and `email_verification_tokens` as Active Record models, however Rails has a built-in `generates_token_for` now that we could use for this.

However, using the built-in would look like this:

```ruby
class User < ApplicationRecord
  generates_token_for :email_verification, expires_in: 2.days

  def self.find_by_email_verification_token(token)  = find_by_token_for(:email_verification, token)
  def self.find_by_email_verification_token!(token) = find_by_token_for!(:email_verification, token)

  def email_verification_token
    generates_token_for :email_verification
  end
end
```

Note: we technically don't need these wrapping methods, however passing the purpose around willy-nilly also doesn't seem great.

Instead this trials an abstraction I've been thinking of for ActiveRecord::AssociatedObject, so we can do:

```ruby
class User < ApplicationRecord
  has_object :email_verification
end

class User::EmailVerification < ActiveRecord::AssociatedObject
  generates_token(expires_in: 2.days) { user.verified? } # Uses purpose `:email_verification` automatically

  def verify
    user.verified! # Technically verification is just an attribute stored on the user.
  end
end

User::EmailVerification.first.token => token
User::EmailVerification.find_by_token(token)
User::EmailVerification.find_by_token!(token)
```

Which means we get objects that we can enrich with further behavior, cutting what the user has to accomplish.

Note: I'm also trialing a `User::Mailer` organization I've been thinking of for a while. So we can have `User.first.mailer # => User::Mailer.with(user:)`.


config/application.rb
```diff
@@ -18,6 +18,18 @@
...
+  def self.generates_token(expires_in:, &)
+    purpose = attribute_name
+    record.generates_token_for(purpose, expires_in:, &)
+
+    define_singleton_method(:find_by_token)  { find_by_token_for(purpose, _1) }
+    define_singleton_method(:find_by_token!) { find_by_token_for!(purpose, _1) }
+
+    define_method(:token) { record.generate_token_for(purpose) }
+  end
+}
```
kaspth, line 31: I can't quite figure out if this whole thing is going too far, so I'm curious to hear your thoughts.

I've also been thinking about whether this should just be `signed_id` delegation instead of tokens.

app/controllers/identity/email_verifications_controller.rb
```diff
@@ -1,28 +1,19 @@
...
   def show
-    @user.update! verified: true
-    redirect_to root_path, notice: "Thank you for verifying your email address"
+    verification = User::EmailVerification.find_by_token(params[:sid])
+
+    if verification
+      verification.verify!
+      redirect_to root_path, notice: "Thank you for verifying your email address"
+    else
+      redirect_to edit_identity_email_path, alert: "That email verification link is invalid"
```
kaspth, line 11: Far as I can tell we omit sign in from `show` but then here we redirect to `identity/emails#show` which requires sign in. So this seems borked.

config/application.rb
```diff
@@ -18,6 +18,18 @@
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
+ActiveRecord::AssociatedObject.class_eval {
```
marcoroth, line 21: Does this need to live in `config/application.rb`? Or could this be in an initializer/concern/superclass somewhere?

config/application.rb
kaspth: Seems like you're in!

config/application.rb
kaspth: Yeah, I thought about that and that could work too. I just pushed a commit putting this in an initializer.

adrienpoly:
heads on this one once #916 will be merge all email/password auth code will be removed so I ll be closing this one. thanks for the exploration

kaspth:
Awesome! Way better to delete code that we don't need.

PR #699, opened 2025-05-10
Refactor sessions

kaspth:
Just hacking around here a bit.

`password_challenge` is a built-in that comes with `has_secure_password`.

Seems like the passwords and identity/emails controllers are no longer used?

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -23,4 +23,10 @@ class ApplicationController < ActionController::Base
     ruby_events_url = Rails.env.production? ? "https://www.rubyevents.org#{path}#{query_string}" : "https://staging.rubyevents.org#{path}#{query_string}"
     redirect_to ruby_events_url, status: :moved_permanently, allow_other_host: true
   end
+
+  def sign_in(user)
+    user.sessions.create!.tap do |session|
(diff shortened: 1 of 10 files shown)


adrienpoly:
> Seems like the passwords and identity/emails controllers are no longer used?

yes in the very first version I had a standard login with username password but now the only way to sign in is with GitHub Oauth. 

If we could keep just that I think it is very handy but @marcoroth might need to more oauth services for the native app (to pass AppStore reviews). This is still un certain at this point.

kaspth:
> If we could keep just that I think

Yeah, I can look at removing the old stuff.

marcoroth:
> but @marcoroth might need to more oauth services for the native app (to pass AppStore reviews). This is still un certain at this point.

Currently we don't have any authentication or any way to sign into the mobile app. And I feel like if we need something more we can build it once we need it.

marcoroth (approved):
Nice, this is great refactor! Thanks @kaspth!

Would you mind resolving the conflicts so we can merge this 🙌🏼

marcoroth:
Thank you @kaspth! 🙏🏼

kaspth:
Oh you spotted it, nice! Thank you for the merge 🪄

PR #708, opened 2025-05-13
Add slugs and YAML IDs to all videos.yml files

marcoroth:
This pull request adds the slugs from production to the YAML files so they are going to be fixed from now on.

Additionally, it adds an `id` key to make sure we are able to identify talks when are updating the video_id throughout the lifecycle of an event.  

For example, when a talk gets added as "scheduled" we don't have the final YouTube ID yet. But as soon as we get that talk recording on YouTube we would update the ID and therefore loose the identifier we are using to match the record in the database with the data in the static YAML file.

Alternative to #646


adrienpoly:
nice 

you might want keep the rake task `yml:dump_talks_slugs` from #646 to get the latest slugs from production

marcoroth:
@adrienpoly I actually tried to use the rake task but it wouldn't fetch all talks from production. It would only get ~4000 out of the 6000+ talks.

adrienpoly:
hum that is strange. Maybe there is a scope where it retrieves only the talks of a certain kind (with youtube video or something like that ???)

marcoroth:
I added the script and the Ruby script I used for dumping the slugs here:

https://github.com/rubyevents/rubyevents/blob/9c712de185eb268a1cc94d4cac06b80282e3b227/yaml/add_slug.mjs#L4-L5

marcoroth:
Resolved via https://github.com/rubyevents/rubyevents/pull/1154 and https://github.com/rubyevents/rubyevents/pull/1155

PR #711, opened 2025-05-14
Add meetup scraper

hschne:
This updates the Vienna.rb and adds a small meetup scraper script, which outputs relevant meetup data as YAML to stdout.  

```bash
ruby scripts/extract_meetup.rb www.meetup.com/vienna-rb/events/307283556/ >> data/vienna-rb/vienna-rb-meetup/videos.yml
```

The YAML is kinda broken if it contains Emojis, which I saw @adrianthedev also tripped over in [#656](https://github.com/rubyevents/rubyevents/pull/656). So I guess we use JS to format YAML now 🤷

I updated and moved the YAML formatting scripts. The YAML formatter now takes optional arguments to format specific video files (as was the case for me).

```bash
npm run format:yml -- data/vienna-rb/vienna-rb-meetup/videos.yml
```

 I had to restructure the npm scripts a bit for that to work. Feel free to rename/tweak/change 🙌


marcoroth:
Just for added context, Hans and I spoke at Baltic Ruby in Riga. We don't have to scrape Meetup if we get to use the GraphQL API like we do on RubyConferences.org:

https://github.com/ruby-conferences/ruby-conferences.github.io/blob/f26d9e575aed88ef049268915e41f4ee7e60188d/src/queries/events_query.rb

PR #718, opened 2025-05-18
Shorten home page title for the Android app

leonvogt:
As discussed with @marcoroth, the Android app will not have a native home screen. At least not in the first version.  
Therefore, the Android app will use the same home page as the web app, including its title.   
Currently, the title is a bit long for a mobile app. This PR would shorten the title a bit when rendering the HTML for a Hotwire Native app.

| Before | After |
|--------|-------|
| Screenshot_before | Screenshot_after |

### DIFF
--- a/app/views/page/home.html.erb
+++ b/app/views/page/home.html.erb
@@ -1,4 +1,4 @@
-<% title "RubyEvents.org - On a mission to index all Ruby events" %>
+<% title hotwire_native_app? ? "RubyEvents" : "RubyEvents.org - On a mission to index all Ruby events" %>
 
 <main class="container hotwire-native:mt-6">
   <div class="hotwire-native:hidden w-full flex flex-col items-center pt-16 pb-12 text-center">


marcoroth (approved):
Thanks @leonvogt! This seems like a good trade-off for now! 🙌🏼

Another alternative we could think about: is there an easy way to just hide the titlebar alltogether in the Home tab?

leonvogt:
> is there an easy way to just hide the titlebar alltogether in the Home tab?

@marcoroth, Sure! We could quite easily hide the titlebar in the Home tab.
We could also add a custom native titlebar for the Home tab, like the old Turbo Native demo app did:

| Home | Every other page |
|--------|-------|
| Screenshot_2025-05-18-18-14-38-84_64b7da8b7fab34afb5db5c585ce11e05 | Screenshot_2025-05-18-18-14-44-53_64b7da8b7fab34afb5db5c585ce11e05 |

PR #727, opened 2025-05-20
Update Tailwind variants to detect Hotwire Native requests server-side

leonvogt:
I've noticed on my Android phone that the Tailwind variants `hotwire-native` and `non-hotwire-native` have a slight delay.
They don't work immediately but take a short moment to activate. This causes a brief flash of unwanted content in the mobile app.
Here’s a screen recording, where you can see the flash of the top banner and navbar briefly appear.

https://github.com/user-attachments/assets/a358d768-c803-4768-891c-5cb1b5bda224

This happens because the `data-bridge-platform` attribute isn't set right away. It's only set once the native code calls `setAdapter` on the HotwireBridge, which happens after the JS code is loaded and the first native component gets registered.
To be honest, I'm not sure why this isn't an issue on iOS. My guess is that iOS/WebKit is a bit more performant, so the JS loads faster.
Maybe you would see the same issue on an older iOS device 🤷

This PR uses the approach of determining mobile / desktop server-side by checking the user agent. 

---

Note: An alternative approach would be to set the necessary data attributes server-side by parsing the user agent.
I've done this in other projects to prevent flashing of bridge components:

```html
<html data-bridge-platform="<%= current_mobile_app.platform %>" data-bridge-components="<%= current_mobile_app.supported_bridge_components %>">
```

But since RubyEvents currently doesn't make use of bridge components, I though it's best to keep it simple.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" <%= 'data-hotwire-native-app' if hotwire_native_app? %>>
   <head>
     <meta name="viewport" content="width=device-width,initial-scale=1<%= ",maximum-scale=1,user-scalable=0" if hotwire_native_app? %>">
 
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en" <%= 'data-hotwire-native-app' if hotwire_native_app? %>>
+<html lang="en" <%= "data-hotwire-native-app" if hotwire_native_app? %>>
   <head>
     <meta name="viewport" content="width=device-width,initial-scale=1<%= ",maximum-scale=1,user-scalable=0" if hotwire_native_app? %>">
 
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -90,8 +90,8 @@ module.exports = {
     require('@tailwindcss/typography'),
     require('daisyui'),
     plugin(function ({ addVariant }) {
-      addVariant('hotwire-native', 'html[data-bridge-platform] &')
-      addVariant('non-hotwire-native', 'html:not([data-bridge-platform]) &')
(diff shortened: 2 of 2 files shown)


app/views/layouts/application.html.erb
```diff
@@ -35,7 +35,7 @@
     <%= yield :head %>
   </head>
 
-  <body data-controller="preserve-scroll">
+  <body class="<%= hotwire_native_app? ? "hotwire-native" : "desktop" %>" data-controller="preserve-scroll">
```
marcoroth, line 38: What if we put a simple `data-hotwire-app` attribute on the `<html>` element itself? Then it would be on the same element as before

marcoroth:
Thanks for the pull request @leonvogt! 

I definitely saw this flash in the iOS app before, so this approach makes sense to me!

leonvogt:
@marcoroth Interesting! Do you remember if you experienced the flashes on iOS when the app was pointing to prod / stage? I was only able to reproduce the flashes when the app was pointing to the development server; unlike Android, which flashes in all environments on every reload.

app/views/layouts/application.html.erb
leonvogt: Sure, sounds good to me! I usually try to use primitive selectors like `.desktop` instead of `:not([data-hotwire-app])` when possible, but I see the benefits of leaving it on the `<html>` element. I'll update the PR.

marcoroth:
@leonvogt it also happens in the app from the app store pointing to production.

At least when you pull-to-refresh. It might also be the time it need to download The stylesheet.

leonvogt:
@marcoroth Yeah, that’s a good guess! Maybe it just comes down to internet speed.

Hmm, the more I think about it, the CSS and JS assets should be loaded from cache when performing a pull-to-refresh… and everything else on the page seems fine too.
I don't quite understand why I can’t reproduce it using the App Store version.
But I guess it doesn’t really matter if we change it to be server-side defined.

marcoroth:
Thanks @leonvogt! 🙌🏼

PR #728, opened 2025-05-20
Prevent overflow in the events header

leonvogt:
Tiny PR that prevents an overflow which could occur in the mobile app when an event has a long URL

| Before | After |
|--------|-------|
| before | after |


marcoroth (approved):
Awesome, I checked it a few times but was never able to figure out what caused it! Thanks @leonvogt! 🙌🏼

PR #735, opened 2025-05-24
Added Yearly conferences and fixing readme

palladius:
1. As mentioned to Marco, added a dashboard with yearly conferences. This can help move the needle in perspective on Ruby livelihood :)
2. Amended README making it explicit that it requires docker. Wasn't obvious to me before.

Technically, I noticed that the other talks dashboard uses a Rollup functionality which is not available for conferences. I presume to make the call much faster, we'd need to populate Rollup with some event aggregation.

**Note**. I want to do a colorful aggregation by continent (EMEA, Americas, JAPAC) but the city info isn't populated. This seems like a bug: eg Toronto (`Event.last`) has all the info in the YAML, yet the city and country are nil. I presume its a bug in the seeding script?


palladius:
How it looks like:

image

PR #740, opened 2025-05-25
Fix tooltip in talk partial, remove tooltip-top class 

Nittarab:
This PR improves the _talk partial using the `ui_tooltip` helper method, where possible, and by removing the tooltip-top class 

Before

CleanShot 2025-05-25 at 12 40 45

After

CleanShot 2025-05-25 at 12 41 15


marcoroth (approved):
Sweet, great one! 🙌🏼 

Thank you @Nittarab! 🙏🏼

PR #744, opened 2025-05-26
Events show page

marcoroth:
| SF Bay Area Ruby | Vienna.rb |
| --- | --- |
| CleanShot 2025-05-26 at 20 34 42@2x | CleanShot 2025-05-26 at 20 34 07@2x |

### DIFF
--- a/app/views/events/events/show.html.erb
+++ b/app/views/events/events/show.html.erb
@@ -32,8 +32,7 @@
           </div>
           <div
             class="absolute inset-0 z-0 h-full w-full bg-[--featured-background] text-[--featured-color]"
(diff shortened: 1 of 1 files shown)


app/views/events/events/show.html.erb
```diff
@@ -0,0 +1,172 @@
+<main class="container">
+  <div class="flex w-full flex-col overflow-hidden bg-[#333333] md:min-h-[calc(100dvh-80px)] rounded-xl" style="transform-origin: 0% 0%; transform: none;">
+    <div class="relative bg-[#333333] text-[--featured-color]" style="--featured-color: <%= @talk.event.featured_color %>;">
```
adrienpoly, line 3: do you think we could name this color bg-[#333333] and define in in tailwind?

marcoroth:
Going to merge this, as it's only accessible via a direct URL for now.

PR #752, opened 2025-05-27
Add tiny ruby #{conf} and reorganize Helsinki Ruby

marcoroth:
This pull request renames the Helsinki Ruby organisation so we can use the organisation for the new tiny ruby #{conf} as well.

CleanShot 2025-05-28 at 00 29 01@2x

CleanShot 2025-05-28 at 00 26 11@2x


marcoroth:
@matiaskorhonen thanks! I forgot to run the seed command in production. Should be fine now 🙌🏼

PR #758, opened 2025-05-28
Skip authentication for `ContributionsController#show`

leonvogt:
Tiny PR that skips the `authenticate_user!` method for the `ContributionsController#show` action.
The motivation is that currently, clicking on some tabs on the contribution page results in a Turbo Frame "Content missing" error, when you're not logged in, since the controller responds with a 302 redirect to `/sign_in`.

Please correct me if some of the show content is not meant to be accessible without authentication and I'm misinterpreting the situation.


marcoroth (approved):
Thanks @leonvogt! 🙌🏼

PR #759, opened 2025-05-28
Add metatags

leonvogt:
While exploring the mobile app, I noticed some screens don’t have native titles.
This PR adds titles to those screens.

### DIFF
--- a/app/models/topic.rb
+++ b/app/models/topic.rb
@@ -81,6 +81,24 @@ class Topic < ApplicationRecord
     canonical || self
   end
 
+  def to_meta_tags
+    {
+      title: name,
+      description: description,
+      og: {
+        title: name,
+        type: :website,
+        description: description,
+        site_name: "RubyEvents.org"
+      },
+      twitter: {
+        card: "summary",
+        title: name,
+        description: description
+      }
+    }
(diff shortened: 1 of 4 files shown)


app/models/topic.rb
```diff
@@ -81,6 +81,24 @@ def primary_topic
...
+      title: name,
+      description: description,
+      og: {
+        title: name,
+        type: :website,
+        description: description,
+        site_name: "RubyEvents.org"
+      },
+      twitter: {
+        card: "summary",
```
leonvogt, line 95: I've set `summary` here instead of `summary_large_image` like in the other existing places, because the topic doesn't have an image. 
I'm not familiar with the meta-tags gems and don't know what would happen if `summary_large_image` is set without an image or what the best practice is in this case.

marcoroth: I think there was a way to let it default to a standard image, but I'm not sure if we have set that up. I don't see anything in the [`config/initializers/meta_tags.rb`](https://github.com/rubyevents/rubyevents/blob/750fe0fee1af9d6bebb248eb0e820a77027553c9/config/initializers/meta_tags.rb) file.

PR #764, opened 2025-05-30
Add Nepal Ruby Meetup March 2025

marcoroth:
CleanShot 2025-05-31 at 00 20 30@2x

/cc @coolprobn


coolprobn:
Thanks @marcoroth 🙏❤️

PR #768, opened 2025-06-01
Add Ruby Romania Meetup, events, talks and artwork

marcoroth:
CleanShot 2025-06-01 at 23 14 23@2x

CleanShot 2025-06-01 at 23 14 31@2x


marcoroth:
Depended on #769

PR #770, opened 2025-06-02
Fill in RubyConf 2022 dates/times

amomchilov:


data/rubyconf/rubyconf-2012/videos.yml
```diff
@@ -135,7 +135,7 @@
   video_id: "l07J_5luC3M"
 
 - title: "Intimate Chat with Matz"
-  raw_title: "Ruby Conf 12 - Intimate Chat with Matz featuring Yukihrio 'Matz' Matsumoto with Evan Phoenix"
```
amomchilov, line 138: This speaker page should be deleted: https://www.rubyevents.org/speakers/yukihrio-matz-matsumoto

amomchilov:
Hmm, I tried `bundle exec rake extract_thumbnails`, but that didn't seem to do anything. `verify_thumbnails` was still failing 🤔

marcoroth:
Hey @amomchilov, thanks for the pull request and for adding the cues for the lightning talks!

Sorry for the lack of docs in that area. The `extract_thumbnails` rake tasks only works if you downloaded the video for that talk beforehand.

The `download_missing_meta_talks` rake task should download the talks that have missing thumbnails using `yt-dlp` into `tmp/videos/youtube/[youtube_video_id].mp4`. Once that file is there it is able to extract the thumbnails using `ffmpeg` with the `extract_thumbnails` task.

Let me know if you want to give it another go, otherwise I'm happy to commit and push the thumbnails to your branch! 🙏🏼

data/rubyconf/rubyconf-2012/videos.yml
marcoroth: Just deleted it 👍🏼 Thanks!

amomchilov:
@marcoroth Ah got it. In that case, should the `download_missing_meta_talks` be made into a dep of `extract_thumbnails`?

I got this working with:

```sh
brew install yt-dlp ffmpeg
bundle exec rake download_missing_meta_talks
bundle exec rake extract_thumbnails
```

marcoroth (approved):
Awesome, thank you so much @amomchilov! 🙏🏼

marcoroth:
> In that case, should the download_missing_meta_talks be made into a dep of extract_thumbnails?

You mean on a rake level? If so, I didn't know that was a thing! TIL. How would that look like and how would we "satisfy" the dependency of `download_missing_meta_talks`  so that `extract_thumbnails` can run?

amomchilov:
@marcoroth Indeed! It's also a dependency of the `environment` task, `download_missing_meta_talks` could be added to that.

https://github.com/rubyevents/rubyevents/blob/a08b2dca4a18ed29e765743e4fb3277aaec0eff2/Rakefile#L58

```diff
- task extract_thumbnails: :environment do |t, args|
+ task extract_thumbnails: [:environment, :download_missing_meta_talks] do |t, args|
```

> How would we "satisfy" the dependency of download_missing_meta_talks

I'm not sure, because the Homebrew deps just aren't tracked like gems are. I have two ideas for this:

1. Rather than shelling out to ffmpeg and yt-dlp, switch to using Rubygems for the same purpose, which bundler can manage for us. (Those gems might very well call ffmpeg/yt-dlp under the hood, but that's not our issue anymore)
2. Try to call those tools, but if they don't exist, just rescue the error and print a friendly error message :)
    * This can fail the first task, so we don't even try to run `extract_thumbnails` (which wouldn't do anything in this case, anyway)

PR #771, opened 2025-06-02
Update hover effect on the talk cards

leonvogt:
This PR removes the hover effect from the talk cards on mobile.

https://github.com/user-attachments/assets/40d5e418-0bd3-4238-8600-3f9706c3fa29

I really like the hover effect on the desktop, it adds a nice touch to the overall experience. 
But in my opinion, it feels a bit out of place on the mobile app.

Totally subjective change, so feel free to close this PR if you’d prefer to keep it!

### DIFF
--- a/app/views/talks/_card.html.erb
+++ b/app/views/talks/_card.html.erb
@@ -12,27 +12,27 @@
     <% if language && language != "English" %>
       <div class="absolute top-0 left-0 z-20 flex p-1 px-2 m-3 rounded-full bg-black/15 backdrop-blur-md">
         <span><%= language_to_emoji(language) %></span>
-        <span class="items-center hidden ml-1 text-sm text-white group-hover:flex"><%= language %></span>
+        <span class="items-center hidden ml-1 text-sm text-white lg:group-hover:flex"><%= language %></span>
       </div>
     <% end %>
 
     <% if talk.scheduled? || talk.parent_talk&.scheduled? %>
       <div class="absolute top-0 right-0 z-20 flex gap-2 p-2 m-3 rounded-full bg-black/15 backdrop-blur-md">
-        <span class="items-center hidden ml-1 text-sm text-white group-hover:flex">Scheduled</span>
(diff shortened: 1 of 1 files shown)


app/views/talks/_card.html.erb
```diff
@@ -46,7 +46,7 @@
     <% if talk.parent_talk %>
       <div class="absolute bottom-0 left-0 z-20 flex content-center gap-2 p-2 m-3 rounded-full bg-black/15 backdrop-blur-md">
         <span><%= fa("rectangle-history", size: :sm, class: "fill-white") %></span>
-        <span class="items-center hidden gap-1 ml-1 text-xs font-normal text-white group-hover:inline-flex line-clamp-1">
+        <span class="items-center hidden gap-1 ml-1 text-xs font-normal text-white group-hover:inline-flex line-clamp-1 hotwire-native:!hidden">
```
leonvogt: With this change, the icon for "part of ..." will still be displayed. Not sure if this might be confusing without any additional text.

Maybe hide the icon itself too? The information that the talk is part of a recording would be visible on the next screen, once you click on the talk.

adrienpoly:
> Totally subjective change, so feel free to close this PR if you’d prefer to keep it!

+1 for this change 

And even for the desktop maybe we could start playing the video on hover like in the [Tailwind Compass template demo ](https://tailwindcss.com/plus/templates/compass/preview) rather than just displaying the play icon (which doesn't really play the video when we click)

app/views/talks/_card.html.erb
marcoroth: I was wondering if we can just always show the text for some of these? Like the "status" and language of the talk would be helpful to be always visible.

Though, I think we can hide the "part of ..." icon.

leonvogt: >  just always show the text for some of these? Like the "status" and language of the talk would be helpful to be always visible.

I totally see that! I think that would be beneficial, I'll update the PR.

PR #772, opened 2025-06-02
Prevent iframe overflow

leonvogt:
This PR prevents an overflow on the "Slides" tab of a talk.

https://github.com/user-attachments/assets/f4fad750-ec16-49af-b5e2-91d21ade8700


adrienpoly (approved):
thanks, looks great

PR #778, opened 2025-06-04
Add XML feed for latest talks and update routes

enderahmetyurt:
This PR introduces an RSS feed for the `/talks` page to enable external consumption of recent Ruby-related talks published on the site.

**What’s Included**
Adds a new feed action in the TalksController

Defines an XML builder view (feed.xml.builder) that generates an RSS 2.0 feed

Limits feed to the 20 most recent talks, ordered by creation date

Adds a <link rel="alternate"> tag in the HTML head for better discoverability by feed readers

**Endpoint**
RSS feed available at: `/talks/feed.xml`

**Why?**
Screenshot 2025-06-04 at 18 18 27
This allows users and feed readers (e.g., podcast apps, Notion widgets, RSS aggregators) to stay updated on new Ruby talks without manually checking the site.

Screenshot 2025-06-04 at 18 13 41

### DIFF
--- a/app/controllers/talks_controller.rb
+++ b/app/controllers/talks_controller.rb
@@ -43,6 +43,13 @@ class TalksController < ApplicationController
     end
   end
 
+  def feed
+    @talks = Talk.order(created_at: :desc).limit(20)
+    respond_to do |format|
+      format.xml
+    end
+  end
+
   private
 
   def order_by
--- a/app/views/talks/feed.xml.builder
+++ b/app/views/talks/feed.xml.builder
@@ -1,0 +1,19 @@
+xml.instruct! :xml, version: "1.0"
+xml.rss version: "2.0" do
+  xml.channel do
+    xml.title "Ruby Events - Talks Feed"
+    xml.link talks_url
+    xml.description "Latest talks in the Ruby ecosystem"
+    xml.language "en-us"
+
+    @talks.each do |talk|
+      xml.item do
+        xml.title talk.title
+        xml.description talk.description
+        xml.pubDate talk.created_at.to_s
+        xml.link talk_url(talk)
+        xml.guid talk_url(talk)
+      end
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


app/controllers/talks_controller.rb
```diff
@@ -43,6 +43,13 @@ def update
     end
   end
 
+  def feed
+    @talks = Talk.order(created_at: :desc).limit(20)
```
marcoroth: We've been adding a lot of old talks and using `created_at` might be a bit misleading since it would also include talks from years ago. 

It might be better to use the `published_at` field instead and make sure we also add the `watchable` scope on it so we only get the talks with videos.

Something like:
```suggestion
    @talks = Talk.past.watchable.order(published_at: :desc).limit(20)
```

marcoroth (commented):
Thanks for opening this pull request and exploring this!

I know there's also this other PR from way back: https://github.com/rubyevents/rubyevents/pull/35

enderahmetyurt:
Thank you @marcoroth I didn't know PR #35 Why didn't you merge it? Should we continue on from this one or there?

adrienpoly:
> Thank you @marcoroth I didn't know PR #35 Why didn't you merge it? Should we continue on from this one or there?

thanks for bringing this back to the surface. Back then it wasn't merged but I can't remember exactly why. any how happy to get this feature back to life

marcoroth:
> Why didn't you merge it?

I think it mostly has to do with which videos to feature and based off the metadata we had it wasn't really possible to select the right talks. I'm still not sure if this pull request now does the right thing and if we have enough valid/good data to pull it off.

I'd imagine there are different use-cases for RSS feeds, which might also be why we want to add multiple feeds:

* Newly announced events
* Newly announced talks
* Recently published recordings of talks

enderahmetyurt:
https://rubyconferences.org/feed.xml has a RSS feed. Maybe we can start with events?

adrienpoly:
I think events could be a great start yes

marcoroth:
Yeah I think so too, let's do that!

enderahmetyurt:
Thank you folks. I'll take care of it 🙌

PR #782, opened 2025-06-05
Add Kaigi on Rails 2025 so that it's listed in upcoming events

okuramasafumi:


okuramasafumi:
Oops, I didn't notice it's already there.

marcoroth:
Thanks @okuramasafumi! 🙏🏼

PR #786, opened 2025-06-09
Optimize events#index and events/past#index

adrienpoly:
This PR adds the start_date and end_date attribute to the event model. This allow to simplify the query and leverable more active record rather than pure ruby sorting filtering. 

### before 
CleanShot 2025-06-09 at 07 30 25@2x

### After 
CleanShot 2025-06-09 at 07 48 00@2x

### before 

CleanShot 2025-06-09 at 07 56 20@2x

### after

CleanShot 2025-06-09 at 07 55 01@2x

further performance optimisation can probably be achieve with caching


adrienpoly:
It wasn't meant to be closed

marcoroth (approved):
Thanks @adrienpoly! 🙌🏼

PR #791, opened 2025-06-13
Add Vienna.rb June Videos

hschne:
What it says on the tin 🪧

Added a new video for the Ruby Quiz, which is not strictly a talk but still nice to share :)


marcoroth (approved):
Awesome, thank you @hschne! 🙏🏼

PR #802, opened 2025-06-16
add task to remove suspicious visits

adrienpoly:
adds a rake task to remove ip having more than 200 visits. This should do a major clean up of the visits generated by bots and largeely reduce the size of our db


enderahmetyurt:
why didn't you use Maintenance Tasks?

adrienpoly:
> why didn't you use Maintenance Tasks?

This gem is not (yet) part of the project. I do use it in some other projects but I didn't feel it was worth it here to bring that gem for that tasks. Might come back with a different point of view if the task crash in prod du to the huge amount of bots visits...

PR #804, opened 2025-06-17
Support granular date precision for event dates

dennispaagman:
I noticed that both Helvetic Ruby and Baltic Ruby 2026 have a date of dec 31, which is probably not the actual date 😜

I also noticed @marcoroth set a 'date_precision' field for it (but that's currently unused) in the yml file:

https://github.com/rubyevents/rubyevents/blob/d887326de64fc81eeaf1cf487b80cec47bb1eb17/data/helveticruby/playlists.yml#L59

This implements this field in the database and the backend and shows the date with different granularities when possible.

Example:

image

Some things I noticed but left out:
- This does not use the `date` field, but the `start_date` field like the method already does in the existing logic. 
- An exact date still needs to be set for the ordering etc to work, I guess setting it to 12-31 or the last or first of the month is fine? Just some explicit knowledge you need to have when configuring the dates.
- This field is not yet set for Baltic Ruby 2026 in the .yml.
- Not sure why, but: Baltic Ruby 2026 doesn't show up in the upcoming events list locally after seeding (the event existis).
- Helvetic Ruby and Baltic Ruby 2026 need to be updated in the admin after this gets deployed.

### DIFF
--- a/app/controllers/analytics/dashboards_controller.rb
+++ b/app/controllers/analytics/dashboards_controller.rb
@@ -23,7 +23,7 @@ class Analytics::DashboardsController < ApplicationController
     #   conferences at the moment. To optimize this code, we probably want to add yearly info to the Rollup class.
     # 2. Note that some years are integers and some are strings. We should probably convert them all to either format in import.
     #   (to reproduce, remove the `.to_s` in the map below)
-    @yearly_conferences = Event.all.select(&:conference?).select { |event| event.start_date.present? || event.year.present? }.group_by { |event| event.start_date&.year || event.year.to_i }.map { |year, events| [year.to_s, events.count] }.sort
+    @yearly_conferences = Event.all.select(&:conference?).select { |event| event.start_date.present? || event.static_metadata.year.present? }.group_by { |event| event.start_date&.year || event.static_metadata.year.to_i }.map { |year, events| [year.to_s, events.count] }.sort
   end
 
   def yearly_talks
--- a/app/models/event/static_metadata.rb
+++ b/app/models/event/static_metadata.rb
@@ -35,6 +35,10 @@ class Event::StaticMetadata < ActiveRecord::AssociatedObject
     event.talks.map(&:date).max
   end
 
+  def date_precision
+    static_repository.date_precision || "day"
+  end
+
   def year
     static_repository.year.present? ? static_repository.year : event.talks.first.try(:date).try(:year)
   rescue => _e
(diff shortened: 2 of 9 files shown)


marcoroth (approved):
Great spot, thanks for working on this @dennispaagman!

> This does not use the date field, but the start_date field like the method already does in the existing logic.

That's fine, we are anyway supposed to remove the `date` column after the work @adrienpoly did in https://github.com/rubyevents/rubyevents/pull/786

> An exact date still needs to be set for the ordering etc to work, I guess setting it to 12-31 or the last or first of the month is fine? Just some explicit knowledge you need to have when configuring the dates.

Yeah that's fine. That's how we did it on RubyConferenes.org too. But maybe we can copy the sentence somewhere:

CleanShot 2025-06-18 at 12 26 33@2x

> This field is not yet set for Baltic Ruby 2026 in the .yml.

Good spot, we can add it later (or you push it up as part of this PR).

> Not sure why, but: Baltic Ruby 2026 doesn't show up in the upcoming events list locally after seeding (the event exists).

This might have to do with the recent changes in #794. Maybe your local copy of Baltic Ruby doesn't have the `kind` column set properly?

> Helvetic Ruby and Baltic Ruby 2026 need to be updated in the admin after this gets deployed.

Sweet, thanks for the heads up!

---

If you can get the build green I'm happy to merge this as-is. Thanks for working on this!

dennispaagman:
> Yeah that's fine. That's how we did it on RubyConferenes.org too. But maybe we can copy the sentence somewhere:
> 
> CleanShot 2025-06-18 at 12 26 33@2x

What would be a good spot? Btw I noticed that this uses `full` as as value, while I opted for `day` in this PR. Want me to change it to that?

> Good spot, we can add it later (or you push it up as part of this PR).
> 
> This might have to do with the recent changes in #794. Maybe your local copy of Baltic Ruby doesn't have the `kind` column set properly?

Fixed the yml file to have both in a separate commit!

> If you can get the build green I'm happy to merge this as-is. Thanks for working on this!

🚀 should go green now!

dennispaagman:
I saw that you added this to the static repository as well, do we still need it on the Event as a database column then?

https://github.com/rubyevents/rubyevents/commit/cf3e8efe4e016575f49381a38add78a704e1aca9

marcoroth:
> I saw that you added this to the static repository as well, do we still need it on the Event as a database column then?

I just added it back because it was erroring out from the refactor in #786 [here](https://github.com/rubyevents/rubyevents/pull/786/files#diff-1870ad0b32594fdf58aab774d5489662bef1b2b279468d7b2ed7b0fb82da0c8c).

marcoroth (approved):
Awesome, thank you @dennispaagman! 🙌🏼

PR #807, opened 2025-06-18
Task to dump talks slug data in a yml file

adrienpoly:
@marcoroth this is an extraction (and fix) from #646 and should be helpful for #708 to avoid having to dump the db every time

When I run this script I now have the same number of talks as running `File.write("slugs.json", Talk.pluck(:video_id, :slug).to_h.to_json)`


data/talks_slugs.yml
marcoroth, line 1: Though, I was wondering if we should really check-in this file. If they are meant to live in the respective `videos.yml` file we don't need them here, right?

adrienpoly: I think it depends if we plan to use it for newly generated content or not ?

What will be the recommended workflow for new content with regards to slugs in the videos.yml.

Do we expect yml files to come with those slugs already built in before we seed? Or we seed first and get the slugs from prod that we insert in the yml files?

marcoroth: My plan is that you can either provide them directly in the YAML file or, if you don't want to pass them, that the `db:seed` command will write them back to YAML file when you are seeding locally.

adrienpoly:
I removed the file from the git

PR #809, opened 2025-06-18
Fix "yearly conferences" chart

leonvogt:
Tiny PR that fixes the "Yearly Conferences" chart.
Currently, it displays a "Content missing" Turbo error because the`Analytics::DashboardsController#yearly_conferences` method raises an exception.
Seems like the `Event#year` method was moved to `Event::StaticMetadata#year` in bec4352.

BTW, lovely chart! I really like this metric and the growth in Ruby conferences!

CleanShot 2025-06-18 at 10 51 28


marcoroth (approved):
Thanks @leonvogt! 🙌🏼

PR #811, opened 2025-06-18
Add YAML template form

hschne:
This PR adds a simple form to generate talk/event YAMLs. It should simplify the data input for new events and make it easier to contribute. Available under `/templates/new`.

- All talk fields are supported 👌
- Events (e.g. Meetups) with multiple child talks are also supported 
- Some basic validations
- Copy button for convenience 😏
- Basic `Claude.md`, because why not 🤖

<details>
<summary>Screenshot</summary>

 image 

</details>

There is more to do to make this super useful, but this should serve as a base for further improvements. For example:
- Right now it's not possible to add multiple 'top-level' talks, as would be needed for most conferences. Haven't figured a way how to do this without making the UI extremely cluttered. 
- Various improvements to data input, e.g. loading from existing YAML, copying talks and child talks, ...
- Better validations and defaults (e.g. inferring event names & video IDs)
- Opening PRs directly from the UI if the user is logged in :thinking:


adrienpoly:
🎉  This is great I had this in mind to. I am pretty sure we can also automate a lot of things with some basic AI. Once we have the youtube channel it should be pretty easy to get the raw content and clean it with a simple AI prompt.

hschne:
Ah yes, I forgot that part in my follow ups, obviously pulling in stuff from YouTube/Meetup is a great improvement 👍

app/views/templates/_talk_fields.html.erb
```diff
@@ -0,0 +1,125 @@
...
+            },
+            class: "btn btn-outline btn-error btn-xs" %>
+    </div>
+
+    <div class="grid grid-cols-1 md:grid-cols-2 gap-4">
+      <div class="form-control">
+        <%= form.label :title, "Title *", class: "label" %>
+        <%= form.text_field :title,
+              class:
+                "input input-bordered w-full #{"input-error" if form.object.errors[:title].any?}",
```
adrienpoly: nit picking here but I usually prefer using the class_name helper for conditional classes 
something like 
```ruby
class_name("input input-bordered w-full", "input-error" => form.object.errors[:title].any?)
```
```

File: app/views/templates/_talk_fields.html.erb [hunk 2]
```diff
@@ -0,0 +1,125 @@
...
+      <div class="form-control">
+        <%= form.label :video_provider, "Video Provider", class: "label" %>
+        <%= form.select :video_provider,
+              [
+                %w[YouTube youtube],
+                %w[Vimeo vimeo],
+                %w[MP4 mp4],
+                ["Not Published", "not_published"],
+                ["Not Recorded", "not_recorded"],
+                %w[Scheduled scheduled]
```
adrienpoly: I wonder if we could pull those values from the model directly ?

adrienpoly:
I did further testing with the tool here are some thoughts 
- speakers it would be great if we could have an autocomplete field (https://hotwirecombobox.com/) with always the ability to add a new one
- this template is dedicated to talks and the route in templates/new. I feel we need something similar at the event level. I enter the event/organisation info, it fetch the playlist from youtube and extract a preview of all talk that I can edit. This being said maybe the routes should be `templates/talks/new`, `templates/events/new`
- if we run it locally maybe we could write directly the yml in addition to copy

hschne:
> I did further testing with the tool here are some thoughts
> 
>     * speakers it would be great if we could have an autocomplete field (https://hotwirecombobox.com/) with always the ability to add a new one
> 
>     * this template is dedicated to talks and the route in templates/new. I feel we need something similar at the event level. I enter the event/organisation info, it fetch the playlist from youtube and extract a preview of all talk that I can edit. This being said maybe the routes should be `templates/talks/new`, `templates/events/new`
> 
>     * if we run it locally maybe we could write directly the yml in addition to copy

Thanks for the feedback! I'll tackle it in a follow up, if you don't mind :)
1. Absolutely, great idea!
2. Yes, agreed. I didn't consider events until now, because for me most of the work unit now was updating individual talks :+1: 
3. Don't think I'd use that but if you think it's useful we can add it. 

Like I said, there's tons of potential for follow ups. I think I'll tackle the speaker auto-complete, events and youtube import next, seems to be the most useful.

adrienpoly:
yes sure that can be follow PR's

adrienpoly:
merging this as a first step. As we discussed they are tons of improvement possible but let's starts with that first

PR #813, opened 2025-06-21
Add XML feed for latest events

enderahmetyurt:
We discussed [here](https://github.com/rubyevents/rubyevents/pull/778) adding some endpoints for the feeds, and we’ve decided to start with the events feed.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -1,7 +1,7 @@
 class EventsController < ApplicationController
   include WatchedTalks
   include Pagy::Backend
-  skip_before_action :authenticate_user!, only: %i[index show update]
+  skip_before_action :authenticate_user!, only: %i[index show update feed]
   before_action :set_event, only: %i[show edit update]
   before_action :set_user_favorites, only: %i[show]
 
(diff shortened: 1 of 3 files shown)


enderahmetyurt:
@adrienpoly @marcoroth Just wondering if you’ve had a chance to take a look at this? 🙏

app/controllers/events_controller.rb
```diff
@@ -36,6 +36,14 @@ def update
     end
   end
 
+  def feed
+    @events = Event.order(date: :desc).limit(20)
+
+    respond_to do |format|
+      format.xml
+    end
+  end
```
marcoroth: Hey @enderahmetyurt, thanks for working on this!

We have been trying to scope the controllers and only use the 7 default actions.

What do you think about moving this to a `Events::FeedController` and have the current logic on the `index` action?

enderahmetyurt: Hello @marcoroth Thank you for your feedback. I see, I will do it.

enderahmetyurt: so @marcoroth @adrienpoly what do you think? How may I continue to work on it? 👀

enderahmetyurt:
it has been in open status for a while. I think I can close this PR for now. Maybe we can look at it in the future.

PR #816, opened 2025-06-22
add caching back

jmarsh24:
I'm not sure why this header is here. The default configuration of turbo is to cache navigation as much as possible. 

I have added it back. This introduces a small bug with modals where turbo caches the modal 

- the user dismisses the modal 
- clicks back, it will display the modal in it's open state.

 I added `data-turbo-temporary` to the modal in order to prevent turbo from caching it when navigating.

https://share.cleanshot.com/Dq9TpfXt

### DIFF
--- a/app/components/ui/modal_component.rb
+++ b/app/components/ui/modal_component.rb
@@ -28,6 +28,7 @@ class Ui::ModalComponent < ApplicationComponent
     attributes[:data] = {
       controller: "modal",
       modal_open_value: open,
+      turbo_temporary: true,
       action: combined_action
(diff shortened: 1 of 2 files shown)


marcoroth:
Here's some context regarding the cache tag, it was set in https://github.com/rubyevents/rubyevents/pull/367 and @leonvogt also recently mentioned it in https://github.com/rubyevents/rubyevents/pull/717.

adrienpoly:
I tested it and everything seems to work fine. I remember that I reworked at some point the view transition. probably the bugs I saw initially with cache were fixed during this rework

PR #819, opened 2025-06-24
Attribute talks to Caleb Hearth

calebhearth:
My name changed back in 2020. All of these talks are by me (in addition to the
correctly attributed ones). I'm fine leaving "Thompson" in the titles and
descriptions as that was properly my name when I gave these talks.

Reference: https://calebhearth.com/why-hearth


adrienpoly:
thanks for sharing this story 👍 

I have assigned Hearth as canonical speaker for Thompson

So https://www.rubyevents.org/speakers/caleb-hearth should have all talks now
and https://www.rubyevents.org/speakers/caleb-thompson is a 301 redirect

PR #821, opened 2025-06-24
Allow fetching a speaker with their talks as JSON

calebhearth:


app/views/talks/_talk.json.jbuilder
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  json.speakers talk.speakers do |speaker|
+    json.id speaker.id
+    json.name speaker.name
+    json.slug speaker.slug
+    if speaker.user.present?
+      json.bio speaker.user.bio
+      json.avatar_url speaker.user.avatar_url
+    end
```
adrienpoly, line 37: I got a 500 here I think those attributes should come directly from the speaker model not the user

app/views/talks/_talk.json.jbuilder
```diff
@@ -0,0 +1,53 @@
...
+  json.topics talk.approved_topics do |topic|
+    json.id topic.id
+    json.name topic.name
+    json.slug topic.slug
+  end
+
+  json.transcript do
+    json.raw talk.raw_transcript
+    json.enhanced talk.enhanced_transcript
+  end
```
adrienpoly, line 49: Those can be pretty large fields. before this PR it was only used for the Talks#show view. 
Now the problem is that for a speaker with lots of Talk (hello Marco/Aaron) it generates a huge json. 

I think we should remove the transcript when we fetch a collection of talks

adrienpoly (commented):
I like the idea to improve our json api 
I do have a few concern about the performance with the current implementation

also if you can try on a user with lots of talks like Marco or Aaron, remove the cache and ensure they are no n+1 (I think they are now)

you might want to remove some of the `present?` https://www.speedshop.co/2019/01/10/three-activerecord-mistakes.html

calebhearth:
Oh whoops I hadn't intended to open a PR yet. Thanks for the input! I'll circle back on this and address that.

hschne:
@calebhearth In 74f29d291 speakers were replaced by profiles. There's also Adrian's comments to address. Would you like to update this PR accordingly?

PR #828, opened 2025-07-03
Add hotwire combobox

hschne:
This is an improvement for #811 as suggested by @adrienpoly. Using [hotwire combobox](https://hotwirecombobox.com/) speaker names are not auto-suggested on typing.

image

- Search uses the existing speaker search queries
- Multi-select and custom input (new speaker names) are supported
- Using names instead of IDs to generate the YAML

### DIFF
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -22,7 +22,25 @@ class TemplatesController < ApplicationController
   def delete_child
   end
 
+  def speakers_search
+    @speakers = Speaker.canonical
+    @speakers = @speakers.ft_search(search_query) if search_query
+  end
+
+  def speakers_search_chips
+    @speakers = params[:combobox_values].split(",").map do |value|
+      Speaker.find_by(id: value) || OpenStruct.new(to_combobox_display: value, id: value)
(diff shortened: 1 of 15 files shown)


app/controllers/templates_controller.rb
```diff
@@ -22,8 +22,26 @@ def new_child
   def delete_child
   end
 
+  def speakers_search
+    @speakers = Speaker.canonical
+    @speakers = @speakers.ft_search(search_query) if search_query
+  end
```
adrienpoly: small suggestion here, when we click on the field there is a first query `Speaker.canonical` returns about 3k+ elements and renders them as options that are turbo streamed to the view. 

while this is not a high traffic page the perf don't really matter but as our number of speaker grow over time this will also increase and create bloat queries.

I think we could limit to a certain amount speakers (200?)  WDYT?

adrienpoly (commented):
really like it thanks for this PR

app/controllers/templates_controller.rb
hschne: Great suggestion and good catch! Will fix 👍

Issue #612, opened 2025-01-28
Bookmark button on talk page


nicogaldamez:
It'd be nice to have the "Add to bookmarks" button below the video:

![Image](https://github.com/user-attachments/assets/409fd698-87e0-4402-b214-da376ac65093)

Users can add a video to bookmarks but only from the talks list.

marcoroth:
Oh, great one!

I'm wondering if this only was ever part of the mockups and never of the app itself 🙈

aronwolf90:
I think we can close this issue.

marcoroth:
Yes, thanks @aronwolf90!

Resolved via https://github.com/rubyevents/rubyevents/pull/830

PR #830, opened 2025-07-07
#612 add bookmark button to talk show page

aronwolf90:
This solves https://github.com/rubyevents/rubyevents/issues/612.

I have right now some time because I am in the transition to a new job and desired to contribute a bit to this nice project. It is also a way for me to update my rails frontend skills that are a bit oxidated. 

NOTE: I have added no test because the it seems that the system tests are in progress for login users.

image

image

One problem that remains is the that the top watch list counter does not adjust. I see here two possible solutions:
* https://github.com/hotwired/turbo/issues/56#issuecomment-1877523211. This would require a new controller.
* Use morph, but then we would have to change a bit more code and add `data-turbo-permanent` to some parts

image


adrienpoly:
thanks this is definitely something I wanted to add. Will take a deeper look at your question for the counter update

aronwolf90:
@adrienpoly I will refactor the page so that morph should work, if this is fine for your. My feeling is that morph is much simpler than using turbo streams.

aronwolf90:
Now it works fine and can be merged from my side:

https://github.com/user-attachments/assets/a0cbe7c5-79ff-4349-914a-969a1954a049

aronwolf90:
So, I solved the conflicts and tested it. All seems to be fine.

adrienpoly:
thanks will test it asap

adrienpoly:
thanks it looks great

aronwolf90:
nice

PR #831, opened 2025-07-07
Add RubyCWB 2nd Edition

eduardoyutaka:
We're about to have the RubyCWB 2nd Edition, a local Ruby Meetup that happens in Curitiba.

The first edition happened in 2024 but we don't have video on it. RubyCWB is one of the official Brazilian Ruby communities: https://community.ruby.com.br/

We plan to make more than 1 event a year, so that's why I'm not categorizing by year but by edition

The event will hapen this Friday, on July 11th and I'll record the talk this time and add it to the recently created YouTube Channel

Let me know if there is any fix I need to do


marcoroth (commented):
Hey @eduardoyutaka, thank you so much for this pull request!

It looks like you need to add the organization to `data/organisations.yml` for it to be picked up by the seeds, which should also fix the failing CI.

marcoroth:
Thanks @eduardoyutaka! 🙏🏼

marcoroth:
If you have the details of the first edition please add them too, even if there is no recording. We can have it as a `video_provider: "not_recorded"` so it's still listed and searchable.

PR #842, opened 2025-07-19
Content: Adds 2011 and 2013 editions of Ruby Midwest

### LINKED ISSUE #836: Ruby Midwest before 2015
We have the 2015 edition on the site, but there is at least one more in 2011

zhephyn:
Fixes #836

### DIFF
--- a/data/ruby-midwest/playlists.yml
+++ b/data/ruby-midwest/playlists.yml
@@ -1,12 +1,44 @@
 ---
 - id: ruby-midwest-2015
   title: Ruby Midwest 2015
-  description: ''
+  description: ""
   location: Kansas City, MO
   kind: conference
-  start_date: '2015-08-28'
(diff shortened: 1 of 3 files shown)


zhephyn:
@marcoroth I decided to add the 2013 edition as well instead of just the 2011 edition which the issue had mentioned originally. 

As for the 2015 edition, sadly I don't think that event ended up happening. So far, as per the googling around and searching You tube plus the conference twitter page, I can only find info linking to only the 2011 and 2013 editions. 

image

Rubyconferences.org shows that the event happened, but I can't find the published videos anywhere.

marcoroth:
Looks like you might be right @zhephyn, I cannot find anything about the 2015 edition. I think you can remove it from the dataset here

zhephyn:
new changes for this are up

PR #853, opened 2025-07-25
adds the ability to sort talks

### LINKED ISSUE #296: Update talks#index default sort
On the front page, the latest talks are shown and a link is provided to "see all talks" When you visit this page, the videos don't match the front page and are static. I propose that this should be the latest talks as that's what I would expect to see and it would be more engaging. A shuffle could be added so that way it's more varied.

CleanShot 2024-10-24 at 23 35 52@2x

When I click the link I would expect to attest see rails world and balkan ruby after clicking this link.

CleanShot 2024-10-24 at 23 36 11@2x

adrienpoly:
adds a way to sort talks by relevance (if we performed a search) or by date (asc, desc) 

CleanShot 2025-07-25 at 08 26 22@2x

Close #296


adrienpoly:
updated the code to remove all custom Js and display the sort menu in a more subtle way 

CleanShot 2025-08-03 at 23 23 53@2x

PR #859, opened 2025-07-29
move playing video to bottom right when scrolling down

adrienpoly:
largely inspired by the UX from https://highleveragerails.com/ and TailwindPlus [Compass Template](https://tailwindcss.com/plus/templates/compass)

The main advantage from a UX point of view is that on larger screen you keep the video visible while being able to read the transcript or read the description 
 
https://github.com/user-attachments/assets/1ee1a555-07a8-4fca-8656-c830905feee2

### DIFF
--- a/app/javascript/controllers/video_player_controller.js
+++ b/app/javascript/controllers/video_player_controller.js
@@ -31,6 +31,7 @@ export default class extends Controller {
   // methods
 
   init () {
+    if (this.isPreview) return
     if (!this.hasPlayerTarget) return
 
     this.player = new Vlitejs(this.playerTarget, this.options)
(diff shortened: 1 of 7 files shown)


app/javascript/controllers/video_player_controller.js
```diff
@@ -122,4 +142,38 @@ export default class extends Controller {
...
+      } else {
+        this.playerWrapperTarget.classList.remove('picture-in-picture')
+        this.playerWrapperTarget.querySelector('.v-controlBar').classList.remove('v-hidden')
+      }
+    }
+
+    // Check if View Transition API is supported
+    if (document.startViewTransition && typeof document.startViewTransition === 'function') {
+      document.startViewTransition(toggleClasses)
+    } else {
```
adrienpoly, line 160: Toggling the class inside the startViewTransition block creates this smooth effect

PR #869, opened 2025-08-03
fix: redirect issue in events search results.

nageswarchedella:
Hi @marcoroth, when I was searching for events happened in a year, the searched results are not shown in a page via  `see results` link. 

Wrong url for events `see results` link. So, I have updated it properly to redirect it to the events archive page.

Hoping this is the right fix.✌️

Before:

https://github.com/user-attachments/assets/1058d3dd-9043-4f4b-85ad-acc00e2733d5

After:

https://github.com/user-attachments/assets/e2b4168e-1d43-424c-8cba-cc108190905a

### DIFF
--- a/app/views/spotlight/events/index.turbo_stream.erb
+++ b/app/views/spotlight/events/index.turbo_stream.erb
@@ -3,7 +3,7 @@
     <h2>Events</h2>
     <% if @events.size.positive? %>
       <span class="text-sm text-gray-500">(<%= @events.size %> of <%= @events_count %>)</span>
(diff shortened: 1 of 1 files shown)


adrienpoly:
Thanks this is clearly a bug. 

I’ll merge it as is for a temporary fix but as the search doesn’t only search in archived events, I thinks we should have a result page that can display all events including not archived

nageswarchedella:
Currently, there is no filter as archived on events. I do see all the events in the archive when I search for something.

adrienpoly:
> Currently, there is no filter as archived on events. I do see all the events in the archive when I search for something.

yes you are right

PR #872, opened 2025-08-04
feature: add connect id to users

adrianthedev:
This PR adds the ability to create accounts based on a connection ID.
It also paves the way to create friendships.

**Claim profile**
CleanShot 2025-08-05 at 00 32 17@2x

**Connect with someone new**
CleanShot 2025-08-05 at 00 31 54@2x

**No profile found**
CleanShot 2025-08-05 at 00 31 29@2x

**Logged in user finding another account**
CleanShot 2025-08-25 at 11 35 12@2x

**Logged in user with a passport account finding a blank connection code**
CleanShot 2025-08-25 at 11 37 29@2x


app/controllers/sessions/omniauth_controller.rb
```diff
@@ -3,12 +3,21 @@ class Sessions::OmniauthController < ApplicationController
...
 
   def create
+    state = query_params["state"]
+    # This needs to be refactored to be more robust when we have more states
+    if state.present?
+      key, value = state.split(":")
+      connect_id = (key == "connect_id") ? value : nil
+      connect_to = (key == "connect_to") ? value : nil
+    end
+
```
adrienpoly, line 13: I remembered having hard time passing state in Google Oauth. I never tried with Github Oauth. Have you tested it in real life. Does the state is preserved during both sign up and sign in ?

app/controllers/profiles_controller.rb
```diff
@@ -0,0 +1,8 @@
+class ProfilesController < ApplicationController
+  skip_before_action :authenticate_user!
+
+  def connect
+    @connect_id = params[:id]
+    @found_user = User.find_by(connect_id: @connect_id)
+  end
```
adrienpoly: when ever we can we try to favor REST route. Maybe something Profiles::ConnectionController either the show route with a singular resource or the new route

app/controllers/sessions/omniauth_controller.rb
adrianthedev: No, I haven't tested tested it with GH in particular.
I was successful in the past with Shopify's Oauth.

I was thinking that we could push this feature and silently test it before we release it to the public.

app/controllers/profiles_controller.rb
adrianthedev: Got it!
I'll do that 🙌

app/controllers/sessions/omniauth_controller.rb
adrienpoly: sure I like that. We also have a staging env where we can do some tests and Github Oauth work in local dev too which is pretty handy

adrienpoly: @adrianthedev this branch is live on staging at : https://staging.rubyevents.org/

adrianthedev: we tested this together 🙌

app/controllers/profiles_controller.rb
adrianthedev: Done!

PR #873, opened 2025-08-05
remove annotate warnings and comments

adrienpoly:
following to https://github.com/drwl/annotaterb/issues/172 this PR removes teh magic comment required in associated objects to silence the annotations warnings

Normally the problem should be fix with the new `annotaterb` gem and `active_record-associated-object` but at the time of this writing when I run the command I still have those warnings

```
❯ bundle exec annotaterb models
Annotating models
Unable to process app/models/event/cfp.rb: undefined method 'cfp' for true
Unable to process app/models/event/schedule.rb: undefined method 'schedule' for true
Unable to process app/models/event/static_metadata.rb: undefined method 'static_metadata' for true
Unable to process app/models/organisation/static_metadata.rb: undefined method 'static_metadata' for true
Unable to process app/models/speaker/profiles.rb: undefined method 'profiles' for true
Unable to process app/models/talk/agents.rb: undefined method 'agents' for true
Unable to process app/models/talk/downloader.rb: undefined method 'downloader' for true
Unable to process app/models/talk/thumbnails.rb: undefined method 'thumbnails' for true
Model files unchanged.
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -49,7 +49,7 @@ GEM
     active_link_to (1.0.5)
       actionpack
       addressable
-    active_record-associated_object (0.9.1)
+    active_record-associated_object (0.9.2)
       activerecord (>= 6.1)
     activejob (8.0.2)
       activesupport (= 8.0.2)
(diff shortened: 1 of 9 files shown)


kaspth:
I just shipped a new release that should fix this https://github.com/kaspth/active_record-associated_object/releases/tag/v0.9.3

I verified the fix on this PR and it shouldn't bail on Associated Objects anymore. There's a separate failure about an enum column not being backed by a database column now though.

adrienpoly:
Thanks @kaspth 

for the enum error I know where this is coming from when you have an old db that didn’t run the migrations from a long time 

I have a fix that I will push in another PR

PR #874, opened 2025-08-06
add base cursor rules

adrienpoly:
add a base cursor rules we can improve it over time with more specific rules. 

Some inspiration can be taken from Maybe repo

### DIFF
--- a/.cursor/rules/cursorrules.mdc
+++ b/.cursor/rules/cursorrules.mdc
@@ -1,0 +1,308 @@
+---
+alwaysApply: true
+---
+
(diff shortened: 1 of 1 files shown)


marcoroth:
I haven't used Cursor Rules or similar yet, but wouldn't it make more sense to have this as an `AGENTS.md` file which would work across these AI tools?

adrienpoly:
The agent.md looks interesting but from my understanding it is not natively supported and you need to make a simlink. Not sure this would be the best DX for an open source project. 

I am not at the state of this art in this domain open to feedbacks and ideas

PR #877, opened 2025-08-07
add opengraph metadata

v19i:


yarn.lock
marcoroth, line 1: I wonder what happened here, since it changes the url from `yarnpkg.com` to `npmjs.org`.

marcoroth (commented):
Thanks for the pull request @v19i! 🙌🏼

yarn.lock
v19i: It's because I have corepack always enabled, I can disable it to revert it back, but it is handy when using multiple projects each with its own package manager.

marcoroth: Got it, no worries, I guess it will eventually get overwritten anyway, thanks! 🙌🏼

PR #881, opened 2025-08-07
serve GitHub picture profile from local asset

adrienpoly:
CleanShot 2025-08-07 at 23 47 18@2x


marcoroth:
Do we really want to commit all these profile pictures to the repo?

marcoroth:
Maybe we can run this on the fly on production and have it save the files on disk and serve them from there instead of actually committing them to the repo.

adrienpoly:
you are right let's do this

PR #883, opened 2025-08-08
style a bit the stamps

adrienpoly:
CleanShot 2025-08-08 at 09 53 48@2x

### DIFF
--- a/app/views/shared/_stickers_display.html.erb
+++ b/app/views/shared/_stickers_display.html.erb
@@ -26,7 +26,7 @@
         <% (9 - events.count).times do %>
           <div class="aspect-square flex items-center justify-center h-24 sm:h-32 md:h-40 lg:h-48">
             <div class="w-full h-full border-4 border-dotted border-gray-300 rounded-xl p-2 bg-white">
-              <div class="w-full h-full rounded-xl bg-gray-100 flex items-center justify-center relative overflow-hidden">
+              <div class="w-full h-full rounded-xl flex items-center justify-center relative overflow-hidden">
               </div>
             </div>
           </div>


marcoroth (commented):
Hey @adrienpoly, nice, this looks awesome!

Though I think the concept of stickers vs. stamps is different in my mind. The stickers we have on the platform right now are trying to mimic (or are actually) the sticker designs that conferences have done in real-life which you would put on your laptop.

With stamps I see them more as a collecting mechanism where you have a stamp card/or a set of stamps of some sort. Or as a mechanism of collecting "achievements" (like "I attended my first Ruby conference", "I attended 5 Ruby conferences", "I connected with a RubyFriend") where you would get a stamp for each of the ones you achieved.

Some conferences had stamp cards where you had to go a visit all sponsors and get their stamp. And the design of those are usually just an outline from an actual stamp.


image

([Credit: Helm App](https://x.com/helm_app/status/1931729608950559106), I shared this image as inspiration a few days ago, which I think is also where this design in this PR is inspired from 🙈 )

TL;DR: While these concepts are similar, I think we should keep them separate in the app and not put the "stickers" on a stamp card like this. But I really like the design, so maybe we can keep both of the "design" in the app?

adrienpoly:
yeah the design is clearly inspired by Helm. Ok I understand better the concept of sticker vs stamps. What confused me at first was maybe the initial design with the border that was somehow looking like a passport border. 

I think this design integrates better into the current style of the app (IMHO). Maybe we could :
- remove the dark border that mimic a passport 
- replace stamps collected by sticker collected 

Once we have Stamps create another page/tabs for that. But I think this is really another complete new subject as creating/assigning those stamps seems like a lot of things to do

marcoroth:
I guess we lost some of the laptop esthetic when I copied it from the [`/stickers`](https://www.rubyevents.org/stickers) page to the speaker profile tab:

CleanShot 2025-08-08 at 11 47 26@2x

But yeah, agreed, the laptop layout isn't ideal as it is right now, especially on the way it looks on mobile. So yeah, maybe we can arranged the stickers on the profile in a more grid-like way like you did here in this PR.

adrienpoly:
got it ! I didn't understood it was a Mac with sticker probably du to the form factor being very different in the tab. 

Going off topic here but most probably we will want to change the navigation. On some profiles this give I think too much tabs.

Maybe one tab only for videos but with filters so that we can select the kind of talk rather than having all those tabs for Q&A etc

marcoroth:
We are using the design from this pull request in https://github.com/rubyevents/rubyevents/blob/93317aad825e6639d03c305e33ac3f710f0b06e6/app/views/shared/_stamps_grid.html.erb, so we can close this one.

PR #890, opened 2025-08-10
Refactoring TalksAgent to use active agent

TonsOfFun:
Refactoring TalksAgent to use active agent's action prompt for action… view style prompt message template rendering


app/views/talks_agent/analyze_topics.text.erb
```diff
@@ -0,0 +1,53 @@
+
+You are tasked with figuring out of the list of provided topics matches a transcript based on the transcript and its metadata. Follow these steps carefully:
+
+1. First, review the metadata of the video:
+<metadata>
+  - title: #{title}
```
marcoroth, line 6: These interpolations won't work in ERB files

app/views/talks_agent/summary.json.erb
```diff
@@ -0,0 +1,22 @@
+<%= {
```
marcoroth, line 1: Is there a reason why this is in ERB and in a separate file if it's just a Ruby hash anyway?

app/agents/talks_agent.rb
```diff
@@ -0,0 +1,47 @@
...
+  def update_summary
+    raw_response = JSON.repair(response.dig("choices", 0, "message", "content"))
+    summary = JSON.parse(raw_response).dig("summary")
+    talk.update!(summary: summary)
+  end
+
+  def update_topics
+    raw_response = JSON.repair(response.dig("choices", 0, "message", "content"))
+    topics = begin
+      JSON.parse(raw_response)["topics"]
```
marcoroth, line 37: I feel like this should either be directly handled in the framework or maybe the schema needs to be updated?

Isn't the whole idea of structured output that it gives you JSON back?

app/views/talks_agent/analyze_topics.text.erb
TonsOfFun: Indeed this needs to be updated and tested still, thanks for pointing this out.

app/agents/talks_agent.rb
TonsOfFun: Yeah, this also needs testing, I mostly wanted to show how it will look overall compared to how it was before. That way I could get your feedback before spending too much time on it.

app/views/talks_agent/summary.json.erb
TonsOfFun: Separate file to separate of concerns, but doesn't have to be erb, though I have been using erb to render things like action names dynamically.

adrienpoly:
Hello 

thanks for working on this proposal. Just a quick heads up that this part of the code has been working quite well for us up to now and I kind of liked the combination of ActiveJob-performs and associated object we ended up using. 

At this point of the PR I don't really see a clear benefits of introducing a new dependency/framework.  I am following ActiveAgent development I feel like it is made for more complex Agent workflows and we are not there yet our usage is very simple and I think will remain simple for quite some time. 

If we where to introduce a new framework I would prefer that this happens on some new feature where it really brings something new to the table

but happy to be convinced otherwise

TonsOfFun:
I appreciate your perspective and the old adage of don't fix what isn't broken certainly applies to your sentiments. 

I was planning to start with this conversion/refactor as a POC for what you already have as an example of how active agent could be used to make even simple use cases more concise.   

The ultimate goal would be to continue on to having more advanced agent feature sets for collecting talks, events, and sponsors data. 

Let me know if you'd like to discuss this proposal/draft PR more or if there is any interest in continuing down this path.

marcoroth:
Closing this for the time being, thanks for your pull request 🙌🏼

PR #894, opened 2025-08-11
Transfer user to speaker table

adrienpoly:
This is a first step toward merging User and Speaker tables

In this PR I add all the missing fields of the User table in the Speaker table and add a task to start migrating User to Speaker table

In a follow up PR we will 
- drop the user table
- update all association pointing to user to point to speaker table
- rename speaker to user


marcoroth (approved):
Looks good to me!

I assume that we are going to update and match the `ConnectedAccount` associations on the GitHub handle? If so, then this should be fine!

Also, I think we should rename `speakers` back to `users`, once we dropped our current `users` table.

adrienpoly:
> Looks good to me!
> 
> I assume that we are going to update and match the `ConnectedAccount` associations on the GitHub handle? If so, then this should be fine!
> 
> Also, I think we should rename `speakers` back to `users`, once we dropped our current `users` table.

yes this is the very first step toward the final goal of having just a User table

adrienpoly:
currently looking at doing it the other way round

PR #908, opened 2025-08-13
Enhance Sponsor Avatar and Logo Handling with Fallbacks

DiegoPisa2003:
Overview

This PR improves the visual consistency and reliability of sponsor displays by implementing fallback systems for both avatars and logos. It addresses cases where external logo URLs become inaccessible and ensures all sponsors have proper visual representation.

### Enhanced Avatar System
- **UI Avatars Integration**: Added `generate_avatar_url` method that creates dynamic avatars using [ui-avatars.com](https://ui-avatars.com) API
- **Smart Avatar Fallback**: Modified `avatar_image_path` to fallback to generated avatars when local images aren't available
- **Priority System**: Local images > logo fallback > generated avatars

### Robust Logo URL Handling
- **URL Accessibility Validation**: Added `logo_url_accessible?` method that checks if external logo URLs are actually reachable
- **Network Resilience**: Implements timeout handling (5s open, 10s read) and graceful error handling
- **Logo Fallback**: When logo URLs are inaccessible, automatically falls back to avatar images instead of broken links

## Future enhancement

I was thinking that for the `logos_url` that are inaccessible, we could mark them someway (maybe with a boolean) as inaccessible in a way that the next time we the app go through the `logo_url_accessible?` can see that boolean and does not have to do a HTTP request to see if it is accessible. 

We would have to take into consideration if we implement that, that maybe the `logo_url` was momentarily inaccessible and then became accessible again, so periodically the app should try again to access the `logo_url`

Even if we do not merge this PR, I would like to set the record that we should somehow manage inaccessible logo_url in the backend, not only in the views

### DIFF
--- a/app/models/sponsor.rb
+++ b/app/models/sponsor.rb
@@ -61,11 +61,21 @@ class Sponsor < ApplicationRecord
   end
 
   def avatar_image_path
-    sponsor_image_or_default_for("avatar.webp")
+    if sponsor_image_for("avatar.webp")
+      sponsor_image_or_default_for("avatar.webp")
+    else
+      generate_avatar_url
+    end
   end
 
   def banner_image_path
     sponsor_image_or_default_for("banner.webp")
+  end
+
+  def generate_avatar_url(size: 200)
+    url_safe_initials = name.split(" ").map(&:first).join("")
+
+    "https://ui-avatars.com/api/?name=#{url_safe_initials}&size=#{size}&background=f3f4f6&color=4b5563&font-size=0.4&length=2"
   end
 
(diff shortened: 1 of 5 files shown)


marcoroth:
We handle this on the client-side now, so we fall back to the initials when the image fails to load. See: https://github.com/rubyevents/rubyevents/pull/930

The reason why the approach in https://github.com/rubyevents/rubyevents/pull/930 is preferable is because we don't have to make a request on the server-side just to check if the image is accessible. That said, I appreciate the pull request, @DiegoPisa2003! Thank you! 🙌🏼

PR #917, opened 2025-08-15
Use Rails main

adrienpoly:
Good or bad idea?

We are not doing crazy things so we should definitely be able to run main, yet our tests suite is not  extensive


marcoroth (approved):
Let's do it!

PR #918, opened 2025-08-16
refactor: use YouTube player when available

adrianthedev:
Reverting to the YouTube player when available


marcoroth:
I'm all for this change, thanks!

While testing it, it seems like it only shows the controls when you hover the video in the bottom 20% of the player. Do you have an idea why?

https://github.com/user-attachments/assets/59f69d88-45fa-447a-9598-eee6e6f17c8e

marcoroth:
We might also want to set the [`rel` argument](https://developers.google.com/youtube/player_parameters) to disable the related videos:

CleanShot 2025-08-16 at 14 20 27@2x

CleanShot 2025-08-16 at 14 20 44@2x

adrienpoly:
I tested it, I am all in too to use Youtube default Player

There is yet an issue it that currently it breaks the page view transition effect and we have this error in console

CleanShot 2025-08-17 at 08 49 40@2x

This is pretty common error whenever a page view transition name is duplicated on a page it completely disable all transition on that page

What is strange is that when I inspect the dom the name is not duplicated. So it looks like VliteJs is doing something differently in that case and it creates a temporary duplicate. Not sure how to debug fix that though

adrianthedev:
Yeah, the show controls on hover behavior is somehow broken from what I can see. I'll look into it.

The `rel` attribute has [changed](https://developers.google.com/youtube/player_parameters#release_notes_08_23_2018:~:text=The%20behavior%20for%20the%20rel%20parameter%20is%20changing%20on,same%20channel%20as%20the%20video%20that%20was%20just%20played.). I'll set it to `0` to show only content from that channel.


I haven't hit the duplicated transition issue.
Are there specific steps I can take?

adrianthedev:
Figured the hover bug. Vlite has an overlay which was messing wiht the default hover state. Fixed that.

I could reproduce the view transition error...

adrienpoly:
tested it again and it works great not sure why I had this error let's ship it and we can fix later if we see something

marcoroth:
Thank you @adrianthedev! 🙏🏼

PR #919, opened 2025-08-17
Merge speaker and user tables

adrienpoly:
This PR replaces #894 when digging into the code I realised it would be simpler to copy all current speaker into the existing User table rather than doing the oposite and renaming the speaker table

Yet this change is quite massive. I have done a fair amount of tests but chances are high that I missed something

## Changes

For now I have disabled the ability for a User (Speaker) to update its profile or for anonymous users to suggest improvements. It was somehow broken and I thought that with the new User profile we might want to review that part anyway

## How to test 

Clone and migrate then run this task to transfer the speakers to the user table

```
rake speakers:migrate_to_users
```

close #894


adrienpoly:
https://github.com/rubyevents/rubyevents/issues/922

This has to be fixed

PR #923, opened 2025-08-20
add watched talks progress

### LINKED ISSUE #862: Recently Watched feature with video progress tracking
## Description:
When users start watching a video, it would enhance their experience if RubyEvents remembered their viewing progress. Upon returning to a previously watched video, users could automatically resume from where they left off.

## Proposed Implementation:

For authenticated users, periodically send requests to update their current playback position.

Utilize this progress data to build and display a "Recently Watched" list of talks, allowing users easy access to content they haven't finished watching.

## UX Benefits:

- Improves user convenience by reducing the need to manually seek through videos.
- Encourages users to revisit and complete videos, increasing engagement and retention.
- Provides an new entry point to the content where you can easily find the videos you started watching

jmarsh24:
fixes #862

### DIFF
--- a/app/views/talks/_talk.html.erb
+++ b/app/views/talks/_talk.html.erb
@@ -25,7 +25,7 @@
         video_player_end_seconds_value: talk.end_seconds,
         video_player_watched_talk_path_value: talk_watched_talk_path(talk),
         video_player_current_user_present_value: signed_in?,
-        video_player_progress_seconds_value: talk.watched_talks.select { it.user_id == Current.user.id }.min_by(&:created_at)&.progress_seconds || 0        
+        video_player_progress_seconds_value: talk.watched_talks.select { it.user_id == Current.user.id }.min_by(&:created_at)&.progress_seconds || 0
       } do %>
 
   <div class="aspect-video banner-img card-horizontal-img relative rounded-xl overflow-hidden outline outline-1 bg-gray-300 outline-gray-300" data-turbo-permanent=true>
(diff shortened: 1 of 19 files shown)


app/views/talks/_talk.html.erb
```diff
@@ -141,24 +144,6 @@
...
-            <% else %>
-              <%= ui_tooltip "Mark talk as watched" do %>
-                <%= ui_button url: talk_watched_talk_path(talk), method: :post, kind: :pill do %>
-                  <%= fa("circle", size: :xs, style: :regular) %>
-                  <span class="text-xs text-nowrap">Watched</span>
-                <% end %>
-              <% end %>
-            <% end %>
-          <% end %>
-
```
marcoroth: I think we should still allow people to manually mark a talk as watched, like if they saw it on YouTube directly or were there for the talk in-person

justinallenmarsh: That's a good use case. I added it back in! 🙂

marcoroth: Thanks!

adrienpoly:
I tested it, thanks this is a great addition 👏 . I see many improvements that are becoming possible with this first step. Let's move this brick to production and keep building upon it.

Some of the improvements I see:
- get the talk duration and in the recently watch view display a progress bar
- set the talk automatically as watched when the user reached 90% (?) of talk duration or press watched
- give the ability to remove an item from the recently watched list 
- I think we could have a more compact horizontal layout for the card to use in the recently watch view (to try ?)

justinallenmarsh:
Thank you for reviewing and merging this @adrienpoly! 🙏🏼 🚀 

I also thought to add some of these features, but I also wanted to keep the PR in scope and small. It was already doing a lot.

I will create some new issues now to add more functionality.

PR #927, opened 2025-08-22
Add filter for status

### LINKED ISSUE #507: Filter for Talks page

justinallenmarsh:
Fixes #507


adrienpoly:
thanks this is looking great

PR #930, opened 2025-08-23
Move avatar into view component

### LINKED ISSUE #439: Extract User Avatars into Component

aronwolf90:
Resolves #439

NOTE: there are some changes in the size and colors to make avatars more consistent.

For example

| Before | After | Diff |
|-----------|---------|-------|
| image |image | Color & initials |
| image | image | Initials can have more than two letters & small size change |

### DIFF
--- a/app/components/ui/avatar_component.html.erb
+++ b/app/components/ui/avatar_component.html.erb
@@ -2,7 +2,7 @@
   <div class="<%= size_class %> rounded-full <%= kind_class %> text-neutral-content <%= outline && "outline outline-2" %>">
     <% if content %>
       <%= content %>
-    <% else avatarable.custom_avatar? %>
+    <% else %>
       <%= content_tag :span, class: [text_size, avatarable.custom_avatar? && "hidden"].select(&:itself) do %>
         <%= avatarable.name.split(" ").map(&:first).join %>
       <% end %>
--- a/app/components/ui/avatar_component.html.erb
+++ b/app/components/ui/avatar_component.html.erb
@@ -1,0 +1,13 @@
+<div class="avatar placeholder">
+  <div class="<%= size_class %> rounded-full bg-primary text-neutral-content <%= outline && "outline outline-2" %>">
+    <% if content %>
+      <%= content %>
+    <% elsif avatarable.custom_avatar? %>
+      <%= image_tag avatarable.avatar_url(size: image_size), loading: "lazy" %>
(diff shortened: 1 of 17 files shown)


aronwolf90:
~NOTE: It is only in draft because I want to test it a bit more tomorrow.~

aronwolf90:
> NOTE: It is only in draft because I want to test it a bit more tomorrow.

It is now ready for review

app/components/ui/avatar_component.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="avatar placeholder">
+  <div class="<%= size_class %> rounded-full <%= kind_class %> text-neutral-content <%= outline && "outline outline-2" %>">
+    <% if content %>
+      <%= content %>
+    <% elsif avatarable.custom_avatar? %>
+      <%= image_tag avatarable.avatar_url(size: image_size), loading: "lazy" %>
+    <% else %>
+      <%= content_tag :span, class: text_size do %>
+        <%= avatarable.name.split(" ").map(&:first).join %>
+      <% end %>
```
marcoroth: Can we always render the initials so that when the image fails to load that we can fall back to the initials?

We had this in some places where we had `onerror="this.remove()"` on the `` tags which then revealed the initials if it failed to load.

marcoroth (commented):
This aleady looks great, thanks for working on this @aronwolf90! 🙏🏼 

We also have this page here: https://www.rubyevents.org/components

Would you mind adding your new component in all sizes to that page too? Thank you!

app/components/ui/avatar_component.html.erb
aronwolf90: done

aronwolf90:
> This aleady looks great, thanks for working on this @aronwolf90! 🙏🏼
> 
> We also have this page here: https://www.rubyevents.org/components
> 
> Would you mind adding your new component in all sizes/variants (image, initials, custom content) to that page too? Thank you!

Done:

image

Issue #34, opened 2023-06-24
Are you looking for a conferences to contribute to rubyvideo.dev? Start here!


marcoroth:
- [ ] wroc_love.rb 2012 - [YouTube Playlist 1](https://www.youtube.com/watch?v=wOYQDSeKthY&list=PLEEED8A6DA34E6F5B&pp=iAQB) and [YouTube Playlist 2](https://www.youtube.com/watch?v=wXQLil_SGCI&list=PL48893DE9CC60A17B&pp=iAQB)
- [ ] wroc_love.rb 2013 - [YouTube Playlist 1](https://www.youtube.com/watch?v=q0BQMbwzPJw&list=PLoGBNJiQoqRD4y6XcTGuxn2SE2zsFA03z&pp=iAQB) and  [YouTube Playlist 2](https://www.youtube.com/watch?v=arsK-CN5YDg&list=PLoGBNJiQoqRA-2pYwxRSNAxOte3x-C19W&pp=iAQB) and [YouTube Playlist 3](https://www.youtube.com/watch?v=P6_illGY_00&list=PLoGBNJiQoqRDrOb-hgC9LD5wyTvicHfbn&pp=iAQB)
- [ ] wroc_love.rb 2014 - [YouTube Playlist](https://www.youtube.com/watch?v=0OxNxlGSAdQ&list=PLoGBNJiQoqRCYNOfsPoxVWChbT5x5D5WP&pp=iAQB)
- [ ] wroc_love.rb 2018 - [YouTube Playlist](https://www.youtube.com/watch?v=FmJ6YMW92wA&list=PLoGBNJiQoqRDSMznObXnqyq5HbOqLucTz&pp=iAQB)
- [ ] wroc_love.rb 2019 - [YouTube Playlist](https://www.youtube.com/watch?v=bwUueshN6Rw&list=PLoGBNJiQoqRDJvwOYLuu7jnprRKhuc7Cp&pp=iAQB) 
- [ ] wroc_love.rb 2022 - [YouTube Playlist](https://www.youtube.com/watch?v=o_PuVhFH6U0&list=PLoGBNJiQoqRDWafhnWni-CwHkPJJuc0e5&pp=iAQB)
- [ ] Pivorak Conf. 2.0 (2019) - [YouTube Playlist](https://www.youtube.com/playlist?list=PLa4SYMEyNCu_0s6TbyKQwRGz97gYVchQM)
[...]

PR #937, opened 2025-08-28
Schwad talks

Schwad:
I added a couple organizations and talks from back of the day so my complete catalog shows on Ruby events. :) I tested it all locally to ensure the pages render right. Thanks for doing this project, it's nice to have a canonical home for all my speaking work!


marcoroth:
Hey @Schwad, thanks for the pull request and adding your talks!

We have been holding off with adding non-Ruby events to the website (see the `Non-Ruby Conferences with Ruby talks` section in #34). I just added these two talks to that list.

Maybe we can revisit that if we have a good way to show these events/talks on the platform. I think it's weird to have the whole event on the platform if there's just one talk. Especially if it's a non-Ruby event.

Schwad:
@marcoroth I don't know the fine intricacies of the data model etc. of the application but. Can a "talk" exist without an "event"? Or can we create an organization/event/etc. that is "Non-Ruby Conference" and allow the talks (for now) to point to that? I don't think it would be that controversial if there were only _one_ playlist/organization/etc. that could be repository for all the Ruby talks we find that aren't Ruby-explicit events?

Though I respect that the website is literally _rubyevents_, not _rubytalks_. Just from an archivist's prespective it would be great to find a way to point to these records without disrupting the overall quality of the rubyevents website

marcoroth:
Yeah that's the current problem. We don't really have a way to have talks on the website without them being attached to an event. I 100% agree and definitely want to have talks about Ruby, even if not at a Ruby event, on the website.

One hack we could do for now is to add a "Non-Ruby Conference" org and event, mark it as `type: meetup`, which makes it so that the `event_name` from the `videos.yml` file is shown in the card.

marcoroth:
Okay that works, let's do that!

CleanShot 2025-08-29 at 11 41 22@2x

marcoroth:
Okay, they are live on https://www.rubyevents.org/speakers/nick-schwaderer

Schwad:
@adrienpoly just told me at railsworld that you looked at this and merged it, thank you so much marco that's amazing!

PR #951, opened 2025-08-29
Add progress bar on video cards

### LINKED ISSUE #946: Add progress bar on video cards
#923 Introduced the ability for watched talks to be tracked. There are additional features that we would like to add here.

https://github.com/rubyevents/rubyevents/pull/923#issuecomment-3234997405

Add a progress bar that's visible anywhere the user sees video cards with the same design as youtube.

Image

justinallenmarsh:
fixes #946

### DIFF
--- a/test/controllers/watched_talks_controller_test.rb
+++ b/test/controllers/watched_talks_controller_test.rb
@@ -12,9 +12,9 @@ class WatchedTalksControllerTest < ActionDispatch::IntegrationTest
     get watched_talks_url
     assert_response :success
 
-    watched_talk_ids = assigns(:talks).map(&:talk_id)
+    talk_ids = assigns(:talks).map(&:id)
     user_watched_talk_ids = @user.watched_talks.pluck(:talk_id)
 
-    assert_equal user_watched_talk_ids.sort, watched_talk_ids.sort
+    assert_equal user_watched_talk_ids.sort, talk_ids.sort
   end
 end


app/models/watched_talk.rb
```diff
@@ -18,4 +19,15 @@
...
+  before_save :calculate_progress_percentage
+
+  private
+
+  def calculate_progress_percentage
+    return unless progress_seconds && talk.duration_in_seconds
+    return if talk.duration_in_seconds.zero?
+
+    self.progress_percentage = (progress_seconds.to_f / talk.duration_in_seconds * 100).round(2)
+  end
```
adrienpoly: I wonder if we should store this  VS computing it on the fly when we display the card. It doesn't seems to be a very expensive operation so I am unsure of the current benefit

justinallenmarsh: Good thing to point out. I agree with you. I removed the database column and I moved it into a instance method.

app/controllers/watched_talks_controller.rb
```diff
@@ -1,4 +1,6 @@
 class WatchedTalksController < ApplicationController
+  include WatchedTalks
+
   def index
     @talks = Current.user.watched_talks
       .includes(talk: [:speakers, {event: :organisation}, {child_talks: :speakers}])
```
adrienpoly, line 6: not really related to this PR but while we are on it I think it could be a good time to improve that 

here we have an @talk instance variable but it is a collection of WatchedTalk. I was tracking a small N+1 on that page and realised that the current code isn't really easy to follow. 

I think we should rely more on the watch_talk and try to avoid having this line 

<% watched_talk = user_watched_talks.find { |wt| wt.talk_id == talk.id } %> in the _card partial

adrienpoly (approved):
the feature is great, left a comment let me know if you prefer to address it in another PR

app/controllers/watched_talks_controller.rb
justinallenmarsh: I agree that it's a bit confusing. Do you have an idea of how to cleanly do this around the application?

Is the idea that we should render talks and only include watched_talk for the **Current.user**?

so if this is joined to the talk, then we render the progress? I think we want to prevent from including all of the watched talks for a talk and then searching for it in memory in the partial, right?

```ruby
<% watched_talk = talk.watched_talks.find { it.user_id == Current.user.id } %>
```

Do you have a proposed way of doing this?

adrienpoly: Let’s address this later in a follow up PR. This is something that was already here. 

Maybe a very tiny quick win would be to rename the current instance variable to watched_talks rather than talks

justinallenmarsh: okay I moved **map(&:talk)** to the controller so that way we are actually passing talks.

justinallenmarsh: I will create another issue for the refactoring the watched_talks logic. #958

PR #953, opened 2025-08-29
Populate duration

justinallenmarsh:


adrienpoly:
@jmarsh24 if you can rebase this branch so that the changes are easier to see thanks

justinallenmarsh:
@adrienpoly I rebased with main to make the changes more visible. I'm not sure about these changes though. The idea was to populate the duration from youtube, but it looks like we are calculating the duration instead.

Either way. I used my youtube API key to sync all the yml files.

PR #954, opened 2025-08-29
Fix XSS vulnerabilities

gregmolnar:
The URL parameter of the `link_to` helper can any valid URL protocols, and javascript is one of those. If a user controlled string is passed to this parameter, it can be used for an XSS. For instance on the following link: https://www.rubyevents.org/talks/the-state-of-security-in-rails-8?back_to=javascript:alert(%27Hi%27)&back_to_title=Greg+Molnar, if you click the back button, it will open an alert window.
The [`sanitize`](https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html#method-i-sanitize) helper strips unsafe protocols to prevent this.

/cc @marcoroth

### DIFF
--- a/app/views/talks/show.html.erb
+++ b/app/views/talks/show.html.erb
@@ -7,7 +7,7 @@
   <div class="w-full flex">
     <div class="flex-grow">
       <div class="flex items-center justify-between w-full hotwire-native:hidden">
-        <%= link_to back_path do %>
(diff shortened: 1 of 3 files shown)


marcoroth (approved):
Good one, thanks for catching this @gregmolnar! 🙌🏼

PR #955, opened 2025-08-29
Set proper event type

weppos:
I was looking at the code, and I noticed there is a way to specify the event type. It seems, so far, they are almost exclusively set (or I'd rather say "guessed") from the title.

Can you confirm it is correct to use the "kind"?

I made some adjustments to events I know, also as test.


adrienpoly:
yes when we run seed it will update the data https://github.com/rubyevents/rubyevents/blob/main/app/models/talk.rb#L518

PR #965, opened 2025-08-31
Add latest talks to home page

### LINKED ISSUE #964: Add progress bars to latest talks on home page
Progress bars were added in #946 and the home page was forgotten.

justinallenmarsh:
fixes #964


adrienpoly:
given the change made in #959 I think it doesn't work any more. This being said I think I am ok with not haveing the progress being display on the homepage.

PR #966, opened 2025-08-31
Collaborative Filtering for Video Recommendations

### LINKED ISSUE #932: Collaborative Filtering for Video Recommendations
Enhance watch history tracking (#862) by implementing collaborative filtering to recommend videos based on similar users' viewing patterns.

- Identify users with similar watch histories.
- Recommend videos watched by similar users but not by the current user.

justinallenmarsh:
CleanShot 2025-10-27 at 17 48 55@2x

CleanShot 2025-10-27 at 17 51 11@2x

Resolves #932

### DIFF
--- a/test/controllers/recommendations_controller_test.rb
+++ b/test/controllers/recommendations_controller_test.rb
@@ -12,7 +12,8 @@ class RecommendationsControllerTest < ActionDispatch::IntegrationTest
 
     get recommendations_path
     assert_response :success
-    assert_select "h2", "Recommended for you"
+    assert_select "h1", "Recommended for you"
+    assert_select "h2", "No recommendations yet"
   end
 
   test "should display no recommendations message when user has no watch history" do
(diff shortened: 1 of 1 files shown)


app/views/page/recommended.html.erb
```diff
@@ -0,0 +1,30 @@
+<main class="container">
+  <div class="flex flex-col w-full gap-4">
+    <div class="flex items-center justify-between w-full">
+      <h2 class="text-primary shrink-0">Recommended for you</h2>
+    </div>
+
+    <% if @recommended_talks.present? %>
+      <div class="grid min-w-full grid-cols-1 gap-8 sm:grid-cols-2 md:grid-cols-3 md:[&>:nth-child(4)]:hidden lg:grid-cols-4 lg:[&>:nth-child(4)]:block">
+        <%= render partial: "talks/card", collection: @recommended_talks, as: :talk, cached: true, locals: {back_to: root_path, back_to_title: "Home"} %>
```
marcoroth: We might have to pass something along, so that people can also bookmark talks they get recommended:

CleanShot 2025-10-26 at 00 06 17@2x

justinallenmarsh: thank you for the feedback! I modified it and also moved the endpoint to /recommendations instead.

justinallenmarsh:
Yes this is a great idea! 

This is a basic attempt to create a recommendation algorithm. We can add more parameters in the future. 

We can also add different filters or topics to easily filter the recommendations#index.

I would like to merge this and then iterate to improve the recommendation experience later on.

marcoroth:
@justinallenmarsh yes, this is going on main as soon as the CI passes 👍🏼

PR #968, opened 2025-08-31
Add option to remove talk from recently watched list

### LINKED ISSUE #944: Remove item from recently watched list
#923 Introduced the ability for watched talks to be tracked. There are additional features that we would like to add here.

https://github.com/rubyevents/rubyevents/pull/923#issuecomment-3234997405

Add an X button at the top right of the watch history which allows the user to remove it from their watch list.

Image

justinallenmarsh:
fixes #944


justinallenmarsh:
done!

PR #974, opened 2025-09-01
Merge duplicate speakers

weppos:
This PR merges some speakers that I found to be duplicate, generally because of a typo in their name.


adrienpoly:
thanks I ll review this after RailsWorld as I think some of the changes won't be directly applied to prod. We probably need to change them in the admin

adrienpoly:
thanks for this contribution. I have also updated the production DB to assigns the canonical users

Issue #929, opened 2025-08-22
Slides Available Icon is off


marcoroth:
Maybe this came from the migration to Font Awesome in #416, but the positioning of the "slides available" and "bookmark" icon on the talk card is slightly off:

<img width="444" height="300" alt="Image" src="https://github.com/user-attachments/assets/5e76a5d8-8d6c-4da4-bc8b-bbf146b8eaf4" />

marcoroth:
Resolved via #976

PR #976, opened 2025-09-02
fix: bookmark icon alignment to match slides icon

nageswarchedella:
Fixes issue: #929 

| Before | After |
|:-----|:---:|
| before alignment |  after alignment |


adrienpoly:
thanks

PR #981, opened 2025-09-04
Event Attendees

adrienpoly:
This PR adds the ability for user to add events to their list of attended events

It changes the way we link a user to an event for the speakers as we now have this new talk EventAttendance with different kind of attendance (keynote_speaker, speaker, visitor)

@marcoroth locally you will need to run this script

```
rake backfill:speaker_participation 
```

In a subsequent PR I will update the seed scrip 

## preview of the Event page

CleanShot 2025-09-04 at 07 25 22@2x


justinallenmarsh:
@adrienpoly do you sleep? 😴 🤣

PR #986, opened 2025-09-04
Update passport tooltip to be gender inclusive

erinosher:
Just a simple improvement to use “their” instead of “his” on the passport tooltip. As far as I can tell, the GitHub API doesn’t return user pronouns even if set.


adrienpoly:
Thanks this is an excellent point

PR #992, opened 2025-09-05
Use ReactionView

adrienpoly:
because why not

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -177,3 +177,5 @@ gem "ostruct"
 gem "hotwire_combobox", "~> 0.4.0"
 
 gem "rails-i18n", "~> 8.0"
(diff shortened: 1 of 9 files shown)


adrienpoly:
@marcoroth FYI, I forked AVO and changed [the component in AVO](https://github.com/avo-hq/avo/commit/b0efc70c21d183edc4dc35cc56cc027bcfbd4794) but now it fails in another component for no obvious reasons from what I can tell.

adrienpoly:
It is now running live in staging https://staging.rubyevents.org/
To make it work I had to [patch AVO ](https://github.com/avo-hq/avo/pull/4069) as we had compilation errors in production environements

PR #995, opened 2025-09-07
Add Organizer as a kind of Attendance

adrienpoly:
for people like Amanda 

This is what we currently have which feels wrong 😂
CleanShot 2025-09-07 at 16 40 41@2x


adrienpoly:
look better

CleanShot 2025-09-07 at 18 12 25@2x

PR #1001, opened 2025-09-12
Show Event Participants List on Event Page

### LINKED ISSUE #983: Show attende list on events show page
image

nageswarchedella:
This PR adds the ability to see a list of attendees for an event.

  Summary of Changes:

   * An "Attendees" tab now appears on event pages, leading to a new page that lists all registered attendees.
   * Attendee names link to their respective profiles.
   * Closes #983 
   
attendee list

  Technical Details:

   * Implemented Events::VisitorsController and corresponding views.
   * Added the /events/:event_slug/visitors route.
   * Included controller tests.
   * Optimized speaker and visitor queries to prevent N+1 database queries.

  How to Test:

   1. Go to an event page.
   2. Click the "Attendees" tab.
   3. Confirm the list of attendees is displayed correctly.

### DIFF
--- a/test/controllers/events/participations_controller_test.rb
+++ b/test/controllers/events/participations_controller_test.rb
@@ -8,14 +8,14 @@ class Events::ParticipationsControllerTest < ActionDispatch::IntegrationTest
   end
 
   test "should get index" do
-    get event_participations_url(@event)
+    get event_participants_url(@event)
     assert_response :success
     assert_select "div", /#{@user.name}/
   end
 
   test "should get index for signed in user" do
     sign_in_as @user
-    get event_participations_url(@event)
+    get event_participants_url(@event)
     assert_response :success
     assert_select "div", /#{@user.name}/
   end


adrienpoly:
I tested it. It works great 

I would like to suggest a minor change rather than displaying only the visitors, I think we could display all kind of participants. This avoids to have an event page without attendees. For most events we will have at least the speakers to start from and then user can gradually add their participation.

If we do this maybe the controller could be renamed from visitors to attendees or participants

WDYT?


## Example for FriendlyRB
CleanShot 2025-09-13 at 07 35 33@2x

nageswarchedella:
> I would like to suggest a minor change rather than displaying only the visitors, I think we could display all kind of participants. This avoids to have an event page without attendees. For most events we will have at least the speakers to start from and then user can gradually add their participation.

 Thanks for the feedback! That's a great suggestion.

 I agree that a unified list of all participants would be more useful and would prevent the page from being empty.

 My one question was about the existing "Speakers" tab. Following your suggestion, my plan would be to:
 
 1.  Rename the new tab to "Participants".
 2.  Update it to show all participant types (speakers, visitors, etc.).
 3.  Remove the old "Speakers" tab to avoid duplication.
 
 This would create a single, clear place to see everyone involved in the event.

 Does that sound like the right approach to you? If so, I'll go ahead and make the changes.

adrienpoly:
Sounds good 👌

marcoroth:
Yeah, I think a combined tab could make sense. But then I'd suggest that we also have some kind of sub-category, so you can look at the speakers/participants separately, similar to how we do it on the profile page:

CleanShot 2025-09-25 at 17 19 06@2x

adrienpoly:
> Yeah, I think a combined tab could make sense. But then I'd suggest that we also have some kind of sub-category, so you can look at the speakers/participants separately, similar to how we do it on the profile page:
> 

in such case maybe the tab should be visible only if we have several categories?

PR #1007, opened 2025-09-15
Add basic export to ics (calendar format)

dennispaagman:
This adds a simple icalendar export of the upcoming conferences. This way you can have an up-to-date list of upcoming conferences in your calendar. 

http://localhost:3000/events.ics

This file can be read by all major calendar apps, [Google][1], [Apple][2] etc. My experience is that Google updates it once every ~12 hours, so server load should be ok.

I haven't added a link to it in the UI yet, what do you think would be a good place for that?

[1]: https://support.google.com/calendar/answer/37100?co=GENIE.Platform%3DDesktop&hl=en&oco=1
[2]: https://support.apple.com/guide/calendar/subscribe-to-calendars-icl1022/mac

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -12,6 +12,19 @@ class EventsController < ApplicationController
     @events = Event.includes(:series, :keynote_speakers)
       .where(end_date: Date.today..)
       .order(start_date: :asc)
+
+    respond_to do |format|
+      format.html
+      format.ics do
+        calendar = Icalendar::Calendar.new
+
+        @events.where(date_precision: :day).each do |event|
+          calendar.add_event(event.to_ical)
+        end
+
+        render plain: calendar.to_ical
+      end
+    end
   end
 
   # GET /events/1
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -357,6 +357,18 @@ class Event < ApplicationRecord
     }
   end
 
+  def to_ical
+    Icalendar::Event.new.tap do |event|
+      event.uid = "RUBYEVENTS-#{id}"
+      event.dtstart = start_date
+      event.dtend = end_date + 1.day # dtend is exclusive, add 1 day to make it inclusive
+      event.summary = name
+      event.description = description
+      event.location = static_metadata.location
+      event.url = website
+    end
+  end
+
   private
 
   def todos_data_path
--- a/app/views/events/years/_header.html.erb
+++ b/app/views/events/years/_header.html.erb
@@ -22,6 +22,37 @@
             across <%= events_by_month.keys.count %> months
           <% end %>
         </span>
+
+        <%= ui_button "Subscribe to calendar", size: :sm, kind: :link, data: {toggle: "modal", target: "modal-middle"} %>
+        <%= ui_modal position: :middle, id: "modal-middle" do %>
+          <h1>Subscribe to calendar</h1>
+          <div class="flex flex-col gap-3">
+            <p>Copy the URL below and add it to your calendar application as an external calendar.</p>
+            <div
+              class="flex justify-between items-center"
+              data-controller="copy-to-clipboard"
+              data-copy-to-clipboard-success-class="scale-95">
+              <pre
+                class="text-sm font-mono text-gray-800 whitespace-pre-wrap break-words"
+                data-copy-to-clipboard-target="source">
+                <%= events_url(format: :ics) %>
+              </pre>
+              <button
+                class="btn btn-sm btn-outline transition-transform duration-150 ease-in-out"
+                data-action="click->copy-to-clipboard#copy"
+                data-copy-to-clipboard-target="button">
+                Copy
+              </button>
+            </div>
+            <p>Instructions for your device can be found here:</p>
+            <ul>
+              <li><%= external_link_to "Google Calendar / Android", "https://support.google.com/calendar/answer/37100?co=GENIE.Platform%3DDesktop&amp;hl=en&amp;oco=1" %></li>
+              <li><%= external_link_to "Microsoft Outlook", "https://support.microsoft.com/en-us/office/import-calendars-into-outlook-8e8364e1-400e-4c0f-a573-fe76b5a2d379" %></li>
+              <li><%= external_link_to "Mac", "https://support.apple.com/kb/ph11523?locale=en_US" %></li>
+              <li><%= external_link_to "iPhone (iCloud)", "https://support.apple.com/en-us/HT202361" %></li>
+            </ul>
+          </div>
+        <% end %>
       </p>
     </div>
 
--- a/Gemfile
+++ b/Gemfile
@@ -282,3 +282,5 @@ group :test do
   gem "vcr", "~> 6.1"
   gem "webmock"
 end
+
+gem "icalendar", "~> 2.12"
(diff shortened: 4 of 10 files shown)


saiqulhaq:
I **agree with the addition of .ics export**—this will be very useful for users who want to sync upcoming events with their calendar apps.

For performance, I recommend using `find_each` instead of `each` when iterating over the events. `find_each` loads records in batches and is much more efficient for potentially large datasets.

Additionally, given how calendar clients like Google pull updates (~12 hours), I suggest adding a cache for the generated .ics file to reduce database load. The cache could be refreshed daily or after relevant updates, which aligns well with client sync intervals.

dennispaagman:
Thanks @saiqulhaq for your input!

> For performance, I recommend using `find_each` instead of `each` when iterating over the events. `find_each` loads records in batches and is much more efficient for potentially large datasets.

The default batch size for `find_each` is 1000 though, I don't think we'll ever have a 1000+ upcoming Ruby events 😅 but I agree in general that's it's a good best practice to use it.

> Additionally, given how calendar clients like Google pull updates (~12 hours), I suggest adding a cache for the generated .ics file to reduce database load. The cache could be refreshed daily or after relevant updates, which aligns well with client sync intervals.

I think the load will be very minimal, it's usually only ~20 records and generating the ics is also very fast as it's just a bit of string manipulation. The extra round trip to grab the cached record from the database is probably about as fast, and just not caching is less complex.

saiqulhaq:
I agree with your assessment—since there aren’t that many upcoming Ruby events, the current implementation makes sense and should work well for now.

However, for added robustness, I suggest adding an integration test that uses `Icalendar::Calendar.parse` to validate the generated .ics data directly (rather than just checking for the presence of "BEGIN:VCALENDAR"). This way, we can be sure the ICS output remains valid and compatible with common calendar clients, even as we add more data or hit edge cases in the future.

What do you think?

benjaminoakes:
Thank you for the work so far on this.

I miss the ICS feed feature from [Ruby Conferences](https://rubyconferences.org/).  Is it still planned?

adrienpoly:
Thanks for reviving this PR I'll give it a look asap

dennispaagman:
I rebased it to fix the conflicts and updated icalendar as it had a [CVE](https://github.com/advisories/GHSA-pv9c-9mfh-hvxq). Happy to help out more if you think anything is missing feature-wise :)

app/controllers/events_controller.rb
```diff
@@ -12,6 +12,21 @@ def index
...
+
+        @events.where(date_precision: :day).each do |event|
+          cal.add_event(event.to_ical)
+        end
+
+        cal.publish
+
+        render plain: cal.to_ical
+      end
+    end
```
ChaelCodes, line 27: Can you tell me a bit more about the lifecycle of this? I read through the gem readme and looked at the publish method but it was missing some detail. 

I'm mainly concerned with how events are added, whether Google or Apple call this endpoint, and what publish does.

app/models/event.rb
```diff
@@ -357,6 +357,18 @@ def to_mobile_json(request)
     }
   end
 
+  def to_ical
```
ChaelCodes, line 360: This method is good, I just want a threaded location to discuss link placement.

Maybe in the subtitle of this page? https://www.rubyevents.org/events/2026
"Add to calendar" or "Subscribe to calendar"

ChaelCodes (commented):
Thank you for building this out! I know it'll be a popular feature once it's released. Maybe we could add a calendar for CFPs too in a separate PR once this one merges?

I have some open questions about how the gem works, and I'd like to test this over lunch tomorrow.

As for the link, I'm having trouble placing it too. The calendar view feels most appropriate.

app/models/event.rb
```diff
@@ -357,6 +357,18 @@ def to_mobile_json(request)
...
+    Icalendar::Event.new.tap do |event|
+      event.id = id
```
ChaelCodes: https://github.com/icalendar/icalendar/blob/main/lib/icalendar/event.rb

Image

I'm getting an error saying `id` does not exist. Is it looking for uid?

dennispaagman: Hmm not sure how I missed that, the test seems not to render any events so that didn't catch it. I'll fix it and improve the tests.

app/controllers/events_controller.rb
dennispaagman: I basically cargo-culted in the `publish` call from another codebase that does something similar, I looked into it and it seems unnecessary, so I'm going to remove it.

Basically the lifecycle is:

1. create a `Icalendar::Calendar` object
2. Add events to it(as `Icalendar::Event` objects or directly through `.event { .. }` or the object returned from `.event`, which is a bit weird)
3. Transform it into ical string format (`to_ical`) and return that to the browser

app/models/event.rb
dennispaagman: I implemented a modal with some instructions, as it can be a bit hard to figure out what to do otherwise if you've never encountered such a file.

Feedback on the copy is more than welcome as I winged it a bit.

Image

Image

app/controllers/events_controller.rb
ChaelCodes: Thanks!

app/models/event.rb
ChaelCodes: That looks good!

app/views/events/years/_header.html.erb
```diff
@@ -22,6 +22,37 @@
...
+                Copy
+              </button>
+            </div>
+            <p>Instructions for your device can be found here:</p>
+            <ul>
+              <li><%= external_link_to "Google Calendar / Android", "https://support.google.com/calendar/answer/37100?co=GENIE.Platform%3DDesktop&amp;hl=en&amp;oco=1" %></li>
+              <li><%= external_link_to "Microsoft Outlook", "https://support.microsoft.com/en-us/office/import-calendars-into-outlook-8e8364e1-400e-4c0f-a573-fe76b5a2d379" %></li>
+              <li><%= external_link_to "Mac", "https://support.apple.com/kb/ph11523?locale=en_US" %></li>
+              <li><%= external_link_to "iPhone (iCloud)", "https://support.apple.com/en-us/HT202361" %></li>
+            </ul>
```
ChaelCodes, line 53: Love this. 😍

ChaelCodes:
I haven't thoroughly investigated, but I've rerun your CI job, and I don't recognize the failing spec as a flaky one. It might be legitimate?

dennispaagman:
Ah yeah it looks legit, probably because I changed the fixture. I'll see if there's a different way to get that future_conference fixture working properly with my new tests.

dennispaagman:
The issue I had was solved in main already, I rebased and this should be good now!

ChaelCodes:
We are so close to merging this!!

But right now, two day events look like one day events, because the time is midnight. :/
Screenshot 2026-05-05 at 1 05 40 PM
Screenshot 2026-05-05 at 1 09 06 PM

dennispaagman:
ah yes that's a good catch! fixed it now @ChaelCodes :)

ChaelCodes (approved):
YAY!! It works!!!

Image

dennispaagman:
Awesome, I'm getting it into my Google Calendar to verify, but that seems to still sync on their backend. I'll keep a tab open and check back tomorrow!

It might be wise to keep an eye on the number of requests on that new endpoint, I think it's probably fine but not sure how much polling will be going on.

Thanks @ChaelCodes for pushing this over the line!

PR #1009, opened 2025-09-15
Change Verified Speaker tooltip

### LINKED ISSUE #1003: "Verified Speaker" for non-speakers
There is a "verified speaker" badge for non-speakers.

I assume this is historical, perhaps when there wasn't the concept of regular users on the site(?). But now, `verified` probably shouldn't refer to being a verified "speaker" so much as "user has claimed their profile on the site".

Possible options:

1. Remove the badge for non-speakers
2. Keep the badge and change the language to not mention speaker
3. Have separate badges for speakers and less cool people

Image

shkm:
Not a great deal to say here 😅. Seems like the new language fits to me. Thoughts?

Closes #1003

### DIFF
--- a/app/views/profiles/_header_content.html.erb
+++ b/app/views/profiles/_header_content.html.erb
@@ -10,7 +10,7 @@
           <span class="text-sm content-center text-[#737373]">(<%= user.pronouns %>)</span>
         <% end %>
         <% if user.verified? %>
-          <%= ui_tooltip "#{user.name} is a verified user" do %>
+          <%= ui_tooltip "#{user.name} is a verified Rubyist" do %>
             <%= fa("badge-check", class: "fill-blue-500", size: :sm) %>
           <% end %>
         <% end %>


app/views/profiles/_header_content.html.erb
```diff
@@ -10,7 +10,7 @@
           <span class="text-sm content-center text-[#737373]">(<%= user.pronouns %>)</span>
         <% end %>
         <% if user.verified? %>
-          <%= ui_tooltip "#{user.name} is a verified speaker" do %>
+          <%= ui_tooltip "#{user.name} is a verified user" do %>
```
marcoroth: What if we say `Rubyist`? 🙈 

```suggestion
          <%= ui_tooltip "#{user.name} is a verified Rubyist" do %>
```

shkm: Ooh that makes sense!

PR #1011, opened 2025-09-16
Make icons of favorite videos visible

enderahmetyurt:
Favorites videos in the list should be self-explanatory and distinguishable from others.

# Before

https://github.com/user-attachments/assets/1e686764-e503-4237-8392-25c75faeaf4a

# After
Screenshot 2025-09-16 at 13 05 47

Screenshot 2025-09-16 at 13 05 40


justinallenmarsh:
@adrienpoly  @marcoroth Should we consider renaming this route from /watch_lists to favorites? 

From a domain perspective I find watch_list as a confusing term now that we have added "recently watched". I think this also has further implications of renaming watch_list in the model directories also. 

I'm guessing, but it seems that watch_list was intended to be a watch history for a user that they had to manually mark, but now this is automatic and we have a separate watched_talk model to store this information.

PR #1006, opened 2025-09-15
Set up devcontainers

ChaelCodes:
# Description
closes #1004

This pull request adds support for devcontainers to streamline development environment setup. It is purely additive, and does not impact the existing dev setup.

- git auth is handled using gh auth login, and it supports SSH
- docker-compose.yml is duplicated inside .devcontainers, so the existing dev setup will still work
- the db:prepare is run on creation
- the server automatically starts on start-up, and the port is forwarded to localhost:3000

## Devcontainer support and configuration

* Added `.devcontainer/Dockerfile` to define the development container, including installation of Ruby, Node.js, Yarn, and all required system dependencies.
* Added `.devcontainer/docker-compose.yml` to orchestrate the Rails app and Meilisearch services, including port mappings, volumes, and environment variables.
* Added `.devcontainer/devcontainer.json` to configure the devcontainer for VS Code, specifying services, port forwarding, post-creation commands, and feature integrations like SSH and GitHub CLI.

## Documentation updates

* Updated `README.md` to document devcontainer support, including setup instructions, workflow notes, and usage tips for working within the containerized environment.

PR #1014, opened 2025-09-16
DevContainers support interactive rebase and commit messages

ChaelCodes:
# Description

Introduced in #1006 

Right now, if you try to do an interactive rebase or write a commit message in the editor, it can't find a text editor, and raises an error.

This change opens commit messages and interactive rebases in VS Code.


ChaelCodes:
Sorry @adrienpoly , I was investigating this right as you merged. 😅

PR #1017, opened 2025-09-17
Make empty state page consistent

enderahmetyurt:
Recently watched and favorites empty state should be same.

# Before
Screenshot 2025-09-17 at 19 30 01
Screenshot 2025-09-17 at 19 30 16

# After
Screenshot 2025-09-17 at 19 43 31


adrienpoly:
thanks

PR #999, opened 2025-09-10
Style: Apply automated formatting and style fixes from CI

nageswarchedella:
This pull request addresses various code style and formatting issues identified by running the `bin/ci`
  script. The changes ensure the codebase adheres to the project's coding conventions and improves overall
  readability and maintainability.

  Changes Included:

   * Code Formatting: Applied automated formatting rules across the codebase, primarily adding newlines for
     better readability and consistency. This change is the result of running bin/ci.
   * Data File Updates: Updated numerous YAML data files to use multiline descriptions (description: |)
     instead of empty strings. This allows for more descriptive content for events and talks in the future.
   * Code Refinements:
       * Improved Readability: In app/views/talks/_talk.html.erb, a block variable was renamed from it to
         watched_talk for better clarity.
       * Simplified Syntax: Removed redundant begin...end blocks from the speakers.rake task, making the code
         more concise.

  How to Verify:

  To verify the changes, run the CI script and ensure it passes without errors: `bin/ci`

PR #1023, opened 2025-09-19
Add talks and sponsor details for RubyConf India 2025

Selectus2:


Selectus2:
There are linting issues in 89 other yml files. Will raise a different PR for the same to avoid confusion

adrienpoly:
have you upgrade some dependencies? 
I know that upgrading some js dependencies would format slightly differently the yml when we run the linter. 

This is what we saw in #999

marcoroth:
> There are linting issues in 89 other yml files. Will raise a different PR for the same to avoid confusion

Just make sure to run `yarn format:yml` and these should go away! 🙌🏼

marcoroth (approved):
Thank you @Selectus2! 🙏🏼

PR #1025, opened 2025-09-20
Event country code

adrienpoly:
We have a country_code attribute on the event model but most of the time it is empty and we get the country from the static metadata. This PR moves the static content to the model. This will be useful for stamps delivery

### DIFF
--- a/app/controllers/events/cities_controller.rb
+++ b/app/controllers/events/cities_controller.rb
@@ -12,13 +12,13 @@ class Events::CitiesController < ApplicationController
 
     @cities_by_country = @events_by_city.keys.group_by { |city|
       events = @events_by_city[city]
-      country = events.first..country
+      country = events.first.country
       country&.translations&.dig("en") || "Unknown"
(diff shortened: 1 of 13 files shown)


app/controllers/events/cities_controller.rb
```diff
@@ -12,13 +12,13 @@ def index
 
     @cities_by_country = @events_by_city.keys.group_by { |city|
       events = @events_by_city[city]
-      country = events.first.static_metadata&.country
+      country = events.first..country
```
chuggingCoffee: Hi @adrienpoly - I'm not sure if you're still working on this, but I accidentally clicked this PR instead of mine and this caught my eye.  There are a few instances of `first..country` in these changes.  I want to mention it just in case this accidentally slipped through the cracks.

Cheers,
Mike

adrienpoly: Thanks it was late at night. Good idea that I let it sit for the night. Will correct that. Thanks for the review

Issue #212, opened 2024-09-23
Rethink speaker leaderboard


marcoroth:
#208 introduced a speaker leaderboard. While these are cool metrics I think it shouldn't be a seen as a "competition" or "leaderboard" on who has given the most talks.

Maybe we can pack this information differently. For example, by providing a "sort" column on the /speakers page and having the options  "Name ASC/DESC, Talks Count ASC/DESC" etc. to convey the same information but with a different intent.

PR #1040, opened 2025-09-30
Add basic location to user profile

hschne:
This PR adds a basic location setting to the user profile. See also discussion on [Ruby Europe Discord](https://discord.com/channels/1161951534070890518/1240667547217563688/1422534222589202463). I updated the styling of the website link too, IMO makes sense to change the icon & alignment in this particular case, as it's not a 'generic' external link but rather profile specific. 

I also removed the `speaker_params` from `speakers_controller`, didn't seem like this is used at all :thinking: 

image
image

For now, this is just a basic country filter that can be set when editing the profile. 

image

 
Possible follow-ups: 
- Improving the location selection, could be City + Country as well
- Update the speakers page (could use a face-lift anyway?) to include a filter by location. Related issues/PRs may be #674  [#212](https://github.com/rubyevents/rubyevents/issues/212)

### DIFF
--- a/db/migrate/20250930165602_add_location_to_user.rb
+++ b/db/migrate/20250930165602_add_location_to_user.rb
@@ -1,5 +1,5 @@
 class AddLocationToUser < ActiveRecord::Migration[8.1]
   def change
-    add_column :users, :location, :string
+    add_column :users, :location, :string, default: ""
   end
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -361,7 +361,7 @@ ActiveRecord::Schema[8.1].define(version: 2025_09_30_165602) do
     t.string "github_handle"
     t.json "github_metadata", default: {}, null: false
     t.string "linkedin", default: "", null: false
-    t.string "location"
+    t.string "location", default: ""
     t.string "mastodon", default: "", null: false
     t.string "name"
     t.string "password_digest"


adrienpoly:
Love this idea!

A lot of the user information can be gathered from the GitHub profile. The GitHub profile has a location attribute, which is essentially a free-text field where users can enter almost anything. I was wondering whether we’d want to replicate that approach.

Right now, your PR focuses more on a structured input for country. On GitHub, though, the location field is often used for more detailed information (e.g. a city or region within a country).

One possible approach could be:
- Keep a location field and, if available, populate it with the GitHub value.
- Add a country_code attribute that we try to extrapolate from the location whenever possible.

food for thoughts

hschne:
@adrienpoly It was my impression that modifying `app/models/user/profiles.rb` would take care of the GitHub enrichment, right? If not, happy to make any adjustments to make the GH import work.

> Right now, your PR focuses more on a structured input for country

Kinda. Manually editing in the profile is restrictive, as it allows countries only, but there are no validations. So the GH import should just work, and the location field would be populated with whatever  the user has in their GH profile. 

> Add a country_code attribute that we try to extrapolate from the location whenever possible.

GH allows the user to input literally anything, and if we do the same thing (just allow plain-text input) extrapolating would be challening. I'm leaning free-text input, but with city + country autocomplete (as gentle nudge). That would mean GH import works, we should get mostly structured data we can work with, but users can still put in 'on the moon' if they feel like it). Then, extrapolating as you suggested and using a `country_code` attribute.

Definitely follow-up material, though. Anything I should still take care of in this PR to get this merged?   :thinking:

Issue #169, opened 2024-09-02
Re-function organizations and introduce "conference series"


marcoroth:
Currently we have "organizations" that we use to "group" events together. RailsConf 2011-2024 all belong to the RailsConf "Organization" right now. 

But technically, RailsConf and RubyConf are organized by Ruby Central. Similarly Rails World if organized by the Rails Foundation.

I think it would make sense to either:
* rename "organizations" to something like "conference" or "conference series", since I think what we really want is to group a long-running conference by it's editions.
* or introduce "conference"/"conference series", keep "organizations" and re-organize the data so we have a structure like:

```
RubyVideo
├─ Ruby Central
│  ├─ RailsConf
│  │ ├─ RailsConf 2011
│  │ ├─ ...
│  │ ├─ RailsConf 2024
│  │  
│  ├─ RubyConf
│    ├─ RubyConf 2011
│    ├─ ...
│    ├─ RubyConf 2023
│
├─ Rails Foundation
  ├─ Rails World
    ├─ Rails World 2023
    ├─ Rails World 2024
    ├─ ...
```

marcoroth:
After thinking so more about this, if we want to support videos from meetups, like the "Helsinki Ruby Brigade" we have in our data right now, we shouldn't have to create an "event/conference" that just holds these videos. 

It makes sense that there is an organization in this case, but we probably shouldn't add all the videos from muliple meetups into the same "conference".

Maybe something like this could work:

```
RubyVideo
└─ Helsinki Ruby Brigade            // <- Organization
   └─ Helsinki Ruby Brigade Meetup  // <- Meetup Series
      ├─ Video 2024-01-01           // <- Talk/Video
      ├─ ...
      └─ Video 2024-12-12
```

or maybe even simpler:

```
RubyVideo
└─ Helsinki Ruby Brigade    // <- Organization
    ├─ Video 2024-01-01     // <- Talk/Video
    ├─ ...
    └─ Video 2024-12-12
```

PR #1048, opened 2025-10-10
Event Involvements

marcoroth:
CleanShot 2025-10-10 at 14 02 44@2x

CleanShot 2025-10-10 at 13 59 36@2x

CleanShot 2025-10-10 at 13 59 29@2x


adrienpoly:
Great 

The EventInvolvement table is polymorphic so that it can be attached to User or Organisation 

I wonder what are the links between an event and an Org that are not yet currently covered by our Org -> Events and Sponsor -> Events associations

marcoroth:
> I wonder what are the links between an event and an Org that are not yet currently covered by our Org -> Events and Sponsor -> Events associations

Good point! The plan for the current `sponsors` table it to become `organizations` (similar to what we did with `speakers` -> `users`) and then the join table defines which `Organization` is sponsoring which `Event`.

Additionally, the current `organizations`  should be renamed to "Conference Series" (see #169) and then we should be good!

adrienpoly:
I get it thanks for the clarifications 👍

PR #1049, opened 2025-10-10
Add speaker search on speaker search page

hschne:
This is a revamp of #674 and prep for filtering speakers by location (#1040). 

I did some trials with typeahead search (as suggested in the linked issue) and found the UX not so nice. The current implementation refreshes the list of speakers whenever we type. Had to use a turbo frame rather than stream because the lazy loading pagination caused weirdness otherwise. This was the simplest approach I could find to make it work. 

I also removed the letter filtering. With the search in place it seems pointless :shrug:

### DIFF
--- a/app/views/speakers/_speaker.html.erb
+++ b/app/views/speakers/_speaker.html.erb
@@ -1,19 +1,13 @@
 <%= link_to profile_path(speaker), id: dom_id(speaker), class: "flex justify-between items-center" do %>
-  <div class="flex items-center gap-3 gap-y-4">
-    <div class="flex w-fit">
-      <%= ui_avatar(speaker, size: :sm) %>
-    </div>
-
+  <div class="flex items-center gap-3">
     <span><%= sanitize speaker.name_with_snippet %></span>
   </div>
 
-  <div class="flex items-center gap-2">
-    <%= ui_badge(
-          speaker.talks_count,
-          kind: :secondary,
-          outline: true,
-          size: :lg,
-          class: "min-w-10"
-        ) %>
-  </div>
+  <%= ui_badge(
+        speaker.talks_count,
+        kind: :secondary,
+        outline: true,
+        size: :lg,
+        class: "min-w-10"
+      ) %>
 <% end %>


adrienpoly:
I like it 👍 

I noticed that the infinite scroll it currently broken, the upcoming collection of speakers in not appended to the correct div with the grid. I think you might want to add an id to this div and target it for the update 


I think the PP are a bit too big and they are not perfectly aligned with the text right now
Here is a test I did with `:sm` size for the avatar and an increased y gap to 1rem (`gap-y-4`) for the grid

CleanShot 2025-10-11 at 11 12 14@2x

To fix the alignement I changed that in _speaker partial

```diff
-<div class="relative w-fit">
+<div class="flex w-fit">
   <%= ui_avatar(speaker, size: :sm) %>
</div>
```


About the picture profile we tried that in the past with Marco. We found out that GitHub has rate limits for those images. So if you scroll down a bit very quickly you end up with only read PP. We now fallback to initials so it is not as bad as it was before with broken images.

I had an initial PR to serve our own PP images to avoid this limit https://github.com/rubyevents/rubyevents/pull/881. This is something that I would like to rework in the future with some kind of active storage.

hschne:
@adrienpoly Changes implemented, pagination works now with searches and without. One minor issue: When searching through the omnibar the URL is updated - searches on the speakers search don't do that, and that can be confusing. 

I don't know what the best solution would be, and I think it's a minor issue, hence I didn't touch it. The solution would be to do some history updates via Stimulus I guess... feels weird to do that on auto-submit though. What do you think?

app/views/speakers/_speaker.html.erb
```diff
@@ -1,7 +1,19 @@
 <%= link_to profile_path(speaker), id: dom_id(speaker), class: "flex justify-between items-center" do %>
-  <div class="flex items-center gap-3">
+  <div class="flex items-center gap-3 gap-y-4">
+    <div class="flex w-fit">
+      <%= ui_avatar(speaker, size: :sm) %>
+    </div>
```
marcoroth: Sadly we can't really do this with our current avatar approach, as GitHub will rate limit you and will stop serving you avatars.

hschne: See Adrian's comment, I thought the fallback of rendering initials is acceptable, but I guess I misunderstood. 

WDYT about proxying Avatar image requests through the server and caching them (with, say TTL of 1 day or something)? Seems simpler than involving ActiveStorage :thinking:

hschne: Removed it, it's not strictly necessary for this PR and I'll address the avatar issue at some later point :+1:

adrienpoly: > WDYT about proxying Avatar image requests through the server and caching them (with, say TTL of 1 day or something)? Seems simpler than involving ActiveStorage 🤔

How would you implement this?

marcoroth:
I'm curious if it still makes sense to have a search bar on the speakers page. Now that we have https://github.com/rubyevents/rubyevents/pull/1236 and https://github.com/rubyevents/rubyevents/pull/1226.

hschne:
Agreed 👍

PR #1055, opened 2025-10-11
Add Chromium Support to Devcontainer for System Tests and Improve Setup Commands

### LINKED ISSUE #1054: Automation test doesn't work on Devcontainer
```
> root@e4b97e474e2c:/rails# bin/dev
Installing foreman...
Fetching foreman-0.90.0.gem
Successfully installed foreman-0.90.0
1 gem installed

A new release of RubyGems is available: 3.6.9 → 3.7.2!
Run `gem update --system 3.7.2` to update your installation.

14:19:03 web.1  | started with pid 4429
14:19:03 vite.1 | started with pid 4430
14:19:03 jobs.1 | started with pid 4431
14:19:04 vite.1 | yarn run v1.22.22
14:19:04 vite.1 | error Command "vite" not found.
14:19:04 vite.1 | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
14:19:04 vite.1 | exited with code 1
14:19:04 system | sending SIGTERM to all processes
14:19:04 web.1  | terminated by SIGTERM
14:19:04 jobs.1 | terminated by SIGTERM
```

```
> root@e4b97e474e2c:/rails# bin/rails test --fail-fast
Running 266 tests in parallel using 10 processes
Run options: --fail-fast --seed 2477

# Running:

...........E

Error:
CountriesControllerTest#test_should_sort_events_by_home_sort_date_in_reverse_order_on_show_page:
ActionView::Template::Error: Vite Ruby can't find entrypoints/application.js in the manifests.

Possible causes:
  - The last build failed. Try running `bin/vite build --clear --mode=test` manually and check for errors.

Errors:
  error Command "vite" not found.

Visit the Troubleshooting guide for more information:
  https://vite-ruby.netlify.app/guide/troubleshooting.html#troubleshooting
    app/views/layouts/application.html.erb:21
    test/controllers/countries_controller_test.rb:62:in 'block in <class:CountriesControllerTest>'
```

I am fixing this issue now

saiqulhaq:
Fixes #1054

- **Add Chromium browser and driver** to the devcontainer's Dockerfile, enabling system tests that require a headless browser environment.
- **Update devcontainer postCreateCommand** to run `yarn install` before preparing the Rails database, ensuring JavaScript dependencies are installed automatically.
- **Improve documentation** in the README with clear instructions for running system tests in headless mode using `HEADLESS=true bin/rails test`.

These changes ensure a smoother onboarding experience and make it easier to run system tests in the containerized development environment.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -17,6 +17,7 @@ RUN apt-get update -qq && \
     pkg-config \
     python-is-python3 \
     sqlite3 \
+    chromium chromium-driver \
     && rm -rf /var/lib/apt/lists /var/cache/apt/archives
 
 # Copy application code
@@ -37,7 +38,7 @@ RUN curl -sL https://github.com/nodenv/node-build/archive/master.tar.gz | tar xz
 # Install application gems
 RUN bundle install
 
-# Install node modules
+# Install JavaScript dependencies
 RUN yarn install
 
 # Default command
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,9 +15,7 @@
     "ghcr.io/devcontainers/features/github-cli:1": {},
     "ghcr.io/devcontainers/features/sshd:1": {}
   },
-  "forwardPorts": [
-    3000
-  ],
+  "forwardPorts": [3000],
   "portsAttributes": {
     "3000": {
       "label": "Rails Server",
@@ -25,7 +23,6 @@
     }
   },
   "postCreateCommand": "bin/rails db:prepare",
-  "postStartCommand": "bin/dev",
   "mounts": [
     "source=${localEnv:HOME}/.ssh,target=/home/vscode/.ssh,type=bind,consistency=cached"
   ]
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -10,6 +10,7 @@ services:
     ports:
       - "3000:3000"
     volumes:
-      - ..:/rails:cached
+      - ..:/rails:cached # Mounts the local project directory
+      - /rails/node_modules # Prevents overwriting node_modules
     stdin_open: true
     tty: true
--- a/README.md
+++ b/README.md
@@ -31,7 +31,8 @@ All file changes will be present locally when you close the container.
 
 - Use `gh auth login` to auth with GitHub, so you can push commits from within the container.
 - Do not run `bin/setup`. It starts docker containers, and you're already in one.
-- The application is forwarded to [localhost:3000](localhost:3000) and starts automatically on boot.
+- After the container is set up, run `bin/dev` in the terminal to start the development server. The application will be forwarded to [localhost:3000](localhost:3000).
+- To run system tests, use `HEADLESS=true bin/rails test`. The HEADLESS=true environment variable ensures Chrome runs in headless mode, which is required in the container environment. 
 
 ### Requirements
 


adrienpoly:
Hello 
Thanks for this contribution 
It looks legitimate to me but I don't often use dev containers. 

As @ChaelCodes implemented this feature maybe you have some feedback on those proposed changes?

Thanks

ChaelCodes:
Hey @saiqulhaq ! Thanks for this improvement to the devcontainer. The selenium addition looks perfectly fine to me, and I appreciate you for saving documentation.

As for the yarn install, theoretically, this line should be installing it in the docker image: 

https://github.com/rubyevents/rubyevents/blob/main/.devcontainer%2FDockerfile#L40-L41

However, I recently set up a devcontainer on another machine, and I found that I also needed to run yarn install after the docker container booted.

I'd like to try and figure out why that isn't working instead of running `yarn install` on the devcontainer.

saiqulhaq:
The root cause is .dockerignore file has `node_modules`
So I proposed to remove `yarn install` from the Dockerfile to speed up the build process

I have tried creating `Dockerfile.gitignore` in the `.devcontainer` folder to keep `node_modules`, but it doesn't work.

Docker still uses `/.dockerignore` instead of `.devcontainer/Dockerfile.gitignore` because the `.devcontainer/docker-compose` file uses `..` as context.

ChaelCodes:
Thank you for your research into this! Starting with a foundation of missing `npde_modules` made figuring out what was wrong easier.

I've been doing some testing with this. It's not that the [.dockerignore](https://docs.docker.com/build/concepts/context/#dockerignore-files) is excluding `node_modules`. That prevents `node_modules` from being copied into the build context (which we want), but doesn't affect anything that's built on the container.

However, when we load the volume, it overrides the generated `node_modules` with the empty folder from the local environment.

If we replace the volumes in .devcontainer/docker-compose.yml with the following:
```
    volumes:
      - ..:/rails:cached # Mounts the local project directory
      - /rails/node_modules # Prevents overwriting node_modules
```
then the node_modules will be stored in a separate volume and won't be overwritten. I have tested this generates `node_modules` in the container.

We want to move as many build steps as possible into the docker image. The only reason `bin/rails db:prepare` is in the devcontainer is because it takes ages to run, and I want to give contributors access to the dev environment while it runs in the background.

saiqulhaq:
Thanks @ChaelCodes — you’re absolutely right about the node_modules issue. I just tested it myself and pushed the result to this PR.

One extra thing I found: having this line in devcontainer.json

```
  "postStartCommand": "bin/dev",
```
Causes the build process to never fully stop (though VS Code works fine). It only ends after pressing Ctrl+C, which then leads to this message:

```
postStartCommand from devcontainer.json interrupted.
```

To avoid confusion, I propose removing `"postStartCommand": "bin/dev"` from devcontainer.json and instead documenting `bin/dev` usage in the README.md for contributors to run manually if needed.

Let me know what you think!

ChaelCodes (approved):
Thanks @saiqulhaq ! This looks good.

I tested it by rebuilding the container on my macbook, and running system tests, and all looks good.

ChaelCodes:
> To avoid confusion, I propose removing "postStartCommand": "bin/dev" from devcontainer.json and instead documenting bin/dev usage in the README.md for contributors to run manually if needed.

That would be fine. I'm trying to get it so that the server starts up with the devcontainer, but I see how the experience is confusing.

ChaelCodes:
As a follow-up, Rails's default devcontainer does not start up the server after creation.

https://github.com/rails/rails/blob/1ca947883d1975487e14208e8159726654fc20d1/railties/lib/rails/generators/rails/devcontainer/templates/devcontainer/devcontainer.json.tt#L35-L36

saiqulhaq:
I've just removed postStartCommand from devcontainer.json and updated the README to document manual usage instead.

saiqulhaq:
I've just removed postStartCommand from devcontainer.json and updated the README to document manual usage instead.

adrienpoly:
thanks both for this work 🙏

PR #1062, opened 2025-10-12
STAMPs -> STAMPS

shkm:
The pluralized suffix was not previously upcased.

image


marcoroth:
Thanks @shkm!

Now after thinking about this, I'm wondering if we should just titlecase it anyway 😅

shkm:
> Thanks @shkm!
> 
> Now after thinking about this, I'm wondering if we should just titlecase it anyway 😅

Yeah, the upcasing is maybe... out of place?

PR #1066, opened 2025-10-16
Update Ruby Retreat AU 2025 assets

marcoroth:
This mostly updates the banner for the Ruby Retreat AU 2025 to be more continuous.

CleanShot 2025-10-16 at 01 57 26@2x


ChaelCodes:
Screenshot 2025-10-18 at 8 06 50 AM

Looks like card isn't loading. I'm noticing this on multiple pages and I first noticed it shortly after merge, but I'm unclear on exactly when between the two merges.

My first instinct says "Oh no, what did I break?" but the second says cache issue? When does the cache rotate? Can we force it?

Issue #1073, opened 2025-10-18
Overly presumptive GB stamps


pat:
I'm wondering if it'd be best to avoid specific badges for countries within Great Britain, given the current approach is to add all of them (England, Northern Ireland, Scotland, Wales), despite events being in a single country. I don't feel I've earned the Northern Ireland and Wales badges given I've never been to a Ruby event in either country!

Essentially, I'd be thinking to remove the case statement within `Stamp.create_stamp_from_code`, and just use the `else` block of logic. Does that sound reasonable? Or have people been pondering how we could make events more granular?

marcoroth:
Agreed. I think that this is mostly an oversight from how the `countries` gem categorizes them. Ideally we could only award the stamps of the country you have actually given a talk in/attended.

pat:
If it's kept to just speakers, that's a bit limited in audience - most folks aren't lucky enough to get selected regularly.

But I'll submit a PR to remove the multi-GB-countries stamps.

pat:
Oh, saw your edit! Ignore the first line of my reply 😅

marcoroth:
Resolved via #1075

PR #1075, opened 2025-10-18
Don’t expand GB attendance to all country stamps

pat:
We’ve been expanding attendance to any Great Britain conference to stamps for all the countries within it (England, Northern Ireland, Scotland, Wales).

This provides stamps that feel unearned/inaccurate, so let’s just use the default GB stamp. If there can be a way to flag events as being within specific GB countries, then the specific stamps can return.

As per discussion in #1073.


marcoroth (approved):
Thanks @pat, I think that makes sense! 🙏🏼

PR #1083, opened 2025-10-19
Improve RubyConf AU 2013-2019 data

pat:
Zach’s talk was incorrectly listed for 2013, hence shifting it to 2014


pat:
Not sure if/when I'm going to get to 2020, 2023 and 2024. I'm fine with this being merged in and the rest can be done in a separate PR. There's enough here to review as it is, sorry! 😅

marcoroth (approved):
This is great, thank you so much @pat! 🙏🏼

PR #1084, opened 2025-10-19
ensure github handle are search with case insensitive

adrienpoly:
We have a mix of downcased Github handle and lowered version. It started creating some duplicates of user.

Now :
- we preserve the GitHub handle casing
- search the lower cased version (there is also a dedicated index)

This ensure we don't create anymore duplicates and the index at the db level will enforce that


marcoroth:
/cc @ChaelCodes ☝🏼 

(she hit this "bug" on her last livestream)

adrienpoly:
yes I have seen a few cases in the admin and had to manually fix those. I fixed the last error with the and tested teh seed task locally with a dump from db. I think we should be better now

PR #1095, opened 2025-10-21
Fix the 'Future Events' section being empty when the user has no future events

### LINKED ISSUE #1092: "Future Events" section is empty when user doesn't have a future event
I guess in this case we should either show something like: "This person doesn't have any future events" or we should hide the section all-together.

Image

Tagging @karinevieira since this is related to #1081

karinevieira:
Resolves #1092

I forgot to hide the "Future Events" section when the user has no upcoming events. 
Thanks to @marcoroth for pointing this out!

Without future events:
Captura de Tela 2025-10-21 às 12 57 25

With future events:
Captura de Tela 2025-10-21 às 13 02 30

### DIFF
--- a/app/views/profiles/_events.html.erb
+++ b/app/views/profiles/_events.html.erb
@@ -28,7 +28,6 @@
             <h3 class="text-lg font-semibold mb-4">Past Events</h3>
             <div class="mb-6">
               <div class="grid min-w-full grid-cols-1 gap-8 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4">
-                <% past_events = events.past.sort_by(&:start_date).reverse %>
                 <% past_events.each do |event| %>
                   <%= cache event do %>
                     <%= render partial: "events/card", locals: {event: event, participation: participations[event.id]} %>


app/views/profiles/_events.html.erb
```diff
@@ -1,37 +1,43 @@
...
-              <% past_events = events.past.sort_by(&:start_date).reverse %>
-              <% past_events.each do |event| %>
-                <%= cache event do %>
-                  <%= render partial: "events/card", locals: {event: event, participation: participations[event.id]} %>
+        <% if past_events.any? %>
+          <div>
+            <h3 class="text-lg font-semibold mb-4">Past Events</h3>
+            <div class="mb-6">
+              <div class="grid min-w-full grid-cols-1 gap-8 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4">
+                <% past_events = events.past.sort_by(&:start_date).reverse %>
```
marcoroth: The seems to be duplicate to what we now have on L4
```suggestion
```

marcoroth:
Thanks for also handling the opposite, where we shouldn't show the "Past Events" section!

CleanShot 2025-10-21 at 18 14 40@2x

app/views/profiles/_events.html.erb
karinevieira: Thank you! I missed that 😅

karinevieira:
> Awesome, thank you so much for the quick fix @karinevieira! 🙏🏼

thank you so much for all your help @marcoroth!

PR #1096, opened 2025-10-22
[WIP] use local docker registery

adrienpoly:


config/deploy.staging.yml
```diff
@@ -2,7 +2,7 @@
 service: rubyvideo_staging
 
 # Name of the container image.
-image: adrienpoly/rubyvideo_staging
+image: rubyvideo_staging
```
marcoroth: should we use the opportunity to rename this to `rubyevents`?

config/deploy.yml
```diff
@@ -2,7 +2,7 @@
 service: rubyvideo
 
 # Name of the container image.
-image: adrienpoly/rubyvideo
+image: rubyvideo
```
marcoroth: same here

config/deploy.staging.yml
adrienpoly: yes that should be an easy switch

adrienpoly:
waiting for https://github.com/basecamp/kamal/pull/1679 to resolve and then I ll perform some more testings. I would like to be sure it doesn't end up filing the disk with zombi containers

marcoroth:
Closing in favor of #1600 and #1601

PR #1097, opened 2025-10-23
All Australian camps and retreats

pat:
No extended details, but it's a good starting point.


pat:
It’s a fair question! But in many cases, the numbers were indeed part of the name, more-so than the month/year - especially pre-COVID with the two-a-year cadence.

marcoroth:
> the numbers were indeed part of the name

Oh, never mind then. I thought that was more a measure to be able to uniquely index them here on RubyEvents.org.

marcoroth:
Thank you @pat! 🙏🏼

PR #1099, opened 2025-10-24
Add image proxy

hschne:
This is a follow up to #1049.  Loading GitHub avatars en-masse results in rubyevents.org getting rate-limited by GitHub. To avoid this, we would want to serve the images ourselves. This could be accomplished by using ActiveStorage (as suggested in #881).

This PR instead adds a proxy for image requests to GitHub. Images are downloaded once and served from cache afterwards. This is much easier to set up, but has similar downsides to the ActiveStorage solution. 

## Considerations

1. Initially, loading images is slow because we request them from GitHub. Once the cache is populated, requests are speedy enough. 
2. I opted to query for the user, which allows us to centralize avatar serving (by caching fallback images), but this adds a small overhead to serving images.
3. We might serve stale avatars, which is why TTL is set to 1 day.  Downloading/serving images via ActiveStorage has this issue too. We _could_ manually invalidate the cache under certain conditions (e.g. when the user updates their profile/relinks their GitHub...) but I don't think that's worth the effort. 
3. Follow-ups include refactoring the user model & avatar helper, or maybe proxying images from Bsky as well :thinking: 
4. Heads up, locally Rack Mini Profiler screws with the browser cache, so images are never served from the browser cache locally. 

## Alternatives

Apart from using ActiveStorage, we could use something like Cloudflare workers to do much the same thing, as outlined [here](https://alistairshepherd.uk/writing/cloudflare-worker-image-proxy/). I don't know enough about rubyevents.org's infrastructure to know if that's viable.

### DIFF
--- a/app/controllers/image_proxy_controller.rb
+++ b/app/controllers/image_proxy_controller.rb
@@ -1,0 +1,47 @@
+class ImageProxyController < ApplicationController
+  disable_analytics
+  skip_before_action :authenticate_user!
+
+  def github_avatar
+    user = User.find(params[:id])
+    size = params[:size] || 200
+
+    cache_key = "image_proxy/github_avatar/#{user.id}/#{size}"
+    avatar_data = Rails.cache.fetch(cache_key, expires_in: 1.day) do
+      fetch_github_avatar(user, size) || fetch_fallback_avatar(user, size)
+    end
+
+    if avatar_data
+      expires_in 1.day, public: true
+      fresh_when etag: cache_key, last_modified: 1.day.ago, public: true
+      send_data avatar_data,
+        type: "image/png",
+        disposition: "inline"
+    else
+      head :not_found
+    end
+  end
+
+  private
+
+  def fetch_github_avatar(user, size)
+    username = user.github_handle
+    return if username.empty?
+
+    url = "https://github.com/#{username}.png?size=#{size}"
+    fetch_image(url)
+  end
+
+  def fetch_fallback_avatar(user, size)
+    url_safe_initials = user.name.split(" ").map(&:first).join("+")
+    url = "https://ui-avatars.com/api/?name=#{url_safe_initials}&size=#{size}&background=DC133C&color=fff"
+    fetch_image(url)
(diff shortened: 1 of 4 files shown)


hschne:
@adrienpoly Here's a prototype implementation for cache-based GitHub avatars. Referring to [your previous comment](https://github.com/rubyevents/rubyevents/pull/1049#discussion_r2457099326).

adrienpoly:
That is an interesting solution I will probably push it to staging to test it in real life. Yet I am concerned that visiting the speakers page will generate a bloat of request for the server to handle where we would probably want those assets to be kind of static (or almost). Even if they are cached they still require a web request per image and that might be kind of slow.

If we would go for an active storage route we would need to be sure the storage provider support public routes so that we can serve the assets directly from the storage without going trough the web server

I ll look also in details about the cloudflare solution. We have cloudflare, but i am not a big expert of workers here

thanks for opening this PR to explore solutions

matiaskorhonen:
Alternatively, using [imgproxy](https://docs.imgproxy.net) could be an option. That would allow all third party images to be proxied and cached.

They have an official RubyGem with optional ActiveStorage integration: https://github.com/imgproxy/imgproxy.rb

```ruby
Imgproxy.url_for("http://images.example.com/images/image.jpg")
```

That would let the overhead from the caching to be offloaded to a separate process or even a separate VM. It has support for signed URLs so that prevents the image proxy from becoming an open proxy…

It would be similar to how GitHub handles external images in READMEs and other markdown content.

hschne:
I used Imgproxy in the past, can recommend. Running it should be easy peasy with Kamal, not sure about monitoring it though. In any case, more services always means more work ☠️

It's definitely a great idea if we want to go with a 'grown-up' solution.

marcoroth:
I'm going to close this PR for now. But also just to note, I'm not against adding something like [imgproxy](https://docs.imgproxy.net/) in the future, especially since we also have been experimenting with automatically generating thumbnails for talks, so a solution like imgproxy might be worth it in the future.

PR #1100, opened 2025-10-24
[Draft] Add Verified Attendance Tracking via Passport Scans

adrianthedev:
### Update

After a few conversations with Marco and Adrien, we figured out we need to refactor this in a different way.
We need a `VerifiedEventParticipation` model (`event_id:string connect_id:string verified_with:string details:json`) to hold all the passport scans.
Those would be linked to the user through the `ConnectedAccount` (which holds the passport id  of the user).

Then, compute on the fly the verified event attendance.

---

This PR implements a feature to verify event attendance through passport/badge scan imports and adds supporting infrastructure for attendance verification.

The importing functionality is temporary until we move the scanning capabilities inside the main app.

### Screenshots

http://localhost:3000/events/friendly-rb-2025/participants
CleanShot 2025-10-28 at 15 47 26
CleanShot 2025-10-28 at 15 47 41

http://localhost:3000/profiles/marcoroth
CleanShot 2025-10-28 at 15 45 46


app/avo/actions/import_passport_scans.rb
```diff
@@ -0,0 +1,70 @@
...
+        slug = case row["event"]
+        when "rails_world_2025"
+          "rails-world-2025"
+        when "friendly_25"
+          "friendly-rb-2025"
+        when "euruko_25"
+          "euruko-2025"
+          # when "prug"
+          #   "prug-2025"
+        end
```
marcoroth: Could we make it so that you can "Select Event" in the Avo Action, and then all rows in the CSV are just going to be assigned to the event you selected in the action?

adrianthedev: When we're importing, we're doing it for multiple events at once, so we need a mapping instead of selecting one event.
I could make the export for each event, but this whole import process is temporary until we merge the functionality in the main app.


BTW, it's late and I'm going to bed, which is something you should do too 😅

marcoroth: > When we're importing, we're doing it for multiple events at once, so we need a mapping instead of selecting one event.

I guess what I was trying to say is that it might make more sense to split up the data, so we don't have to hard-code the mapping in code.

> I could make the export for each event, but this whole import process is temporary until we merge the functionality in the main app.

But yeah, I guess even for the meantime, it would be easier if we didn't have to change the code and redeploy the app. Splitting up a CSV should be straightforward enough, but having to deploy new code to handle new events is quite a bit of work and annoying 🙈

adrianthedev: gotcha! makes sense. we'll do that 🙌 (splitting up the imports)

app/avo/resources/event.rb
```diff
@@ -1,8 +1,8 @@
 class Avo::Resources::Event < Avo::BaseResource
   self.includes = []
-  # self.search = {
-  #   query: -> { query.ransack(id_eq: params[:q], m: "or").result(distinct: false) }
-  # }
+  self.search = {
+    query: -> { query.where("lower(name) LIKE ? OR lower(slug) LIKE ?", "%#{params[:q]&.downcase}%", "%#{params[:q]&.downcase}%") }
```
adrianthedev, line 4: Unrelated to this PR: Added search to events.

config/initializers/avo.rb
```diff
@@ -51,6 +51,7 @@
   # config.per_page_steps = [12, 24, 48, 72]
   # config.via_per_page = 8
   # config.id_links_to_resource = false
+  config.click_row_to_view_record = true
```
adrianthedev, line 54: Unrelated: Quality of life improvement.

app/avo/actions/import_passport_scans.rb
```diff
@@ -0,0 +1,44 @@
...
+      rows.each do |row|
+        if row["connect_id"].present?
+          connected_account = connected_accounts[row["connect_id"]]
+          next if connected_account.blank?
+
+          user = connected_account.user
+          next if user.blank?
+
+          # Create or update attendance
+          attendance = EventParticipation.find_or_create_by!(user: user, event: event, attended_as: "visitor")
```
marcoroth, line 38: I think for speakers at these events (which already have a `EventParticipation` record associated for this event) this would create a second `EventParticipation`. Maybe that's fine? One to indicate you have verified attended and one for giving the talk?

adrianthedev: so, the way it does it with this PR, it will create another one as `visitor`.

I think that's fine as far as we remember that a user can have multiple event participations, which is how I modelled the "Vereified attendance" feature. 
From what I've seen, other bits of code do it like that too.

adrianthedev:
Closing in favor or https://github.com/rubyevents/rubyevents/pull/1613

Issue #1068, opened 2025-10-16
Changelog/Rubyevents News


hschne:
There's a lot of cool stuff happening in this repo, but it's a bit hard to follow. I'm subscribed to issues/PRs, and even so, I miss tons of new features/additions (almost didn't see #1051, which I think is _awesome_!). 

Apart from making it easier for interested parties to see what's happening, there's also the promotional angle. I'm fairly certain many Rubyists are unaware of the platform's capabilities, and making ongoing activity more transparent could be a great way to change that.

For starters, I can think of several solutions.
1. Automatically create GH releases with an auto-generated changelog (maybe on a weekly basis?) 
2. Create a changelog page/link on rubyevents.org to the changelog
3. Create a curated list of changes (think newsletter/post) to be publish on rubyevents.org
4. Highlight new features on the platform itself (think something like [driver.js](https://driverjs.com/)) 

Opening this issue to gather some input. 

@marcoroth @adrienpoly thoughts?

adrienpoly:
I fully agree 

> 3\. Create a curated list of changes (think newsletter/post) to be publish on rubyevents.org

I recently pitched the idea to Marco for an announcement page. Some kind of simple blog system with various categories (new features, new content etc)

at first I though this could be built on top of ActionText/Lexxy but maybe I was just trying a find an excuse to test Lexxy somewhere. Static pages in markdown is probably better as anyone can open a PR to create/review new content

Edit : and we already have some page rendered as markdown, it could be cool to add some embeds so that we could display in a post a talk, a collection of talks, an event by just addition some king of slug

hschne:
> Some kind of simple blog system with various categories (new features, new content etc)

Since y'all are using Avo already should be simple to get something like that up pretty quickly 🤔 

IMO 3. & 4. are less of a technical challenge, but it's tricky because that kind of stuff requires constant maintenance - someone needs to write that content at least semi-regularly after all. Just something to keep in mind.

adrienpoly:
I have been thinking more about this and for the announcement pages I think we should keep it as markdown content that gets updated with regular PR. This way anyone can contribute to this not just the admins

I really like the implementation @rossta did for https://github.com/joyofrails/joyofrails.com. I think we can probably take good part of it

PR #1103, opened 2025-10-25
Add automated weekly GitHub release

hschne:
This is an initial attempt to tackle #1068 and adds a GH action that runs weekly (or on manual trigger) to generate GitHub release notes. That's it. 

Since CI deploys continuously, there will be lag between the _actual_ release of features and when the GH release is created. Still, auto-generated releases should make it easier to let people follow what's been going on in the past week.

You can see the result/demo in this [dummy repository](https://github.com/hschne/release-test/releases/tag/25.10.25).

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,10 @@
+changelog:
+  exclude:
+    labels:
+      - dependencies
+    authors:
+      - dependabot
+  categories:
+    - title: New Content
+      labels:
+        - content
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,8 +20,4 @@ jobs:
       - name: Create release
         env:
           GH_TOKEN: ${{ github.token }}
-        run: |
-          TAG=$(date +'%y.%m.%d')
-          gh release create "$TAG" \
-            --title "$TAG" \
-            --generate-notes
+        run: bin/gh-release
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+name: Weekly Release
+
+on:
+  schedule:
+    - cron: '0 12 * * 5'  # Every Friday at noon UTC
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Create release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          TAG=$(date +'%y.%m.%d')
+          gh release create "$TAG" \
+            --title "$TAG" \
+            --generate-notes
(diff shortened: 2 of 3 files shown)


adrienpoly:
Do we have a way to :
- filter out dependabot PRs
- only create a release if there is content during that week? I would like to avoid having an automation creating empty releases if for a period of time we dontt have new PR merged for what ever reasons

hschne:
> I would like to avoid having an automation creating empty releases if for a period of time we dontt have new PR merged for what ever reasons

As if that is every going to happen :stuck_out_tongue: 

I updated the PR to include a [releases.yml](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes) which allows us to customize release notes. If we're a bit more diligent with PR labels, this should create nice release notes. 

I also created a `bin/gh-release` script to check if there are open PRs and such. I can't really test it - in particular because there are no previous releases. IMO we should manually create a release (just this one time) to have a baseline to start from. 

@adrienpoly If you create a manual release I should be able to do some proper testing :ok_hand:

marcoroth:
Thanks for exploring this! I'm not sure if GitHub releases are the right medium for this 🤔 

Ideally we could track the changes in the database itself, so that it can generate a "changelog" automatically. Maybe we could use the "seeding" mechanism as a trigger to compare what changed between before seeding and after and compile the changelog from that.

The tricky part is to make sure we don't track changes if only a typo gets fixed in the speaker name/title etc. This is also the reason why we avoided notifications and RSS feeds for now, as it's really tricky to get right.

The one thing we have now is an `id` key in the YAML files in `data/`, which should now provide a more stable key to track changes.

hschne:
Maybe linking #1068 is misleading. 

This PR covers the generation of GH release notes. We should have both - GH releases/release notes and a proper changelog. 

Release notes on GH are for people who look at the GitHub repository to see what actually changed in the code and when. This can and should include typo fixes and content PRs. Basically, a quick overview for contributors to see what's new. 

Changelog I'd consider curated content that lives in the database and/or Markdown files. @adrienpoly already mentioned doing something similar to Joy of Rails. This is intended for a broader audience - visitors of RubyEvents.org, not necessarily contributors. 

As to when we should generate what: 
- GH changelog should _ideally_ be generated on deploy. Maybe a [Kamal post-deploy hook](https://kamal-deploy.org/docs/hooks/post-deploy/)? Again, this is an initial attempt, so I went with the easy solution.
- Release notes should be created whenever there is something noteworthy to write about. Probably every couple of weeks or so. I'd avoid fully automating this - it will be hard to get rid of things like typo fixes or just non-noteworthy changes in general, like @marcoroth mentioned. 

TL;DR we should have GH release notes and a changelog, they are not the same thing and have different audiences.

marcoroth:
Given we have the monthly newsletter now (https://www.rubyevents.org/announcements) I'm going to close this one for now.

We might also be able to make the current featured section on the home page available more formally as "release notes", given that we are now able to better track changes and how our data changes. Something along the lines of:

- event announced
- CFP closes
- event coming up soon
- event wrapped up
- recordings published
- etc.

PR #1104, opened 2025-10-25
Clean GitHub workflows

hschne:
While doing #1103 I saw that some GH action dependencies are outdated and some of the configuration parameters/commands can be simplified :broom: 

- Rather than using `node_version` rely on the checked-in `.node-version` file, so CI node version and used node version stay in sync
- Use `db:prepare` to combine setup & schema load; combine rails test commands
- Remove 'bundler: default` as that's redundant

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 
 concurrency:
   group: ci-${{ github.ref }}
-  cancel-in-progress: true # Add this line to explicitly enable cancellation
+  cancel-in-progress: true
 
 jobs:
   lint:
@@ -17,18 +17,17 @@ jobs:
       RAILS_ENV: test
       RUBOCOP_CACHE_ROOT: tmp/rubocop
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Setup Ruby
         uses: ruby/setup-ruby@v1
         with:
-          bundler: default
           bundler-cache: true
 
       - name: Setup Node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
         with:
-          node-version: 20
+          node-version-file: ".node-version"
           cache: yarn
 
       - name: Install dependencies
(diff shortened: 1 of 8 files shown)


hschne:
Not sure what's up with those test failures. Looks to me like seeding tries to perform requests to GitHub?! :see_no_evil:

.github/workflows/ci.yml
```diff
@@ -120,9 +112,7 @@ jobs:
         run: bin/vite build --clear --mode=test
 
       - name: Prepare database
-        run: |
-          bin/rails db:create
-          bin/rails db:schema:load
+        run: bin/rails db:prepare
```
marcoroth: @hschne the test failures might be related to this change 🤔

fagnerpereira: sorry if my suggestion is silly, but there is a ci.rb where you might try some approaches to understand/investigate better the problem: https://rubyonrails.org/2025/10/22/rails-8-1

fagnerpereira: i tested locally using `db:test:prepare` and it worked
image

but with 
```
bin/rails db:create
bin/rails db:schema:load
``` 

resulted in errors indeed
image

fagnerpereira: again, sorry if im talking crazy, just trying to help with the investigation and learn more about it.

fagnerpereira: now im gonna try to solve an issue assigned to me, cause dont wanna disappoint my senpai @marcoroth hahahahah

hschne: @marcoroth Indeed. I overlooked that `db:prepare` loads the seeds, which the previous command did not do. Thanks for the hint, it's fixed now. 

@fagnerpereira I think `db:prepare` worked locally because you're not getting rate-limited by GH. Yet :wink:

fagnerpereira: but should we be calling this api in local environment? i mean, in production we are sure that fetch will be working. so why not by pass those external requests dependencies in development environment?
also I said about `db:test:prepare` not `db:prepare`

anyway if you found a solution, im with you. as I said before Im kind of dumb normally 🥲

marcoroth (approved):
Thanks @hschne! 🙏🏼

Issue #1091, opened 2025-10-20
Adopt `geocoder` gem


marcoroth:
For conferences, meetups, user locations, etc. it would be really useful to be operating on actual coordinates, so we can also show this data on an actual map, like #792.

https://github.com/alexreisner/geocoder

enderahmetyurt:
May I :)

marcoroth:
Please @enderahmetyurt :pray:

enderahmetyurt:
Hello nowadays I cannot find to time to focus that one. sorry :/ 🙏 someone would take care of it. @marcoroth

marcoroth:
Hey @enderahmetyurt, no worries! Looks like @hschne picked it up in #1183 🙌🏼

PR #1107, opened 2025-10-26
Refactor Profiles by splitting up tabs to their own resources

marcoroth:
This pull request refactors the profiles, so that each tab has its own resources, route, controller and views. This follows the pattern we have for the Events Page.

Additionally, this extract a `User::LocationInfo` associated object that should help with migrating to a proper `Location` object down the road with #1091.


adrienpoly:
I like the idea to split this into smaller view to keep the profile page fast and easy to maintain. Yet I feel like we are hiding lots of stuff in ProfileData and that for every requests/tab we always query all the data. 

I wonder if the TurboFrame version of this couldn’t be more efficient ?

marcoroth:
Yeah I was wondering about that too. The problem was that I didn't want to duplicate the logic to get the counts for each tab right.

PR #1113, opened 2025-10-28
Introduce `Alias` model for Canonical Users

marcoroth:
This pull request introduces the concept of "aliases" and introduces a new polymorphic `Alias` model that can be used for managing objects that might, or used to, have multiple names. 

To start, we use it for "Canonical Users", so we can match multiple "names" to a single identity in the platform. 

For example, Matz might be referred to with multiple names in the `videos.yml` files:

- Yukihiro Matsumoto
- Yukihiro "Matz" Matsumoto
- Yukihiro 'Matz' Matsumoto
- Matz

With the `Alias` model, we can use `User.find_by_name_or_alias` and map multiple names to the same `User` object. In the future we can also extend the functionality for Events (see #338), Talks, and Organizations.


fagnerpereira:
can i review your prs too @marcoroth ?

Issue #1079, opened 2025-10-18
Add Rails Hackathon


marcoroth:
https://railshackathon.com

marcoroth:
Related #1067

fagnerpereira:
may i get this one @marcoroth or would you suggest another one as my first contribution?

marcoroth:
Hey @fagnerpereira, you can have this one, this should be good to get started with.

fagnerpereira:
tks, gonna start and any doubts I let you know to avoid wrong approachs @marcoroth

fagnerpereira:
@marcoroth sorry if my doubt is dumb, but should we raise an API authentication? In this case, the Google API authentication, even for events where there is no YouTube channel or stuff like that. Sorry if I didn't understand well, but I'm going to share a screen. Maybe because it's weird for me because there is no YouTube content, but at the same time, if you run the script to create the playlist, it seems that it's required to have Google API key authentication, and this is a little confused me. 

https://app.warp.dev/block/tikcUryZfWvS097e28u89T

Maybe we should bypass those cases or it's the right thing to do? Like, to set up a Google API authentication key to set my local machine at a etc.

Firstly, I was trying to create everything manually, but I saw that there are scripts to prepare organizations and then to prepare the playlists. So when I tried to use them, I had this issue with authentication. I just want to confirm you with you that this is right, and it is something that we are required to configure or set up in our local environment to contribute.

Okay, again sorry if I understood something wrong, but it's just a little weird to me. But I have this question. Thanks for the support.

marcoroth:
Resolved via https://github.com/rubyevents/rubyevents/pull/1105

Issue #1067, opened 2025-10-16
Add new event `kind` for Camps, Retreats etc.


marcoroth:
> I wonder if we should add a new kind for Camps like the Ruby Retreat AU 2025 and others so we can have a proper kind for events like this. The other reason why a new kind would be useful is to show something else on the events/show page, as these camps are not focused on talks/speakers etc.

I completely agree. I think `camp`, `retreat`, or `unconference` could all be good names.

_Originally posted by @ChaelCodes in https://github.com/rubyevents/rubyevents/issues/1019#issuecomment-3408515864_

ChaelCodes:
# Proposal

We add a new kind `retreat`. This would cover code retreats, hackathons, and camps. It'll need to be added to organisation and event.
Event::StaticMetadata and Static::Playlist would support a new kind "retreat".
Retreats would be included in all event lists that currently filter by conference.
**What to call this scope?** I'm trying to come up with something that can incorporate future kinds. featured? without_meetup?  listed? notable? discoverable? ~~included_kinds?~~ ~~primary?~~
On the event page - participants and sponsors are fixed, talks, schedule, speakers, and CFP are only shown if data is present.
_We assume this information exists for conferences, and that if it's missing, we want a contributor to add it, however, these fields above might or might not exist for a retreat, and likely are lost forever if not captured contemporaneously._
`classify_event_size` in sponsors_controller should be updated to include retreats.
`home_start_date` in Static::Playlist would support retreats.
Update documentation in `contributing`.

What happens if we don't include a metadata_parser on the event? There's no youtube videos to parse.

## Excluded
No stamps added or modified for retreats
I don't think this will cover [unconferences](https://2024.rubyunconf.eu) which will have talks and speakers, and might even be recorded.

# Events to Update
Update organisations.yml and playlists.yml
_Organisations are currently used to represent an event series. Playlists are currently used to list events in an organisation._

- Ruby for Good
- Rails Camp South
- Rails Camp West
- Ruby Retreat NZ
- Ruby Retreat AU
- Ruby D Camp

# Testing

We group by `kind` on the cities page.
We filter by kind on the /events page.
We filter by kind on past events page as well.
We filter by kind on upcoming events page as well.
We filter by kind on the analytics dashboard.
We show different information based on kind on the event show page.
Sponsor statistics should include Retreat in Event Scale.
Missing sponsor page should include retreats.
Events archive should group by retreat kind.
Exclude retreats from Contributions/Not Published Talks

[...]

PR #1114, opened 2025-10-29
Add a new `hackathon` kind for events and organizations

marcoroth:
Similar to #1067, this pull request introduces a new `hackathon` kind for events and organizations. This will be useful for Events like #1079.


fagnerpereira:
@marcoroth I swear I added a type hackaton, but removed from the branch because I understood that all would be retreat 🥲
gonna update my branch here and change the kind for this value

PR #1132, opened 2025-11-22
Add SF Bay Area Ruby August 2025 Meetup recording

azimux:
Note: there's a thumb added for the open mic segment. If this all works smoothly I'd be happy to take a swing at adding a separate segment for each person who spoke during the open mic segment. But for now I left it as one segment to open this PR and get feedback and maybe merge it first.

Also, I manipulated the yaml data directly, which I assume is OK, but not sure if I did something I shouldn't have done as a result!


marcoroth (approved):
Thank you @azimux! 🙏🏼 

I just added the remaining Open Mic cues 🙌🏼

azimux:
Ooooh awesome!

PR #1136, opened 2025-11-26
Replace broken RubyConf India 2025 video links

Kerrick:
These are from the same channel as before. I'm not sure why the old ones disappeared. I found the new ones on [this playlist](https://www.youtube.com/playlist?list=PLbgP71NCXCqFZiVKZq5pFitVA7DELKJpA).


marcoroth:
Weird, thanks for catching that @Kerrick! 🙏🏼

marcoroth (approved):

PR #1142, opened 2025-11-27
Rename Tropical.rb series to Tropical on Rails

neilvcarvalho:
Tropical.rb has been renamed to Tropical on Rails in the 2025 event, and that's also the case in 2026. All their social media handles and website changed to reflect this rename.

This PR updates the mentioned data.

The event series handle hasn't been updated, still being `tropicalrb`.


marcoroth:
Thank you @neilvcarvalho!

PR #1144, opened 2025-12-01
SF Ruby - Involvements: added a couple more volunteers that were originally missed

stancheta:
Added Gennady and Emmanuel, who volunteered during check-in

CleanShot 2025-12-01 at 10 41 03


marcoroth (approved):
Great, thank you @stancheta! 🙏🏼

PR #1165, opened 2025-12-06
Add involvements for LRUG and Ruby Manor

h-lame:
Browsing the site to see the newly added LRUG stuff I spotted the concept of involvements and selfishly wanted to make sure my own contributions to the ruby community were represented.

On LRUG there is an interesting question for how meetup series and `involvements.yml` interact:

> Open question here is that this is the _current_ list of organisers, and 2 of
> them only started helping out from 2021 onwards. We've got a list of organisers
> who were involved in our meetings before 2021, but no longer. We could add them
> as a new "type" of organiser, but it'd link them to the whole series. I think
> this is likely due to all the meetings for a single meetup being listed as a
> single event so there's only space for one involvements.yml. The flip side of
> course being that creating a folder + event.yml, involvements.yml, videos.yml
> for each meeting is probably also onerous.
> 
> Perhaps we just need to allow for some date ranges in the involvements.yml and
> to enrich the backfill script.

Quoted from the commit message.

### DIFF
--- a/data/ruby-manor/ruby-manor-1/involvements.yml
+++ b/data/ruby-manor/ruby-manor-1/involvements.yml
@@ -1,0 +1,9 @@
+---
+- name: "Organizer"
+  users:
+    - James Adam
+    - Murray Steele
+
+- name: "Camera Equipment"
+  users:
+    - James Cox
--- a/data/ruby-manor/ruby-manor-2/involvements.yml
+++ b/data/ruby-manor/ruby-manor-2/involvements.yml
@@ -1,0 +1,6 @@
+---
+- name: "Organizer"
+  users:
+    - James Adam
+    - Kalvir Sandhu
+    - Murray Steele
--- a/data/ruby-manor/ruby-manor-3/involvements.yml
+++ b/data/ruby-manor/ruby-manor-3/involvements.yml
@@ -1,0 +1,10 @@
+---
+- name: "Organizer"
+  users:
+    - James Adam
+    - Murray Steele
+    - Tom Stuart
+
+- name: "Camera Equipment & Operation"
+  users:
+    - Andrew White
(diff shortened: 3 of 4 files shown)


data/lrug/lrug-meetup/involvements.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: "Organizer"
+  users:
+    - Alessandro Proserpio
+    - Chris Lowis
+    - Frederick Cheung
+    - Murray Steele
+    - Paolo Fabbri
```
marcoroth, line 8: That's a good point and something we can't properly support yet.

In the near future, we want to migrate the Meetups to the same structure that regular conferences have, so you could technically add an involvments.yml for each meetup event individually.

For now, I'd suggest that we add multiple entries to the file, something like:


```suggestion
---
- name: "2023 Meetups Organizer"
  users:
    - Alessandro Proserpio
    - Chris Lowis
    - Frederick Cheung
    - Murray Steele
    - Paolo Fabbri

- name: "2024 Meetups Organizer"
  users:
    - Alessandro Proserpio
    - Chris Lowis
    - Frederick Cheung
    - Murray Steele
    - Paolo Fabbri

- name: "2025 Meetups Organizer"
  users:
    - Alessandro Proserpio
    - Chris Lowis
    - Frederick Cheung
    - Murray Steele
    - Paolo Fabbri
```

or maybe less radical if it's easily group-able

```suggestion
---
- name: "20xx-2021 Meetups Organizer"
  users:
    - ...

- name: "2021-2025 Meetups Organizer"
  users:
    - Alessandro Proserpio
    - Chris Lowis
    - Frederick Cheung
    - Murray Steele
    - Paolo Fabbri
```

PR #1170, opened 2025-12-12
Set a github handle for azimux

azimux:
Not sure if this is entirely correct but works locally when running `bin/dev`


marcoroth (approved):
Thanks @azimux! 🙏🏼

PR #1176, opened 2025-12-16
Add thoughtbot Open Summit 2024 and 2025

ChaelCodes:
# Description

closes #1130 

Add thoughtbot Open Summit 2024 and 2025.

I used the default images, because their logo is a gif. :/

# Issues
I'm just going to document these here so I can add them to issues and investigate them later.

`rails runner scripts/create_events.rb my-conference` needs to add kind to event, or we need to document this needs to be added.
```
   value at `/year` is not an integer at /year
   object property at `/videos_count` is a disallowed additional property at /videos_count
   object property at `/metadata_parser` is a disallowed additional property at /metadata_parser
   object property at `/slug` is a disallowed additional property at /slug
```
app/clients/youtube/playlists.rb < At fault

`rails runner scripts/extract_videos.rb my-conference`
We need id on the video. There are 5 parsers - app/models/youtube/null_parser.rb - each would need to be updated. :/

PR #1179, opened 2025-12-16
Fix event import scripts and validate schema in bin/lint

### LINKED ISSUE #1174: `postal_code` in `venue.yml` is required to be a string, but `bin/lint` does not autocorrect it
# Description
When you're filling out a `venue.yml`, if you supply the `postal_code` without stringifying it, aka set it as a number, and run `bin/lint`, it does not autocorrect. When you run CI on the PR though, the `postal_code` raises an error because it's not a string.

Validate:all in CI uses JSONSchemer
https://github.com/rubyevents/rubyevents/blob/f8b9b625f2873016fc857341cbfd9514d96986dc/lib/tasks/validate.rake#L15-L26

bin/lint only uses `yarn format:yml`
https://github.com/rubyevents/rubyevents/blob/f8b9b625f2873016fc857341cbfd9514d96986dc/bin/lint#L18-L20

Which then calls `yarn run lint:yml --write`
https://github.com/rubyevents/rubyevents/blob/f8b9b625f2873016fc857341cbfd9514d96986dc/package.json#L29

ChaelCodes:
Fixes #1174
Fixes issue discovered in #1176

# Description
Currently, we validate the schema of all data files in CI, but we don't run it when we call `bin/lint`. This means we'll have formatting errors that don't surface until the branch is pushed and CI is run.

Hopefully, contributors are running `bin/lint` after they make their changes to fix formatting issues in their data files. This PR adds `bin/rails validate:all` which checks the schema in `bin/lint` which will at least surface the issues, if not fix them.

Also, the event creation script (scripts/create_events.rb) was adding `slug`, `videos_count`, and `metadata_parser` which are no longer permitted in the schema. This PR removes them.

This PR **does not** fix the missing id on videos. I think that might need more thought and care, because the titles and speaker names were inconsistent with a default run of the script.

### DIFF
--- a/bin/lint
+++ b/bin/lint
@@ -18,3 +18,6 @@ bundle exec erb_lint --lint-all --autocorrect
 # Check YAML file formatting
 echo "Checking YAML file formatting in data/ ..."
 yarn format:yml
+
+echo "Validating Schema of files in data/ ..."
+bin/rails validate:all


marcoroth (approved):
Very good observation @ChaelCodes, thanks for catching and fixing this! 🙏🏼

marcoroth:
> This PR does not fix the missing id on videos. I think that might need more thought and care, because the titles and speaker names were inconsistent with a default run of the script.

My rule of thumb for those have been `${first-name}-${last-name}-${event-slug}`, maybe it's something we can automate too to some degree later, especially since the `validate:all` rake task now makes sure these IDs are unique across all talks.

PR #1181, opened 2025-12-22
correct event name, 'RubyKaigi Takeout YYYY'

sorah:
Correcting the event names of our takeout series, which had happened during COVID-19 pandemic. Canonical form is that having year number after 'Takeout'. You can verify from its websites.

- @sorah on behalf of RubyKaigi Organizers

### DIFF
--- a/data/rubykaigi/rubykaigi-2020/event.yml
+++ b/data/rubykaigi/rubykaigi-2020/event.yml
@@ -1,10 +1,10 @@
 ---
 id: "PLbFmgWm555yZeLpdOLhYwORIF9UjBAFHw"
-title: "RubyKaigi 2020 Takeout"
+title: "RubyKaigi Takeout 2020"
 kind: "conference"
 location: "online"
 description: |-
-  https://rubykaigi.org/2020-takeout
+  https://rubykaigi.org/2020-takeout/
 published_at: "2020-09-03"
 start_date: "2020-09-04"
 end_date: "2020-09-05"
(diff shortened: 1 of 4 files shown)


sorah:
ah ok, I had initially thought so but after looked into https://github.com/rubyevents/rubyevents/blob/main/app/controllers/events/series_controller.rb#L16 and it is sorting by dates.

Honestly it sounds awkwerd or disappointable that the name has to be appeared wrong due to sorting reason, from perspective of event organizer, while I can agree sorting by name should work sufficiently in most cases.

(Can you again confirm that there are some codes that sorting events by its name?)

marcoroth:
Agreed, 100%. I was just trying to share the reason why it's currently "wrong".

A lot has changed in the RubyEvents.org codebase since we renamed the event. Back then we didn't have the "event dates" as columns on the "events" table, so the sorting was a lot trickier, which is why we relied on the name sorting. But since we have all the metadata now, we should be able to rework this and base if off of the event dates.

Let me check if we can change the code (or if it's already doing the right thing) and then get this pull request merged 👍🏼 

Thanks for pointing this out! 🙌🏼

marcoroth (approved):
Looks like we are all good, thanks @sorah! 🙏🏼

sorah:
Thank you for understanding and correction 👍 

(yup, I should have mention that I believe you should be aware that the situation is not right. My earlier comment has no intention of offence!)

PR #1183, opened 2025-12-22
Add event map to countries page 

### LINKED ISSUE #792: Meetup Map
It would be cool to also showcase the upcoming events on a map. Currently we don't have pinpoint locations for each event (only city names). But it might still be useful to start with just the city/country.

Image

Related: 
https://github.com/ruby-conferences/ruby-conferences.github.io/issues/562
https://rubyeurope.com
https://www.rubymap.com

### LINKED ISSUE #1091: Adopt `geocoder` gem

hschne:
This PR adds an event map to the countries view. This is meant as a first PoC for your consideration. For further plans and possibilities, check the follow-ups. 

image
 
##  Implementation

Latitude/Longitude columns, with sufficient precision for our needs, have been added to the event model. This is required for querying for events with available coordinates. Events with locations are then rendered as JSON via `map_controller.js` and displayed as markers on a MapLibre GL map. 

Events are rendered using Map markers to display their avatar icon. Events with identical coordinates are rendered using an avatar group. A scrollable pop-up allows users to view all events in a specific location.  

### Considerations

- I decided to go for map markers rather than symbol layers or similar because this allows for greater customizability. This is less performant and does not offer automatic collision detection or similar, but with the amount of data we are talking about here those downsides don't matter IMO. 
- Marker data is passed as JSON to allow easy extraction/reuse in the future at the cost of bloating the dom a bit.  Alternatively, creating a new controller action and fetching from it in Stimulus also works, but it involves a bit more boilerplate. 
- Coordinates are hard-coded in YAMLs, similar to venues rather than geocoding on the fly to avoid the need for keeping an API key around. 
- We extend `event.yml` to also allow coordinates (same as `venue.yml`) to provide a low-effort alternative to setting the location on an event. This eliminates the need to create a whole new file if you want to set coorindates for old events 
- During seeding & backfilling (using `geolocate files`) we take coordinates from a venue with coordinates if present and fall back to any event coordinates.  
- Integrated Geocoder because why not. This fixes #1091. 

## Testing & Backfilling 

New Thor tasks have been added to handle batch geocoding and backfilling of existing events. 

### Geocode Existing Event Files

First, provide a [google geocoding API key](https://developers.google.com/maps/documentation/geocoding/overview) in your `.env`. 

```bash
# .env
GEOLOCATE_API_KEY=<key> 
```

Run the `geolocate files` task, or optionally pass a file pattern you want to geocode. 
```bash
thor geolocate files
thor geolocate files data/vienna-rb/vienna-rb-meetup/event.yml
```

### Backfill Event Data

Second, run `geoloate events` task to backfill that data. This will take existing venues into account. Optionally overwrite events with existing latitude/longitude.

```bash
thor geolocate events
thor geolocate events --overwrite
```

Afterwards, go to http://localhost:3000/countries/ to check out the map. 

## Other Changes
- Added geocoder, parallel (for faster geocoding). 
- Upgraded reactionview to avoid [this bug](https://github.com/marcoroth/reactionview/issues/12)
- Initial geocoding applied to vienna-rb-meetup/event.yml.
- Sorted the `.env.local` file alphabetically 😅

## Follow-ups & Questions
- Obviously, we can extend this to fix #792 and also use #1040 to render a speaker map
- There's also the possibility of bringing this to the home view or other more prominent views. It might be useful to see where upcoming events are placed locally. Maybe a link to 'event map' on the homepage to show upcoming events on a map? :thinking:  
- We can use this map implementation (or a variation thereof) on the venues page to make it look nicer :wink:

### DIFF
--- a/data/blastoffrails/blastoffrails-2026/event.yml
+++ b/data/blastoffrails/blastoffrails-2026/event.yml
@@ -14,3 +14,6 @@ banner_background: "#fef0d4"
 featured_background: "#fef0d4"
 featured_color: "#b64023"
 website: "https://blastoffrails.com"
+coordinates:
+  latitude: 35.097735421515075
+  longitude: -106.6682148179672
--- a/data/blue-ridge-ruby/blue-ridge-ruby-2023/event.yml
+++ b/data/blue-ridge-ruby/blue-ridge-ruby-2023/event.yml
@@ -14,3 +14,6 @@ banner_background: "#fff"
 featured_background: "#E1EFFA"
 featured_color: "#0C2866"
 website: "https://blueridgeruby.com/2023"
+coordinates:
+  latitude: 35.59816209289096
+  longitude: -82.55246525892764
--- a/data/blue-ridge-ruby/blue-ridge-ruby-2024/event.yml
+++ b/data/blue-ridge-ruby/blue-ridge-ruby-2024/event.yml
@@ -14,3 +14,6 @@ banner_background: "#fff"
 featured_background: "#E1EFFA"
 featured_color: "#0C2866"
 website: "https://blueridgeruby.com/2024"
+coordinates:
+  latitude: 35.59491
+  longitude: -82.553696
--- a/data/blue-ridge-ruby/blue-ridge-ruby-2026/event.yml
+++ b/data/blue-ridge-ruby/blue-ridge-ruby-2026/event.yml
@@ -13,3 +13,6 @@ banner_background: "#fff"
 featured_background: "#E1EFFA"
 featured_color: "#0C2866"
 website: "https://blueridgeruby.com/"
+coordinates:
+  latitude: 35.593957
+  longitude: -82.550029
--- a/data/brightonruby/brightonruby-2024/event.yml
+++ b/data/brightonruby/brightonruby-2024/event.yml
@@ -13,3 +13,6 @@ banner_background: "#FFFFFF"
 featured_background: "#FFFBEB"
 featured_color: "#0F1729"
 website: "https://brightonruby.com/2024"
+coordinates:
+  latitude: 50.823661
+  longitude: -0.138391
--- a/data/brightonruby/brightonruby-2025/event.yml
+++ b/data/brightonruby/brightonruby-2025/event.yml
@@ -13,3 +13,6 @@ banner_background: "#FFFFFF"
 featured_background: "#FFFBEB"
 featured_color: "#0F1729"
 website: "https://brightonruby.com/2025"
+coordinates:
+  latitude: 50.823661
+  longitude: -0.138391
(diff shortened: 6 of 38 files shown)


app/schemas/event_schema.rb
```diff
@@ -50,4 +50,9 @@ class EventSchema < RubyLLM::Schema
   string :featured_color,
     description: "CSS text color for featured cards",
     required: false
+
+  object :coordinates, description: "Geographic coordinates", required: false do
+    number :latitude, description: "Latitude coordinate"
+    number :longitude, description: "Longitude coordinate"
+  end
```
marcoroth, line 57: Is there a reason why we copy this over to the `event.yml` file? I feel like keeping it in the `venue.yml` file helps to avoid duplication

hschne: My thinking was that it simplifies backfilling. 

A lot of events don't have venues yet, and adding a venue is more work than just adding coordinates to the event itself if done manually. It's also easier to automate. I was considering adding 'dummy venues' in the geolocate task instead of doing the event-coordinate thing, but what would that even look like? A bunch of venues with names identical to the event? Didn't much like the prospect of that :thinking:  

With the current approach, we have a low-effort way to generate coordinates for any event that has a location.  WDYT?

marcoroth: I was wondering if we should just write this to a `venue.yml` if there is no `venue.yml` yet instead of putting it in the `event.yml`. 

```yaml
coordinates:
  latitude: 35.59816209289096
  longitude: -82.55246525892764
```

Just so we only have one location where coordinates live. Or is that too confusing?

hschne: That's a possibility, of course, but it means that we have to adapt the venue schema to make the name non-required. Or add a dummy name. 

```ruby
class VenueSchema < RubyLLM::Schema
  string :name, description: "Name of the venue"
  ...
```

Then, we'd also have to update the event navigation to distinguish 'dummy' venues (with only coordinates) and 'real' venues, because I don't think it makes sense to render a venue in the event tab that only has coordinates? Or maybe it does? :thinking: 

```ruby
<% if event.conference? || event.venue.exist? %>
  <%= active_link_to event_venue_path(event), class: class_names("tab", {"text-red-500" => !event.venue.exist?}), active_class: "tab-active", role: "tab" do %>
    Venue
    <% unless event.venue.exist? %>
      <%= fa("question-circle", size: :xs, class: "ml-1 fill-red-500") %>
    <% end %>
  <% end %>
<% end %>
```

Or we just say fuckit, give each venue a generic name, and just roll with that. 

```yml
name: <Event Name> Venue
coordinates:
  latitude: 35.59816209289096
  longitude: -82.55246525892764
```

But again, that means generating a lot of venue files for events where we don't actually have venue data - just for coordinates. I don't know about that. If it's just about consolidating location data into venues, you could make the same argument about `event.location`. It's also redundant, considering we have the venue address. 

I much prefer the current approach, for the reasons outlined above. That being said, I'm happy to go with whatever solution you choose :wink:

marcoroth: Yeah, fair enough! Was not trying to be annoying, I just wanted to explore if there's a better way to go about it! 👍🏼 

Let's continue with `event.yml` approach then 👍🏼

PR #1185, opened 2025-12-22
Add tiny ruby #{conf} 2026

matiaskorhonen:
Adds `tiny ruby #{conf}` 2026. We're planning on it happening in October but don't have a precise date yet…


marcoroth (approved):
Awesome, thank you @matiaskorhonen! 🙏🏼

PR #1187, opened 2025-12-23
SF Ruby Schedule updates and new Talk kind of Demo

ChaelCodes:
# Description

- Added SF Ruby Schedule changes, including "Thank you" from Irina, 25 demos, and 2 talk changes
- Introduce talk kind of 'demo' to support 25 startup demos at SF Ruby Conf
- Update VideoSchema to require video_id and video_provider for sub-talks, the db:seed fails without those fields 
- Updated one talk "RAILS_ENV=demo" that would have been falsely classified as a demo


ChaelCodes:
Thanks! 😊

PR #1191, opened 2025-12-25
Upgrade to Ruby 4.0.0

marcoroth:
https://www.ruby-lang.org/en/news/2025/12/25/ruby-4-0-0-released/


adrienpoly:
It does work great in dev but it fails to deploy initialy the first deploy failure was related to the docker image not being published (I think). I restarted the action but it still fails 

the error is now :  

```
An error occurred while installing commonmarker (2.6.0), and Bundler cannot continue
```

I think we should revert it for now and try to get it to deploy on staging first (with the Github action)

adrienpoly:
probably related to this https://github.com/gjtorikian/commonmarker/issues/427

are we still using this gem?

marcoroth:
I guess it's coming from https://github.com/avo-hq/marksmith?tab=readme-ov-file#1-add-marksmith-to-your-gemfile

But, we can also remove it for now until it's re-released

PR #1194, opened 2025-12-27
Add SF Ruby Conference 2025 recordings

ChaelCodes:
# Description

Add first batch of recordings from SF Ruby (Main Stage videos).

The new title of "Startup Demo" wasn't matching the kind of "demo" so anyone who gave a demo showed up twice - once in speaker and once in demo. The parent talk has one kind, and the sub-talks have a different kind. If these two kinds don't match, the speakers show up twice, once in each kind. (I suspect you know this, but am adding it in case someone else debugs this issue in the future.)

Added YouTube values for playlist and channel.


marcoroth (approved):
Thank you! 🙏🏼

PR #1205, opened 2025-12-30
Add Hackathons and Workshops to Events Archive header

ChaelCodes:
# Description
We added kinds of Hackathon and Workshop but we didn't add them to the header, so the count on the front page and the count on the archive don't match up!

It's still off by 15, but I think that might be the canonical filter.

# Screenshots
Screenshot 2025-12-30 at 3 59 00 PM
Screenshot 2025-12-30 at 3 59 14 PM


ChaelCodes:
Thanks. :)

PR #1219, opened 2026-01-03
Allow talk feedback and improve watched talks managment

marcoroth:
This pull request adds the ability for viewers to share feedback with speakers. I spent some time thinking about how to make this constructive rather than just another rating system.

https://github.com/user-attachments/assets/ca9f7561-0134-4c40-b77d-57e8773ebc2f

Traditional approaches felt problematic:
  - 1-5 star ratings don't encourage reflection, people just click and move on
  - Like/dislike is too binary and doesn't give speakers actionable insight

Instead, I went with specific yes/no questions that prompt viewers to think about concrete aspects: "Did you enjoy it?", "Would you recommend it?", "Is it beginner-friendly?" These give speakers useful signal without enabling pile-ons.

The emoji "feeling" picker adds a bit of personality while keeping things positive, even "Neutral" and "Not for me" are honest without being harsh. 

Speakers can also opt-out entirely in their settings if they'd prefer not to receive feedback.

CleanShot 2026-01-03 at 16 03 21@2x

Also, somehow ended up tweaking the video player and thumbnail styles 😅

CleanShot 2026-01-03 at 16 06 54@2x

CleanShot 2026-01-03 at 16 06 47@2x

CleanShot 2026-01-03 at 16 07 22@2x

On the talks page we now also offer to "hide watched talks":

CleanShot 2026-01-03 at 16 07 37@2x

And also revamped the styles on the "Watched videos" page, including a new "continue watching" section:

CleanShot 2026-01-03 at 16 07 45@2x


ChaelCodes:
This shows so much care and thought around the speaker and viewer experience. I genuinely love that and the emotion selection and I'm glad there's an opt-out!

I really like the questions instead of a rating or like/dislike, but I wonder how actionable they are?
Let's say 5 people didn't enjoy the talk. As a speaker, I still don't know why.
And the emotion selection at the top communicated whether they enjoyed it and how.

Maybe things like:
 - Were the slides easy to read and understand?
 - Was the content relevant or helpful to you?
 - Was the speaker engaging and easy to understand?

Instead of "Good for early career devs?" which biases things towards early career devs, maybe it should have a level? "What level of experience do you need to understand this talk?" (Beginner | Intermediate | Expert) I'm thinking particularly about Kaigi with this one.

I liked these existing question the most.
- Did you learn something new?
- Was it easy to follow?
- Did it inspire you to learn more?

adrienpoly:
> Instead of "Good for early career devs?" which biases things towards early career devs, maybe it should have a level? "What level of experience do you need to understand this talk?" (Beginner | Intermediate | Expert) I'm thinking particularly about Kaigi with this one.

I like this question as it is also a way to improve our metadata and potentially in the future a new search filter option

marcoroth:
Thanks for the feedback, I updated it to have a few more questions and different answers for some of them: 

CleanShot 2026-01-03 at 22 23 51@2x

We can tweak these questions in the future if we feel like there are too many.

PR #1247, opened 2026-01-06
Geocode users

marcoroth:
Similar to `Event` in https://github.com/rubyevents/rubyevents/pull/1183, this pull request adds support for geocoding users.

This should also help with https://github.com/rubyevents/rubyevents/pull/1246

It also allows us to improve the countries/cities pages later and allows us to show "Ruby friends near you" and similar.


adrienpoly:
what do you think to extract geocoding data to it's own table to avoid growing too much the user table over time ?

marcoroth:
Yeah, that could be a good idea, so we could also make it polymorphic for events etc.

PR #1249, opened 2026-01-06
Improve country, state, and city pages

marcoroth:
Follow up on #1247 and #1248.

This pull request adds the `geocoded_by` setup to the `User` and `Event` model. This allows us to use the `User.near` and `Event.near` methods.

We use these to show nearby events:

CleanShot 2026-01-06 at 17 29 05@2x

And nearby Rubyists:

CleanShot 2026-01-06 at 17 28 55@2x

Additionally, we now have a proper understanding of `City`, `State` and `Country` and can specifically link to them from the profiles and from the event pages:

CleanShot 2026-01-06 at 17 30 40@2x

CleanShot 2026-01-06 at 17 30 45@2x

To make sure that always we always have proper coordinates for popular cities (and make sure the slugs are unique) we have a concept of `FeaturedCity`. We have a few popular in the seeds, but each city (found in a user profile or for an event) can be promoted to be a `FeaturedCity`:

CleanShot 2026-01-06 at 17 33 45@2x

We also got some new pages, like 

**`/states`:**

CleanShot 2026-01-06 at 17 34 02@2x

**`/states/us`**

CleanShot 2026-01-06 at 17 34 28@2x

**`/states/us/north-carolina`**

CleanShot 2026-01-06 at 17 34 40@2x

### DIFF
--- a/data/scottish-ruby-conf/scottish-ruby-conf-2013/event.yml
+++ b/data/scottish-ruby-conf/scottish-ruby-conf-2013/event.yml
@@ -2,7 +2,7 @@
 id: "scottish-ruby-conf-2013"
 title: "Scottish Ruby Conf 2013"
 description: ""
-location: "Perthshire, Scotland"
+location: "Perthshire, Scotland, UK"
 kind: "conference"
 start_date: "2013-05-12"
 end_date: "2013-05-13"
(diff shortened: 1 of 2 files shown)


ChaelCodes:
Can we move Meetups to the top of location pages? If you're looking to get involved with your local community, that's what you're looking for. Same goes for organizers looking to promote an event.

marcoroth:
Yes, I think that makes sense 👍🏼 

The idea with this is that we get the meetups as individual events, and then we can also better represent them on the site. I think we also want to limit the amount of "conferences", "events" etc. we want to show on the countries/cities overview page, especially if events happened a long time in the past.

I sketched this up earlier while working on this PR:

CleanShot 2026-01-06 at 18 18 16@2x

PR #1251, opened 2026-01-06
Flag suspicious user and hide their profile detail

marcoroth:
We've got a few spam accounts that are prompting things on their RubyEvents.org profile. Since they can write their own bio and add a website it's pretty much giving them a backlink for free.

This pull request adds `User#suspicious?` and hides most of the user profile (including their avatar) when it's being detected as suspicious, we also avoid setting and SEO tags as long as somebody is detected as suspicious.

CleanShot 2026-01-06 at 23 33 31@2x

It also adds an Avo filter and action to clear a user that has been falsely flagged.

### DIFF
--- a/app/avo/resources/user.rb
+++ b/app/avo/resources/user.rb
@@ -27,7 +27,8 @@ class Avo::Resources::User < Avo::BaseResource
     field :github_handle, as: :text, link_to_record: true
     field :admin, as: :boolean
     field :marked_for_deletion, as: :boolean, hide_on: :index
-    field :cleared_at, as: :date_time, hide_on: :index
+    field :suspicion_marked_at, as: :date_time, hide_on: :index
+    field :suspicion_cleared_at, as: :date_time, hide_on: :index
 
     field :slug, as: :text, hide_on: :index
     field :bio, as: :textarea, hide_on: :index


app/models/user.rb
```diff
@@ -10,6 +10,7 @@
 #  bsky                :string           default(""), not null
 #  bsky_metadata       :json             not null
 #  city                :string
+#  cleared_at          :datetime
```
ChaelCodes: cleared of suspicion at? Not sure cleared_at is clear out of context.

ChaelCodes: Maybe we should store that we think they're suspicious for easy querying.

marcoroth: Good point 👍🏼 

I think inverting it actually makes sense, so that we have a `marked_suspicious_at` column. That way the `suspicious?` call is also more performant and doesn't need to be calculated on the fly for every time we want to check it.

ChaelCodes: Yeah, good call on that `mark_suspicious_at`. I think we still need a `admin_verified_at` or `suspicion_cleared_at` so they don't get marked suspicious again. I assume we're going to run a rake task or something to check if we have any suspicious users?

marcoroth: I have a `suspicion_marked_at` and `suspicion_cleared_at` column, and a `Recurring::MarkSuspiciousUsersJob` that runs every 6 hours with Solid Queue.

ChaelCodes: This is going to be awesome! You got this out so fast.

PR #1253, opened 2026-01-07
Redesign country, state, and city pages and add continent pages

marcoroth:
CleanShot 2026-01-07 at 14 50 08@2x

CleanShot 2026-01-07 at 14 51 00@2x

CleanShot 2026-01-07 at 14 51 49@2x

CleanShot 2026-01-07 at 14 52 33@2x


app/components/ui/avatar_component.html.erb
```diff
@@ -1,15 +1,29 @@
...
-        <%= image_tag avatarable.avatar_url(size: image_size), loading: "lazy", onerror: "this.previousElementSibling.classList.remove('hidden'); this.remove()" %>
-      <% end %>
+<% if show_hover_card? %>
+  <div class="group/card relative" data-controller="hover-card">
+    <% if show_link? %>
+      <%= link_to link_path do %><%= avatar_markup %><% end %>
+    <% else %>
+      <%= avatar_markup %>
     <% end %>
+    <%= render "users/hover_card", user: avatarable %>
```
adrienpoly: I think this approach could result in rendering a lot of hidden HTML, which may negatively impact page performance. Especially if there are many avatars on the same page.

I’d recommend loading the hover card content lazily via a backend request when the user hovers over an avatar instead.

app/javascript/controllers/hover_card_controller.js
```diff
@@ -0,0 +1,38 @@
+import { Controller } from '@hotwired/stimulus'
```
adrienpoly, line 1: maybe we could use the html popover now for this?

app/components/ui/avatar_component.html.erb
marcoroth: Good point, I made them load lazily using Turbo Frames 👍🏼

app/javascript/controllers/hover_card_controller.js
marcoroth: Not 100% sure. I was looking to make sure we don't load the lazy hover card on initial page load, that it works on hover (not clicking), and that they are positioned right (not off-screen). But I couldn't figure it out without JavaScript.  Though, this refactor to extract to make the hover cards lazy simplified the JavaScript controller quite a bit.

PR #1266, opened 2026-01-08
import_involvements! mutates instead of destroys.

ChaelCodes:
# Description

Use nested attributes to mutate the records instead of destroying them. As a bonus, we do fewer queries and saves!

# Tested

Renaming a role (sends it to bottom of the page)
Deleting a person
Adding new people

### DIFF
--- a/app/models/static/event.rb
+++ b/app/models/static/event.rb
@@ -274,6 +274,20 @@ module Static
     def import!(index: SEARCH_INDEX_ON_IMPORT_DEFAULT)
       return if Rails.env.test? && !ENV["SEED_SMOKE_TEST"] # this method slowdown a lot of the test suite
 
+      event = import_event!
+
+      import_cfps!(event)
+      import_videos!(event, index: index)
+      import_sponsors!(event)
+      import_involvements!(event)
+      import_transcripts!(event)
+
+      Search::Backend.index(event) if index
+
+      event
+    end
+
+    def import_event!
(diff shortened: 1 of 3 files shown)


app/models/static/event.rb
```diff
@@ -274,6 +274,20 @@ def static_series
     def import!(index: SEARCH_INDEX_ON_IMPORT_DEFAULT)
       return if Rails.env.test? && !ENV["SEED_SMOKE_TEST"] # this method slowdown a lot of the test suite
 
+      event = import_event!
```
ChaelCodes, line 277: Moved event creation in a separate method. It's cleaner and made testing possible.

app/models/static/event.rb
```diff
@@ -428,17 +437,28 @@ def import_involvements!(event)
           user = ::User.find_by_name_or_alias(user_name)
 
           unless user
+            # raise "User '#{user_name}' not found in speakers.yml" if Rails.env.development?
```
ChaelCodes, line 440: One day, but not today.

app/models/static/event.rb
```diff
@@ -415,7 +421,10 @@ def involvements_file?
     def import_involvements!(event)
       return unless involvements_file?
 
-      event.event_involvements.destroy_all
```
ChaelCodes, line 418: This is the main reason I did this. We were recreating all our event involvements every seed, for every event.

PR #1267, opened 2026-01-09
Improve country page map to show full country/continent bounds

khasinski:
- Always pass location bounds to map when available
- Use visible layer's bounds on initial load (e.g., continent bounds when country has no events)
- Fit to selected layer's bounds when switching tabs
- Add continent bounds to geo-continent layer

### DIFF
--- a/app/controllers/concerns/geo_map_layers.rb
+++ b/app/controllers/concerns/geo_map_layers.rb
@@ -60,22 +60,22 @@ module GeoMapLayers
   def add_country_layer(layers)
     return unless @country.present?
 
-    maybe_add_layer(layers, id: "geo-country", label: @country.name, emoji: @country.emoji_flag, events: @country.events.includes(:series))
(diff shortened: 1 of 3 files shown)


ChaelCodes:
Hey @khasinski ! I'm pretty new to geolocation. Would you mind explaining how this will help our current maps and what I should be looking for on the country page when reviewing this?

khasinski:
Hey @ChaelCodes! This PR basically sets the default zoom level and boundaries for each event page to show the entire country. Current version defaults to showing an area based on all the visible conferences, which for some countries shows a tiny fragment of them. 

Before - Poland :
image

After - Poland:
image

This also takes case of the edge case where show conferences in neighbouring countries if there isn't one in the country we're looking at.

Before - Zambia, shows conference in Kenya, zoomed enough so you might miss it's in another country:
image

After - Zambia, shows map of Africa so you can see where is the nearest conference:
image

ChaelCodes:
This looks fantastic! Thank you!

PR #1292, opened 2026-01-12
Use Gum gem for generating Event Assets

ChaelCodes:
# Description

Instead of installing gum, use the gum ruby gem to generate assets.
I hit a minor roadbump while installing gum, and decided rewriting the rake task to use the ruby gem was easier then figuring out apt-get.


marcoroth:
Yay, thank you for updating this! 🙌🏼

PR #1294, opened 2026-01-13
Replace all umlauts and strip out the region

ChaelCodes:
# Description

There are two Zurich's - Zurich(zurich) and Zürich(zuerich)(with a region of Zürich).

This is causing an error when we run seeds and load the featured cities.

The featured city list (made by Marco!) says Zurich, so I'm taking him as a canonical source.

All Zürich's are updated to Zurich, and the region is replaced by "".

# Admin Tasks
- [ ] Delete Zürich
- [ ] Rerun seeds


marcoroth:
The problem is that the different lookups return different cities names:

```ruby
rubyevents(dev):006> Geocoder.search("Zurich", lookup: :nominatim).first.city
=> "Zurich"

rubyevents(dev):007> Geocoder.search("Zürich", lookup: :nominatim).first.city
=> "Zurich"

rubyevents(dev):008> Geocoder.search("Zurich", lookup: :google).first.city
=> "Zürich"

rubyevents(dev):009> Geocoder.search("Zürich", lookup: :google).first.city
=> "Zürich"
```

I think what we need to do is to account for the aliases that we defined in `featured_cities.yml`, so that we can map them back to the same city record.

PR #1295, opened 2026-01-13
Report which City has a Duplicate name

ChaelCodes:
# Description

Add which city has a duplicate name to the Rails log.


ChaelCodes:
The bonus effect here is that the docker image failed to deploy, so I think we weren't running the new code in the seed. This might pass the "Build Docker Image" in prod, just by retriggering it.

Issue #1256, opened 2026-01-07
Talks#show page view transition is broken


adrienpoly:
The pageview transition from the index to the show page is currently broken 
in the dev console we have this error message, we probably introduced a duplicate

<img width="920" height="270" alt="Image" src="https://github.com/user-attachments/assets/96c87639-b290-4bff-bf9b-c67ab4a631f2" />

Might be related also to #1254

PR #1305, opened 2026-01-13
Remove `preserve-scroll` Stimulus Controller

marcoroth:
This Stimulus Controller is quite confusing when navigating pages, as you are not being navigated to the top of a page.


adrienpoly:
It was introduced to preserve the page transition when opening a talk after having scrolled and then going back to the index

Now since the page transitions are broken (#1256) we cannot really see the effect 

Here is a quick demo of the problem it solves

## With (and with the transition back)

https://github.com/user-attachments/assets/0855c051-d23f-47a7-9e84-520312eb4558

## Without

https://github.com/user-attachments/assets/1da73200-0115-489d-9fda-8eca0578b89d

When going back the page transition is kid of strange

Not saying we should keep it but maybe use an alternative approach to prevent this strange effect when going back

marcoroth:
That's a good point, and I think that makes sense for that use case! Somehow there's something off, that it still "restores" a scroll position even though it's the first time you visit that page, I'll have to check if I can reproduce it.

PR #1329, opened 2026-01-16
Create a new Generator for adding a CFP

ChaelCodes:
# Description

Usage:

```
  bin/rails generate cfp --event-series=series-name --event=event-name
    --name="Call for Proposals" --link="https://example.com/cfp"
    --open-date="2026-01-01" --close-date="2026-12-31"
```

The new Generator is validated against CFPSchema and documented in ADDING_CFPS.md which is included in the template CFP file.

Obviously, it's not particularly difficult to add a CFP file, but that's actually why I started here. My goal is to have generators for all our event files with overridden defaults.

Then, theoretically, we create an event with a generator, and it makes all the necessary files.

As these generators become more powerful, we could potentially replace some of the tools with calls to a generator, but right now, this one can only handle one CFP, not multiple.

# Benefits

- Idiomatic Rails interface for Contributors
- Available without MCP tools
- Formatted correctly and passes linter on generation
- Instead of templates in documentation that fall out of date, the template is verified against the schema

# Resources

- https://guides.rubyonrails.org/generators.html
- https://api.rubyonrails.org/v8.1.2/classes/Rails/Generators/Testing/Assertions.html
- https://www.rubydoc.info/gems/thor/Thor%2FBase%2FClassMethods:class_option
- https://garrettdimon.com/journal/posts/creating-custom-rails-generators

### DIFF
--- a/docs/ADDING_CFPS.md
+++ b/docs/ADDING_CFPS.md
@@ -1,0 +1,116 @@
+# Adding Call for Proposals to RubyEvents
+
+This guide explains how to add a CFP for conferences and events in the RubyEvents platform.
+
+## Overview
+
+CFP data is stored in YAML files within the conference/event directories. Each conference can have its own cfp file that describes multiple CFPs.
+
+## File Structure
+
+CFPs are stored in YAML files at:
+```
+data/{series-name}/{event-name}/cfp.yml
+```
+
+For example:
+- [`data/blue-ridge-ruby/blue-ridge-ruby-2026/cfp.yml`](/data/blue-ridge-ruby/blue-ridge-ruby-2026/cfp.yml)
+- [`data/sfruby/sfruby-2025/cfp.yml`](/data/sfruby/sfruby-2025/cfp.yml)
+
+All permitted fields are defined in [CFPSchema.](/app/schemas/cfp_schema.rb)
+
+## Generation
+
+Generate a CFP using the CfpGenerator!
+
+```
+bin/rails g cfp --event-series=blue-ridge-ruby --event=blue-ridge-ruby-2026 --name="Call for Proposals" --link=https://blueridgeruby.com/cfp --start-date=2025-12-15 --end-date=2026-02-03
+```
+
+Check the usage instructions using help.
+
+```
+bin/rails g cfp --help
+```
+
+This will create a cfp.yml with a single CFP in it.
+If you need to add additional CFPs for start-up demos or lightning talks, add another cfp using the same format to the array.
+
+## Step-by-Step Guide
+
+### 1. Check for Existing CFP File
+
+First, check if a cfp file already exists:
+
+```bash
+ls data/{series-name}/{event}/cfp.yml
+```
+
+### 2. Create or Edit the cfp File
+
+If the file doesn't exist, create it:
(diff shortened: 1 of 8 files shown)


docs/ADDING_CFPS.md
```diff
@@ -0,0 +1,116 @@
...
+### Common Issues
+
+- **Invalid YAML syntax**: Check indentation (use spaces, not tabs)
+- **Missing required fields**: Ensure all required properties are present
+
+## Submission Process
+
+1. Fork the RubyEvents repository
+2. Setup your dev environment following the steps in [CONTRIBUTING](/CONTRIBUTING.md)
+3. Create your sponsors file in the appropriate directory
```
ChaelCodes:
```suggestion
3. Create your cfp file in the appropriate directory
```

docs/ADDING_CFPS.md
```diff
@@ -0,0 +1,116 @@
...
+4. Run `bin/lint`
+5. Run `bin/rails db:seed` (or `bin/rails db:seed:all` if the event happened more than 6 months ago)
+6. Run `bin/dev` and review the event on your dev server
+7. Submit a pull request
+
+## Need Help?
+
+If you have questions about contributing sponsors:
+- Open an issue on GitHub
+- Check existing sponsors files for examples
```
ChaelCodes:
```suggestion
- Check existing cfp files for examples
```

ChaelCodes:
Interesting discovery - if you run the generator and there are changes, it'll let you do a diff if you have the right tool.

> This message is displayed because 'diff.tool' is not configured.
> See 'git difftool --tool-help' or 'git help config' for more details.
> 'git difftool' will now attempt to use one of the following tools:
> kompare emerge vimdiff nvimdiff
> No known diff tool is available.

marcoroth:
This is a really great idea!

I wonder if we can make the `CFPCreateTool` also depend on the internals of the generator, so that both the MCP tools and the generators use the same underlying code, so that it works with both in the same way. Or, that the MCP tool just calls the generator by shelling out.

A first attempt for this was with [`app/models/event/cfp_file.rb`](https://github.com/rubyevents/rubyevents/blob/50b07ec88ca7b5b0085d327302bd36e2e5f85cb7/app/models/event/cfp_file.rb), but I like the idea of the `.tt.yml` files from the generators a lot more!

ChaelCodes:
> This is a really great idea!
> 
> I wonder if we can make the `CFPCreateTool` also depend on the internals of the generator, so that both the MCP tools and the generators use the same underlying code, so that it works with both in the same way. Or, that the MCP tool just calls the generator by shelling out.
> 
> A first attempt for this was with [`app/models/event/cfp_file.rb`](https://github.com/rubyevents/rubyevents/blob/50b07ec88ca7b5b0085d327302bd36e2e5f85cb7/app/models/event/cfp_file.rb), but I like the idea of the `.tt.yml` files from the generators a lot more!

That was one of my long-term goals! We can call the generator with Ruby with no issues. 

I'm thinking about trying append_file to handle multiple CFPs or gsub to update. 

Or just rerender the whole file with an array of cfps.

I just saved it for another day.

Issue #1274, opened 2026-01-09
RubyConf 2022 Mini dates are wrong


marcoroth:
The talks of the RubyConf 2022 Mini conference has the wrong dates

https://www.rubyevents.org/events/rubyconf-2022-mini

PR #1330, opened 2026-01-16
Fix dates in Rubyconf mini 2022

viniciusalonso:
Issue #1274


ChaelCodes:
Thank you for updating the dates!! 🙌

By the way, where did you get the dates from? Is there a schedule link out there? 

If we can put them in the right order and add a schedule.yml, they'll display with a schedule on the site! ✨

marcoroth:
I added some links I found in 90287625b7a275b7fc648d46519cbea17e9bb3ba

Issue #1257, opened 2026-01-07
Known Participants CTA reflects whether the event has happened


ChaelCodes:
# Description

The "Known Participants" CTA says "Were you there?" even if it's before the event has happened. Let's update that to say "Are you going?" until it's after the event's end_date.

<img width="1269" height="159" alt="Image" src="https://github.com/user-attachments/assets/ebfee491-dd4c-4b39-aa73-00b4d8249fa2" />

> [!CAUTION]
> Keep an eye out for Meetups which do not have start_date or end_dates.

marcoroth:
Good catch! 🙌🏼 

The button next to it changes it's copy based on if it's an event in the future/past: https://github.com/rubyevents/rubyevents/blob/29234813f61bec0570fed5eff72ba1389d0cd53e/app/views/events/_participation_button.html.erb#L1-L2

But we should probably account for this fact in the "Were you there?" copy too

PR #1338, opened 2026-01-17
Adapt button participant text if event happened

viniciusalonso:
Issue #1257 

Future event:

Screenshot 2026-01-17 at 01 09 46

Past event:

Screenshot 2026-01-17 at 00 40 53


ChaelCodes:
Thank you!! This looks great! 🙌

PR #1346, opened 2026-01-18
Create Venue Generator

ChaelCodes:
# Description

This is the first pass at a VenueGenerator.
It geolocates using an address and a name to fill out all values.
It will include or exclude locations, hotels, rooms, etc with a flag for --hotel or --no-hotel.
Currently, all values for additional sections are placeholders.

# Geolocation Issues

I discovered a few issues with using nominatim in development while working on this PR. Mainly that several values from the RubyLLM tool are not available, and that the default api key was preventing it from being used.

# Content Update

Ruby for Good 2024 was my testing buddy for this one, so they get a venue too.

# Resources
- https://github.com/alexreisner/geocoder?tab=readme-ov-file#custom-result-handling
- https://docs.ruby-lang.org/en/4.0/Struct.html
- https://github.com/rails/rails/blob/main/railties/lib/rails/generators/erb/scaffold/scaffold_generator.rb
- https://garrettdimon.com/journal/posts/erb-whitespace-and-indentation-in-rails-generator-templates

### DIFF
--- a/data/ruby-for-good/ruby-for-good-2024/event.yml
+++ b/data/ruby-for-good/ruby-for-good-2024/event.yml
@@ -12,5 +12,5 @@ banner_background: "#FFFFFF"
 featured_color: "#008080"
 featured_background: "#FFFFFF"
 coordinates:
-  latitude: 38.9071923
-  longitude: -77.0368707
+  latitude: 39.5075257
+  longitude: -77.7904127
--- a/data/ruby-for-good/ruby-for-good-2024/venue.yml
+++ b/data/ruby-for-good/ruby-for-good-2024/venue.yml
@@ -1,0 +1,21 @@
+# docs/ADDING_VENUES.md
+# Delete unnecessary keys
+---
+name: "Shepherd's Spring Camp & Retreat Center"
+url: "https://www.shepherdsspring.org/facilities"
+
+address:
+  street: "16869 Taylors Landing Road"
+  city: "Woburn Estates"
+  region: "Maryland"
+  postal_code: "21782"
+  country: "United States"
+  country_code: "us"
+  display: "16869 Taylors Landing Road, Sharpsburg, MD 21782"
+
+coordinates:
+  latitude: 39.5075257
+  longitude: -77.7904127
+
+maps:
+  google: "https://maps.app.goo.gl/nwzNbZZgeVgo6Zww5"


config/initializers/geocoder.rb
```diff
@@ -41,7 +41,7 @@ def normalize_key(key)
 
 google_api_key = ENV["GEOLOCATE_API_KEY"]
 
-if google_api_key.present?
+if google_api_key.present? && google_api_key != "YOUR_GEOLOCATE_API"
```
marcoroth, line 44: Where is this default coming from? Do you mean the `.env.sample`?

data/tiny-ruby-conf/tiny-ruby-conf-2026/venue.yml
marcoroth, line 1: I don't think the venue for tiny ruby #{conf} 2026 is announced yet

app/tools/venue_create_tool.rb
```diff
@@ -33,13 +33,13 @@ def execute(event_query:, venue_name:, address: nil, source_url: nil)
     venue_data = {
       "name" => venue_name,
       "address" => {
-        "street" => geo.street_address,
+        "street" => geo.address,
```
marcoroth: Should we use `geo.street_address` as the default and fallback to `geo.address` when it's not available?

```suggestion
        "street" => geo.street_address || geo.address,
```

Looking at the `geo.address` example output from nominatim you had for tiny ruby conf: it looks quite verbose and confusing 🤔

data/tiny-ruby-conf/tiny-ruby-conf-2026/venue.yml
ChaelCodes: You're right! I got confused looking at the 2025 event. 

https://helsinkiruby.fi/tinyruby/

config/initializers/geocoder.rb
ChaelCodes: Yes! I copied the .env.sample over to .env which means that nominatim wasn't used.

app/tools/venue_create_tool.rb
ChaelCodes: I was okay using a messy address because I didn't think we actually displayed it anywhere.

Looking at Geocoder - https://github.com/alexreisner/geocoder/blob/8143fce1820539cdaf6344c3400edf5690ec1611/lib/geocoder/results/google.rb#L86-L88

We can probably recreate this logic here. I submitted a PR to add it to Geocoder's nominatim result too.

Issue #1285, opened 2026-01-11
CFP should have closing-times


marcoroth:
Given we live in a world with timezones, it would be great to have the CFP deadlines recorded as a timestamp, rather than just a date.

PR #1348, opened 2026-01-18
Change CFP dates to datetime

viniciusalonso:
Issue #1285


marcoroth (commented):
Hey @viniciusalonso, sorry for the lack of detail in #1285.

Currently we already have the `CFP` model, which has an `open_date` and `close_date` of type `date`. Ideally, we could update the `CFP` model directly.

We probably also need some logic to account for the `precision`, so we know if we should show the time or not next to the CFP dates.

viniciusalonso:
@marcoroth could you create a to-do list for this issue? I feel like I’m getting a bit lost. 😂

marcoroth:
It looks like you added the `open_date` and `close_date` columns back to the `Event` model with a `datetime` type. The event itself doesn't store the CFP information anymore, we moved it to the `CFP` model. So all these changes you did should be applied to the `CFP` model instead.

viniciusalonso:
I'm sorry for this mistake. I will fix it and update this PR.

viniciusalonso:
@marcoroth fixed and ready for review.

viniciusalonso:
Hey @marcoroth does it make sense yet?

PR #1353, opened 2026-01-20
fix aliases admin pages

adrienpoly:
The aliases pages in the admin currently crashes

The initial problem was aliasable_link not being defined for Cities.

the second problems comes from the `query_constraints` in City model

@marcoroth is it safe to remove it ? of does it breaks something else ?

### DIFF
--- a/test/models/event_test.rb
+++ b/test/models/event_test.rb
@@ -241,5 +241,6 @@ class EventTest < ActiveSupport::TestCase
     event = events(:rails_world_2023)
 
     assert_equal "Amsterdam", event.city_record.name
+    assert event.city_record.events.include?(event)
   end
 end
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -556,7 +556,17 @@ class UserTest < ActiveSupport::TestCase
   end
 
   test "belongs to city record" do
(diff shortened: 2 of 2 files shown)


app/models/city.rb
```diff
@@ -29,8 +29,6 @@ class City < ApplicationRecord
 
   configure_slug(attribute: :name, auto_suffix_on_collision: true)
 
-  query_constraints :name, :country_code, :state_code
```
marcoroth, line 32: I think we still need it, as this is what allows `Event` and `User` to reference `cities` by the (`city`, `country_code`, `state_code`) columns instead of a `city_id` foreign key.

adrienpoly: I could see the City has_many :users association but I don't think user.city or event.city is defined anywhere

for the has_many the contrains are already specified in the association definition so I think we are good with that

I could define a belongs_to association like that (and removed the constrains)

```rb
belongs_to :city_record, optional: true, foreign_key: [:city, :country_code, :state_code], primary_key: [:name, :country_code, :state_code]
```

adrienpoly:
I have updated the PR I think the associations are working fine in both directions now and it is safe to remove the constrains but maybe I am missing something

marcoroth (approved):

PR #1365, opened 2026-01-23
Add wnb rb meetup sponsors

dalmaboros:
Adding annual sponsors of WNB.rb.

Note: We're working on getting OmbuLabs up on the WNB.rb website, but as an organizer of WNB.rb I can confirm they are an active sponsor.

### DIFF
--- a/data/wnb-rb/wnb-rb-meetup/sponsors.yml
+++ b/data/wnb-rb/wnb-rb-meetup/sponsors.yml
@@ -7,7 +7,7 @@
         - name: "Gusto"
           website: "https://gusto.com/"
           slug: "gusto"
-          logo_url: "https://rubyonrails.org/assets/images/logo-gusto.svg"
+          logo_url: "https://www.wnb-rb.dev/packs/static/images/Gusto-logo-fbfed097feb08a784df3.png"
     - name: "Opal"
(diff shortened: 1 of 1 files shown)


data/wnb-rb/wnb-rb-meetup/sponsors.yml
```diff
@@ -0,0 +1,22 @@
+---
+- tiers:
+    - name: "Emerald"
+      description: "Annual sponsorships of $5000 per year"
+      level: 1
+      sponsors:
+        - name: "Gusto"
+          website: "https://gusto.com/"
+          slug: "gusto"
+          logo_url: "https://rubyonrails.org/assets/images/logo-gusto.svg"
```
ChaelCodes: We select the logo url on the admin page from an array of event logos. 

If you don't have one or the sponsor already has one, you don't need to include a logo_url.

No need to find one from another event. :)

dalmaboros: Ah, gotcha. I updated the `logo_url`s to point to wnb-rb.dev, and removed one `logo_url` that isn't up on the website.

ChaelCodes: Perfect!

PR #1384, opened 2026-01-31
Internationalization

adrienpoly:
This PR is a first attempt to make our site localizable. 
as a proof of concepts it extracts hard coded text strings from the navbar and home page and allows to switch from english to Japanese

The current implementation will infer the local from the url /ja/talks to display the Japanese version.

It is one of the recommended [approach from the Rails guide](https://guides.rubyonrails.org/i18n.html#setting-the-locale-from-url-params), The Rails guide also don't recommend using [Session based local storage](https://guides.rubyonrails.org/i18n.html#storing-the-locale-from-the-session-or-cookies)

This approach currently breaks of lot of  path helpers in the app as the locale route scope breaks positional arguments for route helpers

talk_path(talk_id) should now be writing talk_path(id: talk_id)

not sure we want to got that route now this is an exploration PR at this point 

## Preview
CleanShot 2026-01-31 at 17 06 12@2x

CleanShot 2026-01-31 at 17 06 25@2x


config/locales/ja.yml
```diff
@@ -0,0 +1,51 @@
...
+    cfps: "CFPs"
+    favorite_rubyists: "お気に入りのRubyist"
+    contribute: "貢献する"
+    sign_in: "ログイン"
+    sign_out: "ログアウト"
+    admin: "管理"
+    my_profile: "マイプロフィール"
+    my_profile_title: "My Ruby Eventsプロフィール"
+    settings: "設定"
+    suggestions: "提案"
```
okuramasafumi:
```suggestion
    suggestions: "改善提案"
```

config/locales/ja.yml
```diff
@@ -0,0 +1,51 @@
...
+    leaderboard: "リーダーボード"
+    countries: "国"
+    for_you: "あなたへ"
```
okuramasafumi:
```suggestion
    for_you: "あなたへのおすすめ"
```

okuramasafumi (commented):
I found some room for improvements!

PR #1386, opened 2026-02-01
Refactor home page and navigation to use I18n for localization

adrienpoly:
This is mostly an extraction from #1384 and a first step towards internationalization. 
Hard coded strings from the home page are extracted to i18n keys


ChaelCodes:
I think we should try using lazy lookups for the views - https://guides.rubyonrails.org/i18n.html#lazy-lookup

adrienpoly:
> I think we should try using lazy lookups for the views - https://guides.rubyonrails.org/i18n.html#lazy-lookup

I’ve had mixed experiences with lazy lookups in the past. While they do make things simpler to write initially, they can make view refactoring more complicated later on.

Whenever you extract parts of a view into a partial, you have to remember to also update the YAML file so the translation keys are nested under the partial’s scope. I’ve been caught by this a few times before.

That said, I’m completely fine with using this approach, we’ll just need to be a bit more careful whenever we refactor views.

ChaelCodes:
Great point. Maybe we need some CI for untranslated keys? I think there's a config that raises if a key doesn't have a translation.

adrienpoly:
> Great point. Maybe we need some CI for untranslated keys? I think there's a config that raises if a key doesn't have a translation.

yes I can set this up

PR #1400, opened 2026-02-05
fix slug for fastruby

matthewmayer:
incorrect data caused fastruby logo/website to show up at https://www.rubyevents.org/organizations/fastly - they are two different companies

### DIFF
--- a/data/railsconf/railsconf-2025/sponsors.yml
+++ b/data/railsconf/railsconf-2025/sponsors.yml
@@ -141,7 +141,7 @@
           slug: "gitlab"
           logo_url: "https://railsconf.org/images/uploads/gitlab-logo-100.svg"
 
-        - name: "Fastly"
+        - name: "FastRuby.io"
           website: "https://www.fastruby.io/"
-          slug: "fastly"
+          slug: "fastruby-io"
           logo_url: "https://railsconf.org/images/uploads/fast-ruby-io-rails-maintenance-services.png"


ChaelCodes:
Thank you so much for pointing this out! While I was reviewing this task, I noticed that actually all of the fastly sponsorships appear to be fastruby.io.

I checked Ruby SAAS and WNB.rb. Unfortunately, the RailsConf Atlanta one doesn't load on web archive, but I think based on the other sponsorships, we can infer that sponsorship was fastruby.io as well. 

You don't have to, but would you mind updating the yaml for the other ones as well while you've got it open? Then I'll update fastly and the event sponsorships in the admin page.

matthewmayer:
I think its correct in the YAML already for those other 3, its just the one bad slug that is causing it to appear under Fastly not FastRuby

AI explanation:
```
They’re getting mixed up because one of the sponsor YAML entries incorrectly labels FastRuby.io as “Fastly,” and the import logic merges organizations by name/slug.

Specifically, in railsconf-2025 sponsors data you have:

sponsors.yml
Lines 139-147
https://www.fastruby.io/
        - name: "Fastly"          website: "https://www.fastruby.io/"          slug: "fastly"          logo_url: "https://railsconf.org/images/uploads/fast-ruby-io-rails-maintenance-services.png"

Your Static::Event.import_sponsors! logic:
- First tries to match an Organization by domain (parsed from website),
- Then falls back to Organization.find_by_name_or_alias(name) and find_by_slug_or_alias(slug),
- Then find_or_initialize_by(name: sponsor["name"]).

Because this sponsor row says name = "Fastly" but website/logo = FastRuby.io, the importer can end up:

Matching an existing Fastly organization and overwriting its website/domain with fastruby.io, or

Creating/merging data so that the same Organization record represents a mix of Fastly + FastRuby.io.
```

matthewmayer:
this is still showing incorrectly at https://www.rubyevents.org/organizations/fastly - does some cache need to be cleared?

PR #1401, opened 2026-02-05
RubyConfTH: add full schedules for 2019, 2022, 2023; remove 2020

matthewmayer:
Added some missing talks from 2019 without videos
Note the 2020 conf was heavily delayed until 2022, dont think it needs its own entry.

### DIFF
--- a/data/rubyconfth/rubyconf-th-2020/event.yml
+++ b/data/rubyconfth/rubyconf-th-2020/event.yml
@@ -1,0 +1,14 @@
+---
+id: "rubyconf-th-2020"
+title: "RubyConf TH 2020 (Cancelled)"
+description: ""
+location: "Bangkok, Thailand"
+kind: "conference"
+start_date: "2020-10-16"
+end_date: "2020-10-17"
+website: "https://rubyconfth.com/"
+status: "cancelled"
+twitter: "rubyconfth"
+coordinates:
+  latitude: 13.7563309
+  longitude: 100.5017651


data/rubyconfth/rubyconf-th-2020/event.yml
```diff
@@ -1,14 +0,0 @@
----
-id: "rubyconf-th-2020"
-title: "RubyConf TH 2020 (Cancelled)"
```
ChaelCodes, line 3: Normally we want to collect as much information about what happened and when as possible. But we also want to respect organizer wishes about how to represent their event. So I'm going to defer the question of cancelled events to @marcoroth or @adrienpoly and then document our official opinion in the ADDING_EVENTS contributing guide. I think we probably need an FAQ for things like this.

data/rubyconfth/rubyconfth-2019/videos.yml
```diff
@@ -58,7 +55,16 @@
   video_provider: "youtube"
   video_id: "GcxlNOewz18"
 
-# Missing Video: How to collaborate and keep healthy culture in a full-remote company Jarka Košanová
+- id: "jarka-kosanova-rubyconf-th-2019"
+  title: "How to Collaborate and Keep Healthy Culture in a Full-Remote Company"
+  speakers:
+    - Jarka Košanová
```
ChaelCodes, line 61: Normally I'd merge it and make a note to go back, but we're waiting on Marco and Adrien anyway, and you're becoming a bit of an expert in RubyEvents.

It's incredibly helpful to add an entry in speakers.yml with the GitHub profile for any new speakers, because then when they sign up for the site, they'll see their talks associated with their profile, and it looks good on the event when everyone has a face.

No pressure, we often go through and fill in blank speakers or dedupe from ones that have been set in the admin, but if you have time to add missing speakers to speakers.yml, that'd be awesome!

matthewmayer: is there a quick way to see which speakers for a series don't have an entry in speakers.yml?

data/rubyconfth/rubyconf-th-2020/event.yml
matthewmayer: Understood. I think it's not helped by the fact that when you view a series like

https://www.rubyevents.org/events/series/rubyconfth

there's nothing really to show that the 2020 event was cancelled - the (Cancelled) is truncated on my screen size, it's not greyed-out etc.

image

data/rubyconfth/rubyconfth-2019/videos.yml
matthewmayer: i have pushed the missing speakers

data/rubyconfth/rubyconf-th-2020/event.yml
matthewmayer: i have reverted removing this for now to keep this PR simpler, will start a seperate issue for discussing this

matthewmayer: ref: https://github.com/rubyevents/rubyevents/issues/1407

ChaelCodes:
Thank you so much for updating the details of your event and for your patience. This looks fantastic! 🙌

ChaelCodes (approved):

PR #1403, opened 2026-02-06
Fix video stops when clicking bookmark button

jmarsh-gaia:
Bookmark button was causing the video player to stop when clicked because it triggered a full page morph. Now uses turbo_stream to update just the button element.


jmarsh-gaia:
Recreating from justinallenmarsh account

PR #1404, opened 2026-02-06
Fix video stops when clicking bookmark button

justinallenmarsh:
Bookmark button was causing the video player to stop when clicked because it triggered a full page morph. Now uses turbo_stream to update just the button element.


ChaelCodes:
The bookmarking worked great, video kept playing! But now I have to click twice to unbookmark and that still pauses the video.

PR #1410, opened 2026-02-12
Fix hidden grid items on tablet viewports

yukideluxe:
Hello! 👋🏻 

I was looking at my profile (😳) on my iPad when I noticed the following...

CleanShot 2026-02-12 at 12 57 24@2x

ONE TALK IS MISSING but it turns out that it is just hidden by the css depending on the viewport size 😬 

I can see why in the home page we do not want to show extra items if the viewport does not fit them but, in the pages that show all items, we should never hide them. Am I missing something!?

I asked Claude to lookup the commit and PR descriptions to see if there's a good reason for using those styles in some pages and it feels to me it's just a copy & paste mistake 🤔 

## BEFORE

CleanShot 2026-02-12 at 13 00 42@2x

## AFTER

CleanShot 2026-02-12 at 13 00 16@2x

### DIFF
--- a/app/views/countries/state.html.erb
+++ b/app/views/countries/state.html.erb
@@ -48,7 +48,7 @@
             <h2 class="text-primary shrink-0"><%= kind.humanize.pluralize %></h2>
           </div>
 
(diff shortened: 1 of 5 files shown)


ChaelCodes (approved):
Thank you so much for catching and correcting this!

yukideluxe:
yay! thanks for the quick merge @ChaelCodes! <3

Issue #1408, opened 2026-02-07
Turkish developers and organizations not in the map


enderahmetyurt:
<img width="938" height="341" alt="Image" src="https://github.com/user-attachments/assets/04396e93-9395-48ac-b233-33637d874d87" />
<img width="292" height="167" alt="Image" src="https://github.com/user-attachments/assets/6e5cbf34-06a0-4c95-a656-381e441d48bf" />

Hello,

Even my location is Turkey, my profile doesn't show up in the map 🤔 and also Ruby Türkiye organization doesn't show up.

marcoroth:
Interesting 🤔 It looks like it might have trouble finding the country from the URL?

https://www.rubyevents.org/countries/tuerkiye this redirects back to https://www.rubyevents.org/countries

I wonder if it has to do with the `ü` in the name 🤔 

However, cities like Istanbul work fine: https://www.rubyevents.org/cities/istanbul

enderahmetyurt:
Thank you @marcoroth I can take a look if you mind

marcoroth:
@enderahmetyurt please feel free to take a look! 🙏🏼

marcoroth:
Resolved via #1411

PR #1411, opened 2026-02-12
Strip diacritics in Country slug instead of transliterating

enderahmetyurt:
Use Unicode NFKD normalization to strip diacritics (ü → u) instead of relying on Rails' default transliteration (ü → ue) for country slug generation.

This produces cleaner URLs like `/countries/turkiye` instead of `/countries/tuerkiye` for countries like Türkiye.

The `find` method is updated for consistency, ensuring URL lookups match the generated slugs.

Related: #1408

**Before**
Screenshot 2026-02-12 at 16 50 34

**After**
Screenshot 2026-02-12 at 17 20 08

### DIFF
--- a/app/models/country.rb
+++ b/app/models/country.rb
@@ -29,7 +29,7 @@ class Country
   end
 
   def slug
-    name.unicode_normalize(:nfkd).gsub(/\p{Mn}/, "").parameterize
+    name.unicode_normalize(:nfkd).parameterize
   end
 
   def path
@@ -149,7 +149,7 @@ class Country
   class << self
     def find(term)
       term = term.to_s.tr("-", " ")
-      term_slug = term.unicode_normalize(:nfkd).gsub(/\p{Mn}/, "").parameterize
+      term_slug = term.parameterize
 
       return nil if term.blank?
       return nil if term.downcase.in?(%w[online earth unknown])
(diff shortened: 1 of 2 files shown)


app/models/country.rb
```diff
@@ -29,7 +29,7 @@ def name
   end
 
   def slug
-    name.parameterize
+    name.unicode_normalize(:nfkd).gsub(/\p{Mn}/, "").parameterize
```
marcoroth:
```suggestion
    name.unicode_normalize(:nfkd).parameterize
```

app/models/country.rb
```diff
@@ -149,7 +149,7 @@ def uk_nation?
   class << self
     def find(term)
       term = term.to_s.tr("-", " ")
-      term_slug = term.parameterize
+      term_slug = term.unicode_normalize(:nfkd).gsub(/\p{Mn}/, "").parameterize
```
marcoroth:
```suggestion
      term_slug = term.unicode_normalize(:nfkd).parameterize
```

test/models/country_test.rb
```diff
@@ -146,6 +146,19 @@ class CountryTest < ActiveSupport::TestCase
     assert_equal "united-states", country.slug
   end
 
+  test "slug strips diacritics instead of transliterating them" do
+    country = Country.find_by(country_code: "TR")
+
+    assert_equal "turkiye", country.slug
```
marcoroth:
```suggestion
    assert_equal "tuerkiye", country.slug
```

test/models/country_test.rb
```diff
@@ -146,6 +146,19 @@ class CountryTest < ActiveSupport::TestCase
...
+  end
+
+  test "find returns country by slug with diacritics stripped" do
+    country = Country.find("turkiye")
```
marcoroth, line 156:
```suggestion
    country = Country.find("tuerkiye")
```

enderahmetyurt:
Thank you @marcoroth I kept what we have. I added a fallback.

PR #1345, opened 2026-01-17
add Baruco 2012, 2013 and 2014 videos

yukideluxe:
Hello 👋🏻 

I had a moment of nostalgia and realized these conference videos were not added yet! After a bit of digging on archive.org, a lot of AI plus manual intervention to make sure Claude was not messing up... I managed to get all the info needed to fill it up 😊  

In honor of the amazing graphics from the Baruco conferences, I’ll try to track down the proper images and complete the 2015 edition soon 🙏🏻

PS: I want Baruco back!!! 😭  
PS2: I *could* have used existing tools to retrieve the videos, but I’m currently experimenting a lot with Claude Code, so… here we are 🫠

PR #1412, opened 2026-02-12
Baruco conferences assets

yukideluxe:
Hello!

As promised https://github.com/rubyevents/rubyevents/pull/1345, here are the assets for the Baruco conferences! I did my best to pull the images from google and youtube videos. 

It was a bit of a hassle but I think this is a good effort! 🙏🏻


marcoroth (approved):
Awesome, thank you so much @yukideluxe! I know it's quite a bit of effort to add assets for events, so it's much appreciated! 🙏🏼

But I also think it makes such a huge difference in the way you perceive and see the event on RubyEvents.org.

Thank you!

yukideluxe:
> But I also think it makes such a huge difference in the way you perceive and see the event on RubyEvents.org.

It does!!!! I wish I could have found them with better quality, though 🥲

PR #1415, opened 2026-02-13
Better handling of cancelled events and events with no year in YAML

matthewmayer:
* Cancelled events are now shown with opacity:0.5 so they look different on the event series page
* Cancelled events are not included in the EVENT_COUNT or TALK_COUNT in the series summary "FOOCONF is a yearly conference and hosted EVENT_COUNT events between START and END. We have currently indexed TALK_COUNT FOOCONF talks."
* The START and END dates are fixed so that conferences that have a start_date but no date are included in the series summary

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -5,6 +5,7 @@
 @import "components/button.css";
 @import "components/diff.css";
 @import "components/dropdown.css";
+@import "components/event.css";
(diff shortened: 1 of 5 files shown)


matthewmayer:
https://www.rubyevents.org/events/series/rubyconf-india is a good example of an existing series where the dates are obviously wrong because not all events have years which this should fix.

PR #1416, opened 2026-02-13
updated assets for RubyConfTH

matthewmayer:


ChaelCodes (approved):
This is interesting! I don't think we have any cards on the site like these.

ChaelCodes:
Screenshot 2026-02-13 at 7 56 29 PM
Screenshot 2026-02-13 at 7 56 02 PM

ChaelCodes:
I'm going to merge this, but the idea behind the assets is really to give people a feel for your branding. Right now, the year is displayed twice, and it will be everywhere the card displays. In a way, you've cut your branding space in half.

matthewmayer:
Our branding actually leans pretty heavily on the year digits which use a stylized Thai font

matthewmayer:
For some reason on production the 2023 and 2019 cards have updated but not 2026 or 2022

image

marcoroth:
I just cleared the cache, I think it should be fine now! 🙌🏼

PR #1421, opened 2026-02-14
Case insensitive speaker sort

matthewmayer:
Similar to 6660e35977048ebb2c657a3acd7cc726c1e37871 this makes speaker sort case insensitive so the lowercase names (mostly pseudonyms/usernames) don't get pushed to the end

image


marcoroth (approved):
Nice, thanks @matthewmayer! 🙏🏼

PR #1425, opened 2026-02-16
Blue Ridge Ruby talks and TalkGenerator

ChaelCodes:
# Description

Add new talk generator. Each call adds one new unpublished, scheduled talk to videos.yml.

Add Blue Ridge Ruby 2026 Speakers using the generator.

### DIFF
--- a/lib/generators/talk/talk_generator.rb
+++ b/lib/generators/talk/talk_generator.rb
@@ -19,7 +19,7 @@ class TalkGenerator < Generators::EventBase
 
   def add_talk_to_file
     videos_file_path = File.join("data", options[:event_series], options[:event], "videos.yml")
-    template "videos.yml.tt", videos_file_path unless File.exist? videos_file_path
+    template "videos.yml.tt", videos_file_path unless File.exist?(destination_path(videos_file_path))
 
     if File.read(destination_path(videos_file_path)).match?(/- id: "#{@talk_id}"/)
       match_one_talk = /\n- id: "#{@talk_id}"[\s\S]*video_id: "#{@talk_id}"/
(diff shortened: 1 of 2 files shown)


lib/generators/talk/talk_generator.rb
```diff
@@ -0,0 +1,31 @@
...
+  class_option :title, type: :string, default: "TODO", desc: "Title of the talk"
+  class_option :speaker, type: :array, default: ["TODO"], desc: "Speaker name"
+  class_option :description, type: :string, default: "TODO", desc: "Description of the talk"
+  class_option :date, type: :string, desc: "Date of the talk (YYYY-MM-DD)", required: false
+  class_option :kind, type: :string, enum: Talk.kinds.keys, default: "talk", desc: "Type of talk (e.g., 'keynote', 'lightning')"
+
+  def initialize_values
+    event = Static::Event.find_by_slug options[:event]
+    @date = options[:date] || (event&.start_date || Date.today).iso8601
+    @talk_id = "#{options[:speaker].first.parameterize}-#{options[:kind]}-#{options[:event]}"
```
marcoroth, line 17: I wonder if we should leave out the `kind` here but instead add all speakers to match the way we previously manually defined the `id`s? Was there a reason you did it this way?

ChaelCodes: There are some people who do lightning talks and talks at a single event, so I thought adding kind might help distinguish those. Definitely game to add 2 speakers to the id. Those panel ids get long though.

ChaelCodes: To be honest, I thought we put keynote in the ids sometimes, but in review, I see that I was wrong.

ChaelCodes: I haven't run the seeds yet, so we could update all the ids.

marcoroth: Yeah, I guess that makes sense. For panels we usually just did `panel-[event-slug]`. I think it would be fine to manually update the outliers. 

Maybe we can just leave it out when `kind == "talk"`?

PR #1439, opened 2026-02-20
include aliases when ignoring old events from rubyconferences.org

matthewmayer:
Currently some events are listed under “Add missing events” on http://rubyevents.org/contributions even though they are already on the site. This is due to name mismatches between the old name on rubyconferences.org and the current name on the site. Thus:

1. Also include aliases in the list of names to reject against
2. Add some aliases for some conferences to remove them from the pending list

### DIFF
--- a/data/baruco/barcelona-ruby-conf-2015/event.yml
+++ b/data/baruco/barcelona-ruby-conf-2015/event.yml
@@ -17,3 +17,4 @@ featured_background: "#000000"
 featured_color: "#FFFFFF"
 aliases:
   - Barcelona Ruby Conf 2015
+  - Full Stack Fest 2015


data/baruco/barcelona-ruby-conf-2015/event.yml
```diff
@@ -15,3 +15,5 @@ coordinates:
 banner_background: "#000000"
 featured_background: "#000000"
 featured_color: "#FFFFFF"
+aliases:
+  - Barcelona Ruby Conf 2015
```
ChaelCodes, line 19:
```suggestion
  - Barcelona Ruby Conf 2015
  - Full Stack Fest 2015
```

data/ruby-retreat-nz/ruby-retreat-nz-2025/event.yml
```diff
@@ -13,3 +13,5 @@ featured_color: "#000000"
 coordinates:
   latitude: -36.61280051278539
   longitude: 174.81887537984366
+aliases:
+  - Ruby Retreat 2025
```
ChaelCodes, line 17: I went to this one!

PR #1440, opened 2026-02-21
Move incorrectly-attributed @tomstuart talks to @rentalcustard

tomstuart:
As noted in #307, there are two Rubyists ([@tomstuart](https://www.rubyevents.org/profiles/tomstuart) — which is me — and [@rentalcustard](https://www.rubyevents.org/profiles/tom-stuart-mortice) aka [`@mortice`](https://twitter.com/mortice)) called Tom Stuart.

Because of the name collision, two of @rentalcustard’s talks have been incorrectly attributed to me. This change updates both talks’ metadata to correctly attribute them to @rentalcustard instead.

### DIFF
--- a/data/ruby-on-ice/ruby-on-ice-2018/videos.yml
+++ b/data/ruby-on-ice/ruby-on-ice-2018/videos.yml
@@ -187,7 +187,7 @@
 - title: "Fortunately, maths!"
   raw_title: "Fortunately, maths!"
   speakers:
-    - Tom Stuart
+    - Tom Stuart @mortice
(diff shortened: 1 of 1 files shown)


tomstuart:
A test is [failing](https://github.com/rubyevents/rubyevents/actions/runs/22261705695/job/64400942844?pr=1440#step:9:21), but the same test [failed](https://github.com/rubyevents/rubyevents/actions/runs/22219852359/job/64272269763#step:9:21) yesterday, so I don’t think it‘s related.

Issue #1437, opened 2026-02-20
Event involvements for an organization should be sorted by date


matthewmayer:
e.g. https://www.rubyevents.org/organizations/ruby-australia

Currently they display in an arbitrary order

<img width="1309" height="651" alt="Image" src="https://github.com/user-attachments/assets/cdfa7e2e-99d3-4a59-bd4e-2cbdb133ab76" />

sudeeptarlekar:
Same for https://www.rubyevents.org/organizations/bangkok-rb

currently they are ordered by `position`, I would like to take this up, but before creating PR, I am not able to understand why they are ordered by position?

Are events still need to be ordered by `position` or can we change it to order by event date?

ChaelCodes:
Ordering by event date instead of position is fine, but please be aware that [meetups](https://github.com/rubyevents/rubyevents/blob/main/data%2Fwnb-rb%2Fwnb-rb-meetup%2Fevent.yml) don't have event_dates, but they can have sponsors and involvements.

PR #1445, opened 2026-02-22
Order event involvements for organization by date

sudeeptarlekar:
Currently on organization show page, involvements are ordered by position, and when position for event involvements is not correct events are ordered in arbitrary order.
This commit fixes the ordering by either start_date from events column, or created_at, when start_date is empty.

fixes: #1437

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -17,7 +17,9 @@ class OrganizationsController < ApplicationController
 
     @countries_with_events = @events.grouped_by_country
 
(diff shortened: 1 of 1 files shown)


sudeeptarlekar:
Can I get reviews @ChaelCodes @matthewmayer 👀

ChaelCodes:
> Can I get reviews @ChaelCodes @matthewmayer 👀 

I'd be happy to review this in the morning or during lunch tomorrow.

I like to boot up the app and test the changes as part of my review process. Unfortunately, that causes some delays in review.

Issue #1436, opened 2026-02-20
Within each series in event archive, should order by date not alphabetically


matthewmayer:
Current sort order goes wrong when not all events in a series are named "FOOCONFERENCE YEAR", examples attached

<img width="486" height="328" alt="Image" src="https://github.com/user-attachments/assets/d1097ebf-8890-4501-9541-c57fe439b1f6" />
<img width="493" height="306" alt="Image" src="https://github.com/user-attachments/assets/beb43bf6-7158-4f56-a588-62b2d5541462" />
<img width="494" height="223" alt="Image" src="https://github.com/user-attachments/assets/0397dff4-e23f-4d56-b0d0-334a7bbb3f79" />
<img width="619" height="213" alt="Image" src="https://github.com/user-attachments/assets/1a0b8a71-c880-4854-bd6b-8ba650eee0b0" />

PR #1447, opened 2026-02-23
Order events archive by name and start date

viniciusalonso:
Issue #1436

Screenshot 2026-02-22 at 22 15 00

### DIFF
--- a/app/controllers/events/archive_controller.rb
+++ b/app/controllers/events/archive_controller.rb
@@ -2,7 +2,7 @@ class Events::ArchiveController < ApplicationController
   skip_before_action :authenticate_user!, only: %i[index]
 
   def index
-    @events = Event.canonical.includes(:series).order("events.start_date ASC, events.name ASC")
+    @events = Event.canonical.includes(:series).order("series.name ASC, events.start_date ASC")
     @events = @events.where("lower(events.name) LIKE ?", "#{params[:letter].downcase}%") if params[:letter].present?
     @events = @events.ft_search(params[:s]) if params[:s].present?
   end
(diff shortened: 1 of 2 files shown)


ChaelCodes:
Hey, looks like the sort order is still having issues for those conferences where the event names don't match, and events with numbers in the name. :/

Screenshot 2026-02-23 at 2 41 25 PM
Screenshot 2026-02-23 at 2 41 54 PM
Screenshot 2026-02-23 at 2 42 09 PM

viniciusalonso:
Hey @ChaelCodes

Screenshot 2026-02-23 at 18 32 46


Screenshot 2026-02-23 at 18 33 39

Screenshot 2026-02-23 at 18 34 10

ChaelCodes:
Let me do a little thinking here, because now they're sorted with all meetups first, and then by date.

Screenshot 2026-02-24 at 1 48 04 PM

Maybe we should be sorting within the view? After we do the grouping? I haven't looked at the code for that.

app/controllers/events/archive_controller.rb
```diff
@@ -2,7 +2,7 @@ class Events::ArchiveController < ApplicationController
   skip_before_action :authenticate_user!, only: %i[index]
 
   def index
-    @events = Event.canonical.includes(:series).order("events.name ASC")
+    @events = Event.canonical.includes(:series).order("events.start_date ASC, events.name ASC")
```
ChaelCodes: We include the series. Can we sort by the event series name, and then the event date?

viniciusalonso: Done :)

Issue #1438, opened 2026-02-20
Event involvements should be counted in the count for organizations


matthewmayer:
https://www.rubyevents.org/organizations/ruby-australia
e.g. Ruby Australia has 4 involvements but no sponsorships, would expect it to show 4 on https://www.rubyevents.org/organizations but it currently shows 0

<img width="310" height="42" alt="Image" src="https://github.com/user-attachments/assets/d7e142bf-a2c3-4055-8c22-5d5f2a1a51ad" />

ChaelCodes:
I think we need two bubbles. To me, organizing an event, and sponsoring an event feel different, and it feels like a different amount of effort.

Ruby Central both sponsors and organizes events, so they feel like a good example of this - https://www.rubyevents.org/organizations/ruby-central

matthewmayer:
Maybe something like this could work:

<img width="374" height="43" alt="Image" src="https://github.com/user-attachments/assets/41baedb8-6e3a-4a60-bde0-e8af6d902b23" />

This would also make organizers stand out in the list slightly (as they'd have a differently styled count)

marcoroth:
I like that @matthewmayer! Otherwise, I think summing up the two counts on the index would be fine too

PR #1449, opened 2026-02-23
Count event involvements for organizations

sudeeptarlekar:
On current organizations list page, only count for sponsorships are shown, this commit also adds the count for the organized events on index page.

Fixes: #1438

### DIFF
--- a/app/views/organizations/_organization.html.erb
+++ b/app/views/organizations/_organization.html.erb
@@ -26,9 +26,5 @@
     <span class="line-clamp-1"><%= organization.name %></span>
   </div>
 
-  <%= ui_badge(kind: :secondary, outline: true, size: :lg, class: "min-w-10") do %>
-    <span class="text-error"><%= organization.event_involvements.size %></span>
-    <span class="mx-1 opacity-30">|</span>
-    <span><%= organization.events.size %></span>
-  <% end %>
+  <%= ui_badge(organization.event_involvements.size + organization.events.size, kind: :secondary, outline: true, size: :lg, class: "min-w-10") %>
 <% end %>


sudeeptarlekar:
@ChaelCodes can I get review pls 👀 ?

Thanks in advance :)

ChaelCodes:
> @ChaelCodes can I get review pls 👀 ?
> 
> Thanks in advance :)

Hi! I promise I've got you on my list to review. This week, I've really only had lunch to do code reviews, and I go from oldest to newest, prioritizing content reviews.

ChaelCodes:
I really like the faded line between the numbers, and I see what you were going for with the different colors indicating involvement vs sponsorship. Without a key, and without a matching color, I don't think it communicates which value the number is.

Screenshot 2026-02-25 at 1 02 04 PM

I think we have two options:
@marcoroth and @matthewmayer both like summing the involvements and sponsors, so if you go that route, I'll just be outvoted.

OR I think I'd like a tooltip indicating if it's an involvement or sponsorship, and I want both values to be the same text color, since we don't have a key or color associated with sponsorships/involvements.

Sorry that the issue wasn't clear about the exact approach we'd like. There's some debate. I'll try to keep a closer eye on issues in the future.

(In the future, if you can include a screenshot, I can get design feedback to you faster, before I pull it up on my laptop and test it.)

matthewmayer:
If we go with this approach I think we should omit the 0s 

So like

`3 | 2`

If you have both types but just a plain

`3` 

If you only have sponsorships.

sudeeptarlekar:
I am still confused about which approach are we trying to take here? Do we agree on @matthewmayer suggestion?

IMO, I would also sum up involvements + sponsorships on index and on show there is detailed overview...

ChaelCodes:
We'll just do the sum. :)

matthewmayer:
A nice addition to this might be if you go to a organization page with no Supported Events, but has Involvements e.g.

https://www.rubyevents.org/organizations/ruby-australia

Then we default to showing the Involvements tab?

ChaelCodes:
I made that a separate issue, so we can merge this one as-is.

sudeeptarlekar:
Fixed count on index page...
Screenshot 2026-02-26 at 09 17 50


Screenshot 2026-02-26 at 09 18 46

PR #1451, opened 2026-02-26
Show playlists for events with no videos page

matthewmayer:
On /contributions page, in the "Events with no videos" tab, if an event has a playlist in the YML, link to it. This allows prioritising events for which videos are available but not extracted yet.


ChaelCodes:
Are there any events with playlists and no videos? I don't see any.

matthewmayer:
Lots eg https://github.com/rubyevents/rubyevents/blob/main/data/rubyconf-india/rubyconf-india-2018/event.yml

PR #1452, opened 2026-02-26
Rubyconf India and DeccanRubyConf assets

matthewmayer:
- Add various RubyConf India assets for 2010-2022
- Add various DeccanRubyConf asssets from 2014-2018


marcoroth:
Thank you @matthewmayer! 🙏🏼

matthewmayer:
@ChaelCodes https://www.rubyevents.org/events/series/rubyconf-india seems to have cached card images?

ChaelCodes:
@matthewmayer Yeah, I'm looking and it seems like the new banner is reflected, but not the card. 

I will say that 2022's filename is card2.webp.

We've had this problem before, but I've never been the one to fix it, so I'm not sure how to. @marcoroth ?

Maybe we need a cache-busting rake task/GH action?

marcoroth:
Just ran this in production: 
```ruby
EventSeries.find_by(name: "RubyConf India").events.touch_all
```

Maybe we should make the card image path part of the cache key, or remove the cache altogether. I'm not sure the cards are worth caching anyway.

matthewmayer:
Yeah, this only seems to affect cards not avatars or banners.

PR #1462, opened 2026-02-28
sort speakers.yml

matthewmayer:
Some manually added speakers were not in correct alphabetical order

This PR continues to use a naive Unicode case-insensitive sort and doesnt attempt to otherwise prettify the YAML.


matthewmayer:
I wrote a script but as it's in JS not Ruby I thought I'd not include it. Maybe I can port to Ruby later

matthewmayer:
> I wrote a script but as it's in JS not Ruby I thought I'd not include it. Maybe I can port to Ruby later

ported to Ruby and added in https://github.com/rubyevents/rubyevents/pull/1484

PR #1463, opened 2026-02-28
combine windy city rails series

matthewmayer:
Windy City Rails was split into two series with different slugs

https://www.rubyevents.org/events/series/windycityrails
https://www.rubyevents.org/events/series/windy-city-rails


ChaelCodes (approved):
Thank you! :raised_hands: I'll need to remove the old one in production once this merges.

PR #1479, opened 2026-03-05
Fix past scope to include events with missing dates

### LINKED ISSUE #1476: Attended events counter if off
On this page by example 

https://www.rubyevents.org/profiles/inicolas69/events

We see a counter of 1 attended event but no events are listed in the tab
Image

It might be related to the way we count meet ups?

enderahmetyurt:
Events with NULL start_date or end_date were excluded from the `past` scope, causing attended events to appear in the counter but not render in the profile events tab.

## Description
Fix `Event.past` scope to include events with `NULL` start/end dates. Previously, attended events with missing dates were counted in the profile tab counter but never rendered in the list, showing "No participated events yet" despite the counter showing 1+. This affected meetups which often lack date data.

## Screenshots
Screenshot 2026-03-05 at 18 22 41

## Testing Steps
1. Find or create a user with an `EventParticipation` for an event where `start_date` and `end_date` are both `NULL`
2. Visit `/profiles/<slug>/events`
3. Confirm the attended event now appears under "Past Events"

Automated regression tests added in `test/models/event_test.rb` covering:
- Event with `NULL` end_date + past start_date → included in past
- Event with `NULL` start_date and `NULL` end_date → included in past
- Event with future start_date → not included in past

closes #1476

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -127,7 +127,7 @@ class Event < ApplicationRecord
       )
       .distinct
   }
-  scope :past, -> { where("end_date <= ? OR (end_date IS NULL AND (start_date <= ? OR start_date IS NULL))", Date.today, Date.today).order(Arel.sql("COALESCE(end_date, start_date) DESC")) }
+  scope :past, -> { where(end_date: ..Date.today).order(end_date: :desc) }
   scope :upcoming, -> { where(start_date: Date.today..).order(start_date: :asc) }
 
   def upcoming?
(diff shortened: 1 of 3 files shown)


ChaelCodes:
Screenshot 2026-03-06 at 9 18 14 PM

Both my attempts to add a meetup and then view the profile were met with this error. I tried with WNB.rb and Ruby Türkiye.

We have a sort_date method on the event specifically to address cases like these.

https://github.com/rubyevents/rubyevents/blob/6a2c00eb00fc953c117a953b13545aad982642c6/app/models/event.rb#L316-L318

PR #1484, opened 2026-03-06
Add a scripts/sort_speakers.rb to sort speakers.yml alphabetically

matthewmayer:


ChaelCodes:
I'm still thinking about this one. If we were going to add a script to sort them, I'd want to add it to bin/lint and CI. But this looks like it rewrites the file every time.

matthewmayer:
> I'm still thinking about this one. If we were going to add a script to sort them, I'd want to add it to bin/lint and CI. But this looks like it rewrites the file every time.

It does rewrite the whole file every time, but as its just basic string manipulation its very fast, only half a second or so.

I found it useful to be able to just append any new speakers at the end, and then run a script to move them into the correct place.

Issue #1380, opened 2026-01-28
Put Meetups back on the Map!


ChaelCodes:
# Description

In the process of working on the new geolocation and maps pages, meetups fell off the map.
If you're looking at events close to a location, meetups are one of the best ways to connect with that local community.
Let's put meetups back on the Overview tab of a City and State.
For countries, seeing all meetups might be overwhelming for the larger countries, so meetups should be in a separate tab.

<img width="1416" height="835" alt="Image" src="https://github.com/user-attachments/assets/65f78456-440b-4838-98ad-0fb892fd79a1" />

[Example state page](https://www.rubyevents.org/states/us/california)
[Example city page](https://www.rubyevents.org/cities/san-francisco)

> Small PRs merge faster, so if you want to break this up into City, and then State, and then maybe a tab for meetups on the country, feel free!

ChaelCodes:
Closed by #1492!

PR #1492, opened 2026-03-08
Add Meetups tab to countries, states and cities

matthewmayer:
## Description

This adds a Meetups tab on the countries, states and cities pages using the existing events/card partial

Since meetups are modelled as a single event without start_date/end_date, to avoid a date of “unknown” appearing for meetups on the events card, it now returns nil for formatted_dates

## Screenshots

Screenshot 2026-03-08 at 17 29 07

Screenshot 2026-03-08 at 16 45 43

## Testing Steps

Check city/country/state/continent pages

Note meetups will need to be properly geocoded to appear.

## References

First step for #1380

This doesnt yet add Meetups directly on Overview page or Map as the PR was already quite large.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -238,7 +238,7 @@ class Event < ApplicationRecord
   end
 
   def formatted_dates
-    return nil if meetup?
+    return "Recurring" if meetup?
     case date_precision
     when "year"
       start_date.strftime("%Y")


app/models/event.rb
```diff
@@ -238,6 +238,7 @@ def today?
   end
 
   def formatted_dates
+    return nil if meetup?
```
ChaelCodes: We sometimes sort by this field, so it's very important that we return a value. What about "ongoing" for meetups?

app/views/shared/_location_meetups.html.erb
```diff
@@ -0,0 +1,37 @@
...
+
+    <%= ui_badge(meetups.size, kind: :secondary, outline: true) %>
+  </div>
+
+  <% if meetups.any? %>
+    <div class="grid grid-cols-1 gap-4 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4">
+      <%= render partial: "events/card", collection: meetups, as: :event %>
+    </div>
+  <% else %>
+    <p class="text-gray-500 py-8 text-center">No meetups in <%= location.name %> yet.</p>
```
ChaelCodes:
```suggestion
    <p class="text-gray-500 py-8 text-center">No known meetups in <%= location.name %> yet.</p>
```

app/models/event.rb
matthewmayer: how about "Recurring"? "Ongoing" implies its still current, which might not be the case for lapsed meetups?

PR #1505, opened 2026-03-09
add RubyAndRails 2010 (former RubyEnRails)

matthewmayer:
There was a final edition of RubyEnRails called RubyAndRails 2010:

https://rubyonrails.org/2010/9/21/rubyandrails-2010


marcoroth (approved):
Thanks for adding this!

@matthewmayer I'm curious, where/how did you find this?

matthewmayer:
Fairly randomly. I clicked on one of the earlier conferences because I wondered what country it was based (I thought maybe "en" was England). Then I googled the name and found the news article mentioned above referring to the old and new names.

Issue #1490, opened 2026-03-08
Eileen Uchitelle name change


hschne:
Just a quick one. 

I came across http://eileencodes.com/posts/where-identity-begins/

Eileen changed her last name to Alayce. It's an interesting scenario, not sure if we can already handle speaker name changes.

matthewmayer:
Yes its quite straightforward, just need to edit https://github.com/matthewmayer/rubyevents/blob/2c5ea928e36ab1cd54512bb654eeba74e40efd4d/data/speakers.yml#L3989 - make the preferred name the "name" and move any old names to aliases section

ChaelCodes:
It could be improved, but we have docs for name changes - https://github.com/rubyevents/rubyevents/blob/main/docs/FIXING_PROFILE_NAMES.md

marcoroth:
The best way to preserve the records for historical reasons would be to keep the old name as-is in existing files and update the `aliases` key and map the names to a single profile that way, as Matt mentioned.

But it's also fair that people don't want to have their old name still being referenced like that, in which case we can also replace all old names. So it's mostly preference and up to the person and how they want to have it handled.

marcoroth:
Resolved via #1508

PR #1508, opened 2026-03-09
Change name for Eileen Alayce

hschne:
See #1490. I'm a bit confused about where eileencodes in the profile comes from, couldn't find that in the speakers.yml, hence I didn't add it to the aliases. Is that something someone changed the DB? :thinking:


marcoroth (approved):
Thank you @hschne! 

If we have the GitHub handle, we prefer that over the `slug` for the profile URL. But both the `slug` and `github_handle` can we be used to get to the profile.

PR #1509, opened 2026-03-09
Fix polymorphic path in maps controller

sudeeptarlekar:
If User's GitHub handle is different than RubyEvent's profile slug then system redirects user to profile with GitHub slug using polymorphic path generation. But as it is missing index keyword it results in NoMethodError.
This commit fixes the path and thus NoMethodError

Resolves: #1358

## Testing Steps
User model has two fields `slug` and `github_handle`

1. Create User on system with slug different than github_handle
2. Load `/profiles/:slug/maps`
3. Observe that user is getting redirected to `/profiles/:github_handle/maps without NoMethodError

### DIFF
--- a/test/controllers/profiles/map_controller_test.rb
+++ b/test/controllers/profiles/map_controller_test.rb
@@ -1,0 +1,16 @@
+require "test_helper"
+
+class Profiles::MapControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    @user = users(:yaroslav)
+    sign_in_as(@user)
+  end
+
+  test "should redirect user if GitHub handle is different than profile slug" do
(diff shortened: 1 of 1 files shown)


sudeeptarlekar:
@ChaelCodes need your review 👀

marcoroth:
Hey @sudeeptarlekar, were you able to reproduce this issue locally?

sudeeptarlekar:
Yes, I added steps to reproduce. Let me know if they work for you, or if any update needed to them...

Screenshot 2026-03-09 at 23 18 53

marcoroth:
I see. Would you mind adding a test for this so we can avoid a regression in the future. Thank you!

PR #1512, opened 2026-03-10
remove bios from speakers.yml where github is available

matthewmayer:
Since bios can be dynamically pulled from Github, this reduces the weight of speakers.yml

### DIFF
--- a/data/speakers.yml
+++ b/data/speakers.yml
@@ -1257,8 +1257,6 @@
   slug: "arjun-sharma"
 - name: "Arjun Singh"
   github: "arjun810"
-    changes. He's been working with Rails since 2006, and previously founded Gradescope, which is a (Rails-based) AI grading platform used by millions of professors and students
-    worldwide, and was acquired by Turnitin in 2018."
   linkedin: "arjun-singh-629216105"
   slug: "arjun-singh"
 - name: "Arkadiusz Turlewicz"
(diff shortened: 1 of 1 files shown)


ChaelCodes:
Here's the full context of this.

If I understand correctly, we pull bios from GitHub for the profile. Then if there's a bio set in speakers.yml, we override that bio with the one in speakers.yml (most of which were sourced from GitHub). Unfortunately, if the user changes their bio in the app, we overwrite it.

Also multi-line bios are a common linting pain point for speakers.yml which is already massive.

For all of these reasons, @marcoroth and I have talked very briefly about removing bios from speakers.yml.

Because of this, I've been encouraging people to leave out bios when adding speakers.yml and when I add speakers myself.

@matthewmayer took the initiative based on that feedback to build this PR.

This should be good, and I'm going to double-check those assumptions. But I want to get feedback from others before deleting 1K+ lines from speakers.yml, and give others time to object and understand the changes.

marcoroth:
We have been using the bios in the `speakers.yml` to seed, but as soon as a speaker has a bio set I think we shouldn't be overwriting it anymore.

We actually had this at some point, but I guess we might have last that behavior in the migration to the `Static::*` models. So maybe this would be a great opportunity to bring that behavior back.

https://github.com/rubyevents/rubyevents/blob/2525cb8762829b48f1b9a1d0cf68c71c10e3a707/app/models/static/speaker.rb#L35

I think this should change this to something like `user.bio = bio if bio.present? && user.bio.blank?`. Does that make sense?

Issue #1523, opened 2026-03-12
Sort "always-open" meetup CFPs below conference CFPs


matthewmayer:
https://www.rubyevents.org/cfp

Sort conferences with the most imminently closing CFPs at the top. Always-open meetup CFPs can go at the bottom.

PR #1525, opened 2026-03-12
Sort conferences on CFPs at the top

sudeeptarlekar:
## Description
Currently all the CFPs at the moment are ordered based on closing date and they are displayed on the CFP index page. This commit shows conferences at the top on CFP list page ordered by closing date and then meetup after the conference.

## Screenshots
Screenshot 2026-03-12 at 20 42 00

## References
Fixes: #1523

### DIFF
--- a/app/controllers/cfp_controller.rb
+++ b/app/controllers/cfp_controller.rb
@@ -9,6 +9,6 @@ class CFPController < ApplicationController
       cfps = cfps.joins(:event).where(events: {kind: params[:kind]})
     end
 
-    @events = cfps.map(&:event)
+    @events = cfps.map(&:event).group_by(&:kind).values.flatten
   end
 end


app/controllers/cfp_controller.rb
```diff
@@ -9,6 +9,6 @@ def index
       cfps = cfps.joins(:event).where(events: {kind: params[:kind]})
     end
 
-    @events = cfps.map(&:event)
+    @events = cfps.map(&:event).group_by(&:kind).values.flatten
```
ChaelCodes, line 12: I have genuinely never seen this technique before and I quite like it it. We could also have sorted by event.kind, cfp.close_date, right?

sudeeptarlekar: yes, sort by `event.kind` can be done, sorting by close date is not needed as we already ordered them by SQL.
Now I am thinking which one approach is more efficient 🤔 

May be I will create another PR after running benchmarks.

Thanks for pointing it out :)

PR #1526, opened 2026-03-12
Add informations about upcoming and past meetups

viniciusalonso:
## Description
Add informations about upcoming and past meetups

## Screenshots
Screenshot 2026-03-12 at 20 03 14

## References

- closes #


ChaelCodes:
The weird thing about meetup events is that they don't have dates. One event represents every meetup.

There's a branch to refactor that, but it's not done yet.

viniciusalonso:
Interesting, does it make sense this upcoming meetups count?

ChaelCodes:
I don't think it does. I think the missing date is why all meetups are showing up in the past. You could look at videos to get upcoming and past events, but I'm not sure if there are any meetups that input upcoming events into RubyEvents.

This is more of a current data issue than a usability concern though.

@hschne - I'd be curious to hear your perspective on this one. You're really active on RubyEvents and Vienna.rb only shares past events.

hschne:
I'm not 100% sure if I understand correctly, but I do put in Vienna.rb meetups as soon as they are scheduled - so they would qualify as upcoming, and I would love for those to shop up somewhere. 

The fact that they don't have a date IMO is the data issue you described and should (?) be solved by the new data structure (one event per meetup) I think?

matthewmayer:
Agreed. If meet-ups with upcoming editions in their videos.yml show more prominently on the site that's an incentive for meet-ups to add the data for future events.

viniciusalonso:
I've changed the scopes `past_meetups` and `upcoming_meetups` to filter by `published_at` in talks. 

cc @matthewmayer @ChaelCodes

matthewmayer:
Shouldn't it be date rather than published_at?

published_at is the date the YouTube video was published if applicable which might be after the event

viniciusalonso:
@matthewmayer good catch! Fixed

matthewmayer:
i noticed nearly every meetup in the preview section incorrectly shows "Takes place today"

image

Can this easily be fixed together with this change?

viniciusalonso:
It looks easy to fix, but I'd prefer to do it in a new PR. This one is taking longer than I expected 😂

I will send a new PR with this fix tonight.


Screenshot 2026-03-16 at 13 08 56
Screenshot 2026-03-16 at 13 09 10

app/models/event.rb
```diff
@@ -130,6 +130,14 @@ class Event < ApplicationRecord
   scope :past, -> { where(end_date: ..Date.today).order(end_date: :desc) }
   scope :upcoming, -> { where(start_date: Date.today..).order(start_date: :asc) }
 
+  scope :past_meetups, -> {
+    joins(:talks).where(kind: :meetup, talks: {date: ..Date.today}).distinct
+  }
+
+  scope :upcoming_meetups, -> {
+    joins(:talks).where(kind: :meetup, talks: {date: nil}).distinct
```
ChaelCodes: Talks always have a date. https://github.com/rubyevents/rubyevents/blob/3f3402b234a50563e1738af8445e496cd6c21b5c/app/schemas/video_schema.rb#L64

I think we want a range for this one too.

ChaelCodes:
> I'm not 100% sure if I understand correctly, but I do put in Vienna.rb meetups as soon as they are scheduled - so they would qualify as upcoming, and I would love for those to shop up somewhere.
> 
> The fact that they don't have a date IMO is the data issue you described and should (?) be solved by the new data structure (one event per meetup) I think?

Sorry for the confusion. I was looking for some clarity around the use case. You're one of the most active meetup organizers I know on the site and when I checked Vienna.rb only had past ones.

My thinking was that if we only have past meetups at the moment, it might be better to defer this until after the meetups refactor, because they'll all be past anyway.

But it sounds like there is interest in upcoming meetups, and I see that Ruby Tuesday has one on March 31st, so we'll have at least one.

app/models/event.rb
viniciusalonso: Fixed

viniciusalonso:
> i noticed nearly every meetup in the preview section incorrectly shows "Takes place today"
> 
> image
> Can this easily be fixed together with this change?

Hey @matthewmayer, I created a PR for this one https://github.com/rubyevents/rubyevents/pull/1547

ChaelCodes:
Oh no! There's a merge conflict.

viniciusalonso:
> Oh no! There's a merge conflict.

@ChaelCodes fixed :)

ChaelCodes:
@viniciusalonso Thank you!!

PR #1528, opened 2026-03-13
Add slider on user settings page for distance

### LINKED ISSUE #1331: Configure how far "near #{location}" is
Right now we calculate "near" with a fixed value  of "250 km", I feel like people should decide how far "near" is to them. Depending on the country far/near can be interpreted very differently.

Maybe we can make this configurable using a slider on `/settings`?

sudeeptarlekar:
## Description
Currently "near" value for user is fixed to 250KM, this commit adds the slider on settings page with which user can configure how far is near to them. Minimum value is 250KM and maximum is 20,000KM.

## Screenshots
Screenshot 2026-03-13 at 17 40 09

## References
Closes #1331

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -140,7 +140,7 @@ class User < ApplicationRecord
   validates :email, format: {with: URI::MailTo::EMAIL_REGEXP}, allow_blank: true
   validates :github_handle, presence: true, uniqueness: true, allow_blank: true
   validates :canonical, exclusion: {in: ->(user) { [user] }, message: "can't be itself"}
-  validates :distance, comparison: {less_than_or_equal_to: 20_000, greater_than_or_equal_to: 250}
+  validates :distance, comparison: {less_than_or_equal_to: 20_000, greater_than_or_equal_to: 0}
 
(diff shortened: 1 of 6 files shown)


sudeeptarlekar:
@marcoroth can I get reviews 👀 ?

Thanks in advance...

app/models/user.rb
```diff
@@ -139,6 +140,7 @@ class User < ApplicationRecord
   validates :email, format: {with: URI::MailTo::EMAIL_REGEXP}, allow_blank: true
   validates :github_handle, presence: true, uniqueness: true, allow_blank: true
   validates :canonical, exclusion: {in: ->(user) { [user] }, message: "can't be itself"}
+  validates :distance, comparison: {less_than_or_equal_to: 20_000, greater_than_or_equal_to: 250}
```
marcoroth: I think we can allow down to 0, if we let people customize it then we might as well let them do it all the way

sudeeptarlekar:
Screenshot 2026-04-02 at 17 38 39

marcoroth (approved):
Thank you @sudeeptarlekar! 🙏🏼

PR #1511, opened 2026-03-09
Create an initial version of meetups page

viniciusalonso:
## Screenshots
Screenshot 2026-03-09 at 19 42 43

## References
#1502

PR #1529, opened 2026-03-15
Add browse by country section to meetups page

matthewmayer:
* adds browse by country (filtered to only countries with meetups, and linking to their meetups tab)
* links to continents, countries and cities pages

extension of https://github.com/rubyevents/rubyevents/pull/1511

CleanShot 2026-06-08 at 18 29 35@2x


ChaelCodes:
Can you split these? The contribution banner is an easy merge, but I'm not sure about worldwide meetups being on this page. Worldwide meetups feels like a separate page or view. In addition, as we add more meetups, this will move further and further down the page, making it inaccessible.

Also, I haven't pulled this up to test it (and won't have time before my lunch break is over) but this should be tested on mobile as well.

matthewmayer:
banner split to #1545

marcoroth:
Folded this into the switcher at the top right, thank you! 🙌🏼

PR #1538, opened 2026-03-16
hex colors must include #

### LINKED ISSUE #1537: invalid hex colors cause contrast problems for some event previews
Example:

Image

matthewmayer:
fixes #1537

### DIFF
--- a/docs/ADDING_VISUAL_ASSETS.md
+++ b/docs/ADDING_VISUAL_ASSETS.md
@@ -61,7 +61,7 @@ featured_color: "#FFFFFF"         # Text color for featured display
 ```
 
 These colors support:
-- Hex colors (e.g., `#FF0000`)
+- Hex colors (e.g., `#FF0000`) - don't forget the `#`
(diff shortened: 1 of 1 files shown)


ChaelCodes:
I can't tell you why - but this appears to be intentional.

https://github.com/rubyevents/rubyevents/blob/501652afbf3c72d7eb6f39c3407b25079d8bce56/lib/tasks/event_assets.rake#L329-L332

Either that or there's a data issue happening in the generator that needs to be resolved.

marcoroth:
I think the rake task is actually wrong

ChaelCodes:
If that's the case, we should fix the generate event assets rake task to remove deleting the #.

https://github.com/rubyevents/rubyevents/blob/501652afbf3c72d7eb6f39c3407b25079d8bce56/lib/tasks/event_assets.rake#L329-L332

ChaelCodes:
We can capture this with a pattern too - https://github.com/danielfriis/ruby_llm-schema?tab=readme-ov-file#strings

PR #1543, opened 2026-03-17
fix: Leave # in Generate Assets rake task

### LINKED ISSUE #1539: Leave # in Generate Assets rake task
We should fix the generate event assets rake task to remove deleting the #.

https://github.com/rubyevents/rubyevents/blob/501652afbf3c72d7eb6f39c3407b25079d8bce56/lib/tasks/event_assets.rake#L329-L332

_Originally posted by @ChaelCodes in https://github.com/rubyevents/rubyevents/issues/1538#issuecomment-4071735478_

JiwaniZakir:
Fixes #1539

The `EventAssetGenerator` class in `lib/tasks/event_assets.rake` was stripping the `#` prefix from hex color values when writing `featured_color` to event YAML files. The root cause was an explicit `.delete("#")` call on `text_color` in both the update and append branches of the color-writing logic. Removes the `.delete("#")` calls so `text_color` is written as-is, preserving the `#` prefix that other parts of the system expect. Verified by reviewing the before/after output of the `featured_color` field in generated event YAML files.

### DIFF
--- a/lib/tasks/event_assets.rake
+++ b/lib/tasks/event_assets.rake
@@ -327,9 +327,9 @@ class EventAssetGenerator
     end
 
     if content.match?(/^featured_color:/)
-      content.gsub!(/^featured_color:.*$/, "featured_color: \"#{text_color.delete("#")}\"")
+      content.gsub!(/^featured_color:.*$/, "featured_color: \"#{text_color}\"")
     else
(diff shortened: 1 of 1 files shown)


ChaelCodes:
Thank you for fixing this!! Nice work! 🙌

PR #1564, opened 2026-03-22
JSON Schema Validation in VS Code

ChaelCodes:
## Description

We can get in-editor schema validation using the [yaml-language-server](https://github.com/redhat-developer/yaml-language-server) if we store our schemas as json. This allows for autocomplete and validation of our keys without running bin/lint or validate.rake.

This is currently configured for [VS Code](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml), but supports [multiple editors.](https://github.com/redhat-developer/yaml-language-server?tab=readme-ov-file#clients)
 
- [x] New rake task to generate all schemas (manually run upon changing a schema)
- [x] Configuration for installing the Yaml VS Code extension
- [x] Ensure the schemas stay up-to-date

## Screenshots

image

## References

- closes #

### DIFF
--- a/lib/schemas/featured_city_schema.json
+++ b/lib/schemas/featured_city_schema.json
@@ -1,0 +1,48 @@
+{
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "Full city name"
+    },
+    "slug": {
+      "type": "string",
+      "description": "URL-friendly slug for the city"
+    },
+    "state_code": {
+      "type": [
+        "string",
+        "null"
+      ],
+      "description": "State or province code"
+    },
+    "country_code": {
+      "type": "string",
+      "description": "ISO 3166-1 alpha-2 country code"
+    },
+    "latitude": {
+      "type": "number",
+      "description": "Geographic latitude"
+    },
+    "longitude": {
+      "type": "number",
+      "description": "Geographic longitude"
+    },
+    "aliases": {
+      "type": "array",
+      "description": "Alternative names or abbreviations",
+      "items": {
+        "type": "string"
+      }
+    }
+  },
+  "required": [
+    "name",
+    "slug",
+    "country_code",
+    "latitude",
+    "longitude"
+  ],
+  "additionalProperties": false,
+  "strict": true
+}
--- a/lib/schemas/involvement_schema.json
+++ b/lib/schemas/involvement_schema.json
@@ -1,0 +1,28 @@
+{
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "Role or involvement type (e.g., 'Organizer', 'Program Committee member')"
+    },
+    "users": {
+      "type": "array",
+      "description": "Person names involved in this role",
+      "items": {
+        "type": "string"
+      }
+    },
+    "organisations": {
+      "type": "array",
+      "description": "Organization names involved in this role",
+      "items": {
+        "type": "string"
+      }
+    }
+  },
+  "required": [
+    "name"
+  ],
+  "additionalProperties": false,
+  "strict": true
+}
--- a/lib/schemas/speaker_schema.json
+++ b/lib/schemas/speaker_schema.json
@@ -9,13 +9,13 @@
       "type": "string",
       "description": "URL-friendly slug for the speaker"
     },
-    "twitter": {
-      "type": "string",
-      "description": "Twitter/X handle (without @)"
-    },
     "github": {
       "type": "string",
       "description": "GitHub username"
+    },
+    "twitter": {
+      "type": "string",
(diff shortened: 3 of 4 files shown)


rakelib/schema.rake
```diff
@@ -0,0 +1,17 @@
...
+  desc "Export all schemas as JSON Schemas to lib/schemas"
+  task export: :environment do
+    require "fileutils"
+
+    schema_files = Dir.glob(Rails.root.join("app/schemas/*_schema.rb"))
+    schema_files.each do |file|
+      base = File.basename(file, ".rb")
+      schema_class = base.camelize.constantize
+      json_schema_path = Rails.root.join("lib/schemas/#{base}.json")
+      File.write(json_schema_path, schema_class.new.to_json_schema[:schema].to_json)
```
matthewmayer: could this be pretty-printed so the resulting JSON is easier to read?

marcoroth: `to_json` on the schema instance directly produces a nice pretty printed schema:
```suggestion
      File.write(json_schema_path, schema_class.new.to_json)
```

lib/schemas/address_schema.json
```diff
@@ -0,0 +1 @@
+{"type":"object","properties":{"street":{"type":"string","description":"Street address"},"city":{"type":"string","description":"City name"},"region":{"type":"string","description":"State/Province/Region"},"postal_code":{"type":"string","description":"Postal/ZIP code"},"country":{"type":"string","description":"Country name"},"country_code":{"type":"string","description":"ISO country code (e.g., 'US', 'CA', 'JP')"},"display":{"type":"string","description":"Full formatted address for display"}},"required":["street","city","country","country_code","display"],"additionalProperties":false,"strict":true}
```
marcoroth: I'm wondering if we should add `‎lib/schemas/*.json` to the `.gitignore` so that we don't have to rebuild them when we update the schemas. 

We could invoke the rake task to generate them in our regular `seed` command. Or is there a strong reason to have these artifacts committed?

ChaelCodes: I'd rather keep them committed and keep the seeds small and fast.
I could write a GH Action that only runs if there's a change to schemas/* in a pull request to check if the schemas match.

rakelib/schema.rake
ChaelCodes: It does. I'll have to double-check but I think I needed the [:schema] key. Maybe `schema_class.new.to_json["schema"]` would work. I'll just have to test it.

And @matthewmayer , I have it on my to-do list to get the pretty-printing in, that's a good suggestion.

lib/schemas/address_schema.json
ChaelCodes: The strong argument is just that we only need to run the rake task when there's a change, which is much rarer than running seeds, and anyone using codespaces or devcontainers gets the schemas for free.

Actually - there's a non-interactive version of codespaces that's free and doesn't give you access to the terminal, but it has extensions. They could write and validate yaml in the cloud, for free.

Unfortunately, it doesn't reflect our recommended extensions yet, I tested that.

rakelib/schema.rake
marcoroth: If you need the schema key you could use [`JSON.pretty_generate`](https://rubyapi.org/4.0/o/json#method-i-pretty_generate).

lib/schemas/address_schema.json
ChaelCodes: I wrote a test instead - curious what you think about the idea.

test/lib/schema_export_test.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  Dir.glob(SCHEMA_DIR.join("*_schema.rb")).each do |schema_file|
+    base = File.basename(schema_file, ".rb")
+    test "#{base} generates matching JSON schema" do
+      schema_class = base.camelize.constantize
+      generated = JSON.pretty_generate(schema_class.new.to_json_schema[:schema])
+      json_path = JSON_DIR.join("#{base}.json")
+      assert File.exist?(json_path), "Missing JSON schema for #{base}. Run `bin/rails schema:export` to generate it."
+      expected = File.read(json_path)
+      assert_equal expected, generated, "Schema for #{base} is out of date. Run `bin/rails schema:export` to update it."
```
ChaelCodes, line 16: Tests for matching schemas! Capture if the schemas have been updated and the rake task hasn't been run.

ChaelCodes:
Screenshot 2026-03-28 at 10 25 33 PM
It works!

rakelib/schema.rake
ChaelCodes: Done!

lib/schemas/address_schema.json
marcoroth: Great idea! 🙌🏼

PR #1579, opened 2026-03-27
Speakers.yml validations and updates

marcoroth:
This pull request introduces several improvements to the handling and validation of speaker data. The changes ensure better data consistency through add validation tasks.

* Added a `speakerdeck` profile URL and support for an array of `aliases` (with `name` and `slug`) to the `SpeakerSchema` and removed the`bio` field. 
* Updated the speaker import logic to remove setting the `bio` field, aligning with the schema changes.
* Added new Rake tasks to validate that all speaker slugs and GitHub handles are unique, and that all speakers referenced in `videos.yml` exist in `speakers.yml`. These checks help prevent duplicate entries and missing references.


ChaelCodes:
Question - what do we do if we can't find the GitHub?

marcoroth:
@ChaelCodes we have been using `github: ""`, the key just needs to present but doesn't actually need a value. I felt like that way it's super obvious which speakers are still missing a GitHub handle.

ChaelCodes:
@marcoroth Got it!! I got confused and thought this changed that.

PR #1581, opened 2026-03-28
Add Philly.rb

kcdragon:
## Description

Add Philly.rb meetup along with some specific events

## References

- closes #


ChaelCodes (approved):
This looks great! Thank you for adding it! 🙌 

Image

It might be a good idea to [add some assets](https://github.com/rubyevents/rubyevents/blob/main/docs/ADDING_VISUAL_ASSETS.md) though.

PR #1591, opened 2026-03-29
Add schedules and videos for historical Kansai RubyKaigi

ydah:
## Summary
- add `schedule.yml` and `videos.yml` for Kansai RubyKaigi 02, 03, 04, 05, 06, and 2017
- structure the schedules to match the `fukuoka-rubyistkaigi-05` pattern, with talk slots represented by `videos.yml`
- validate the new schedule and video data against the existing schemas

## Notes
- videos are marked as `not_recorded` because these are historical events and no recordings were linked in the sources
- Kansai RubyKaigi 01 is not included because the full timetable could not be reconstructed from the available archives


ChaelCodes:
Hey! It's been a few weeks with no comment from us, so I wanted to let you know what's going on behind the scenes.

The Japanese community is important to us, and we want to support titles in Japanese, but we're having trouble with the url slugs generated by the talk titles. The slugs filter out kanji, so we're ending up with slugs that just match the handful of english words in the title, or for ones that are entirely in japanese - just the event name.

We tried #1605 but it's not transliterating the title in a way that is comprehensible.

We're still trying to figure out a solution to merge this and we want these talks represented, but this is the challenge we're facing.

PR #1592, opened 2026-03-29
Sort past events in profiles by most recent

### LINKED ISSUE #1589: Past events on profile are not chronological
Ideally, we could sort them by most recent first

Image

Example: https://www.rubyevents.org/profiles/tenderlove/events

viniciusalonso:
## Description
Sort past events in profiles by most recent

## Screenshots
Screenshot 2026-03-29 at 19 43 45

## Testing Steps
* Access http://localhost:3000/profiles/tenderlove/events

## References

- closes #1589

### DIFF
--- a/app/views/profiles/_events.html.erb
+++ b/app/views/profiles/_events.html.erb
@@ -1,7 +1,7 @@
 <%# Event Participations v1 %>
 
 <% future_events = events.upcoming.sort_by(&:sort_date).reverse %>
-<% past_events = (events.to_a - future_events.sort_by(&:sort_date).reverse).sort_by(&:sort_date).reverse %>
+<% past_events = (events.to_a - future_events).sort_by(&:sort_date).reverse %>
 
 <% cache [user, events, Current.user] do %>
   <div>


app/views/profiles/_events.html.erb
```diff
@@ -1,7 +1,7 @@
 <%# Event Participations v1 %>
 
 <% future_events = events.upcoming.sort_by(&:sort_date).reverse %>
-<% past_events = events.to_a - future_events.sort_by(&:sort_date).reverse %>
+<% past_events = (events.to_a - future_events.sort_by(&:sort_date).reverse).sort_by(&:sort_date).reverse %>
```
ChaelCodes:
```suggestion
<% past_events = (events.to_a - future_events).sort_by(&:sort_date).reverse %>
```

I think you just need the parens. The order of the future_events array shouldn't matter since we're just using it to subtract.

viniciusalonso: Done, thanks for suggestion @ChaelCodes

PR #1597, opened 2026-04-01
Remove download_sponsors

ChaelCodes:
## Description

`download_sponsors` no longer succeeds because ActiveGenie has changed their API. We're interested in moving in another direction, so I'm removing it entirely.

In addition, the spec has been flaky lately.

Updated minor gem versions as part of the removal.

## References

- closes #


marcoroth (approved):

Gemfile.lock
```diff
@@ -21,7 +21,7 @@ GIT
 
 GIT
   remote: https://github.com/rails/rails.git
-  revision: debeab22780607f22b976d8608d26fc2b997e256
+  revision: 8041099b68f0a44fe9051a35fc15b91882273404
```
marcoroth: Looks like this might be causing some issues with `require_dependency` 🙈

PR #1598, opened 2026-04-01
Grepfruit will display sponsor TODOs

ChaelCodes:
## Description

We're currently updating events with sponsors that are yet to happen, which means sponsors may be added in the future. Unfortunately, events with sponsors files don't display under Contributing for Missing Sponsors.

Grepfruit looks for todos and displays them at the event and event series level. This allows us to surface files that may still be missing sponsors, even if the sponsor file is already generated.

## Screenshots

Screenshot 2026-04-01 at 12 46 32 PM

## Testing Steps

1. Generate a new sponsors file using `bin/rails g sponsors`
2. Observe that there is now a TODO that must be manually reviewed.

## References

- closes #


ChaelCodes:
We could hide this behind and `if Day.today < event.date` but I want to refactor sponsors to add or update them one at a time like videos and support Sponsors adding just their sponsorship.

PR #1614, opened 2026-04-07
Show next mutual event

marcoroth:
image


ChaelCodes (approved):
I really like this! I wonder if it's an N+1 though?

ChaelCodes:
We talked about how this might look a bit better if there was more than one item. I gave it a lot of thought - "<count> unwatched talks" might be the right choice. If we add a new past talk, it'll bump the number. People can mark them all watched if they care to keep up with new talks.

PR #889, opened 2025-08-10
Add London Ruby User Group (LRUG)

h-lame:
The videos.yml file is generated by the lrug site and hosted at
https://lrug.org/rubyevents-video-playlist.yml so we've just copied it into the
rubyevents repo for this PR. Currently fixed at 2020 onwards as while we have
meetings on the site going back to 2006, all the links to recordings now go to
a private vimeo account due to the long-time venue / host going into
administration.

We would be interested in exploring something that allows us to ping ruby 
events when a change is published to our video playlist yml file to automatically
update / enter events.

We have sponsor data, but haven't added it, no have we added CFP or Schedule 
info as these seem (understandably) tailored to one-off events like conferences,
less so to meet-ups.  For example, our CFP never ends.

PR #1615, opened 2026-04-07
Allow passing args for paths to enforce strings on when linting yaml

h-lame:
## Description

The `yaml/enforce_strings.mjs` script lints everything in `./data` by default, however, this is slow if we've only updated one event, or even just one yaml file for that event. This commit updates the script to treat all the arguments in `process.argv.slice(2)` (e.g. not `node` and `yaml/enforce_strings.mjs`) as paths to format.

## References

Brought up while asking about yaml listing for lrug.org data in https://github.com/rubyevents/rubyevents/pull/889#discussion_r3041833061

This discussion suggested making `bin/lint` or `yarn format:yml` take args, but looking at the contents, I don't think it makes sense: `bin/lint` lints _everything_ not just yaml files, and `format:yml` also runs prettier so it's not obvious where the args should go.  Giving the underlying `yaml/enforce_strings.mjs` script the ability to take args gives us what we need for individual usage.

### DIFF
--- a/yaml/enforce_strings.mjs
+++ b/yaml/enforce_strings.mjs
@@ -1,10 +1,37 @@
 import fs from 'fs'
 import { Formatter } from './formatter.mjs'
 
-const files = fs.readdirSync('./data', { recursive: true }).filter(file => file.endsWith('.yml'))
+const files = []
+const paths = []
+if (process.argv.length > 2) {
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
Thank you @h-lame! 🙏🏼

PR #1617, opened 2026-04-07
Replace git clone and bundle with curl in lrug script

h-lame:
## Description

The `bin/update_lrug_meetups` script clones the `lrug/lrug.org` repo, does a `bundle install`, and builds the whole site to get their version of the `rubyevents-video-playlist.yml` to drop into
`data/lrug/lrug-meetup/videos.yml`. However, this file is available on https://lrug.org so we can drastically speed this process up if we just pull it down from there instead.

I was reminded of this script in https://github.com/rubyevents/rubyevents/pull/889#discussion_r3047316678 but realised I wouldn't use it as-is because I already have [`lrug/lrug.org`](https://github.com/lrug/lrug.org) checked out.  By switching it to `curl` instead of `git clone && bundle install && bundle exec middleman build` I'm much more likely to use it, and if we do just make this run periodically on CI or whatever, this'll make it faster.

Appreciate, I may be overreaching here as this script isn't really _for me_, so no worries if this doesn't make the cut.

### DIFF
--- a/bin/update_lrug_meetups
+++ b/bin/update_lrug_meetups
@@ -7,6 +7,7 @@ if command -v mise &> /dev/null; then
 fi
 
 curl https://lrug.org/rubyevents-video-playlist.yml > data/lrug/lrug-meetup/videos.yml
-yarn format:yml
+node yaml/enforce_strings.mjs data/lrug/lrug-meetup/videos.yml
+prettier -c data/lrug/lrug-meetup/videos.yml -w
 git add data/lrug/lrug-meetup/videos.yml
 git commit -m "Update LRUG Meetups"
--- a/bin/update_lrug_meetups
+++ b/bin/update_lrug_meetups
@@ -2,22 +2,11 @@
 
 set -e  # Exit on error
 
-if [ -d "lrug.org" ]; then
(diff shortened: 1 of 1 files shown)


bin/update_lrug_meetups
```diff
@@ -2,22 +2,12 @@
...
 fi
 
-bundle install
-bundle exec middleman build
-cp public/rubyevents-video-playlist.yml ../data/lrug/lrug-meetup/videos.yml
-cd ..
-yarn format:yml
+curl https://lrug.org/rubyevents-video-playlist.yml > data/lrug/lrug-meetup/videos.yml
+node yaml/enforce_strings.mjs data/lrug/lrug-meetup/videos.yml
+prettier -c data/lrug/lrug-meetup/videos.yml -w
```
ChaelCodes, line 11: Can you add the yarn format:yml back? We prefer that to prettier. (Maybe both?)

Also, I'm happy to merge this if it makes it easier for you to update RubyEvents with lrug meetups!

ChaelCodes (approved):
Ah! I see what you're doing, you only want to update the lrug one, that's why you didn't run the whole yarn:format.

PR #1618, opened 2026-04-08
Add missing paris.rb talks

Intrepidd:


data/paris-rb/paris-rb-meetup/videos.yml
```diff
@@ -78,6 +78,1713 @@
...
+  date: "2022-02-01"
+  video_provider: "children"
+  video_id: "paris-rb-february-2022"
+  description: ""
+  talks:
+    - title: "Service Objects, la brique manquante de Rails"
+      raw_title: "Service Objects, la brique manquante de Rails"
+      event_name: "Paris.rb Meetup February 2022"
+      language: "french"
+      date: "2022-02-01"
```
ChaelCodes, line 203:
```suggestion
      date: "2022-02-01"
      speakers:
        - TODO
```

We like to use [TODO](https://www.rubyevents.org/profiles/todo) to catch missing speakers.

data/paris-rb/paris-rb-meetup/videos.yml
```diff
@@ -78,6 +78,1713 @@
...
+  date: "2022-10-04"
+  video_provider: "children"
+  video_id: "paris-rb-october-2022"
+  description: ""
+  talks:
+    - title: "Produire du logiciel dans une coopérative ouvrière"
+      raw_title: "Produire du logiciel dans une coopérative ouvrière"
+      event_name: "Paris.rb Meetup October 2022"
+      language: "french"
+      date: "2022-10-04"
```
ChaelCodes, line 394:
```suggestion
      date: "2022-10-04"
      speakers:
        - TODO
```

ChaelCodes (commented):
Hey! Can you add a TODO speaker for any talks without speakers? It helps us find and fix them.

Also, thank you for adding these! 🙌 Sorry for the delay on review.

Intrepidd:
Thanks ! I added TODO for the talks missing speakers

Intrepidd:
Random thought : we could update the lint check to ensure speakers is always set ?

ChaelCodes:
> Random thought : we could update the lint check to ensure speakers is always set ?

Unfortunately we have some old data that's bad. But there is a warning message in the seeds, which is how I caught this one. If you wanted to make the warning more obvious, but not block CI - like a [GitHub Action annotation](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-a-warning-message) - that would be really cool!

We have some in [validate.rake](https://github.com/rubyevents/rubyevents/blob/265225d32cbbb187315d91f68ae71b36aaecd83d/lib/tasks/validate.rake#L86) right now, but those are for actual CI failures. Printing a warning should still pass in CI though?

PR #1621, opened 2026-04-08
Include ... | RubyEvents.org in <title> and set missing titles

dennispaagman:
## Description
This adds `... | RubyEvents.org` to all `<title>`s and sets it for a couple a pages that didn't have it set. There might be more missing, i'll do a another check later, but this is at least already a nice improvement.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -14,7 +14,7 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
 
-    <%= display_meta_tags site: "" %>
+    <%= display_meta_tags site: "RubyEvents.org", reverse: true %>
 
     <%= vite_client_tag %>
     <%= combobox_style_tag %>
--- a/app/views/announcements/index.html.erb
+++ b/app/views/announcements/index.html.erb
@@ -8,7 +8,7 @@
 
 <div class="container py-8">
   <div class="flex items-center justify-between mb-8">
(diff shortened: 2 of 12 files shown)


app/views/layouts/application.html.erb
```diff
@@ -14,7 +14,7 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
 
-    <%= display_meta_tags site: "" %>
+    <%= display_meta_tags site: "RubyEvents.org", reverse: true %>
```
ChaelCodes, line 17: Can you explain why we should include "| RubyEvents.org" in every page title?

ChaelCodes (commented):
I love all the title additions, and I'm good with the simpler "Contribute" title, but I'm not sure why we want to add "| RubyEvents.org" to the page titles?

dennispaagman:
I think it's a best practice most sites use. I also prefer it personally. I can see a couple of reasons:
- It gives context to the page you have open, with short titles it's still a visible part of the tab. With longer names it falls off, so less benefit there.
- It also shows nicely in your browsing history
- with more browers moving to vertical tab bars (even [Chrome recently](https://blog.google/products-and-platforms/products/chrome/new-chrome-productivity-features/)), there's more space for longer titles.

But happy to get rid of it, at least having a context correct title is better than a generic one for each page.

ChaelCodes (approved):
Thanks for this! I asked, and we're good with | RubyEvents.org!

PR #1640, opened 2026-04-12
Improve memory and performance of Event Seeds

ChaelCodes:
## Description

Importing events takes ages and lots of memory. We run into issues seeding data because of this. This PR aims to reduce object allocations and time by only updating the record when there are changes. In production (and development) we're normally only updating one or two events, and rarely a larger PR might include 10. While this won't affect the initial seed data, it will help with later runs.

Each commit represents the benchmark after running `bin/rails db:seed:event_series[rubyconf]` which was slow and large enough to notice, but not as tedious as the full events seed.

## Testing Steps

```bash
bin/rails db:seed:events
```

Before:
> Allocated objects: 82901672, Real: 167.88938579300884, total: 176.907387

After:
> Allocated objects: 64046016, Real: 110.00759493600344, total: 116.541339

## References

- closes #1608

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -262,6 +262,9 @@ group :development do
   # Guard for watching file changes and auto-importing [https://github.com/guard/guard]
   gem "guard"
 
+  # Benchmark is no longer included in the standard lib - used for benchmarking
+  gem "benchmark"
+
   gem "ruby-lsp-rails", require: false
   gem "standardrb", "~> 1.0", require: false
   gem "erb_lint", require: false
--- a/app/models/static/event.rb
+++ b/app/models/static/event.rb
@@ -353,24 +353,11 @@ module Static
       return unless imported?
       return unless event.videos_file.exist?
 
-      event.videos_file.entries.each do |talk_data|
-        talk = ::Talk.find_or_initialize_by(static_id: talk_data["id"])
-        talk.update_from_yml_metadata!(event: event)
-        Search::Backend.index(talk) if index
-
-        child_talks = talk_data["talks"]
-
-        next unless child_talks
-
-        Array.wrap(child_talks).each do |child_talk_data|
-          child_talk = ::Talk.find_or_initialize_by(static_id: child_talk_data["id"])
-          child_talk.parent_talk = talk
-          child_talk.update_from_yml_metadata!(event: event)
-          Search::Backend.index(child_talk) if index
-        end
-      rescue ActiveRecord::RecordInvalid => e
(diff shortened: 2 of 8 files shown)


lib/tasks/seed.rake
```diff
@@ -25,12 +25,18 @@ namespace :db do
 
     desc "Seed one event series by passing the event series slug - db:seed:event_series[rubyconf]"
     task :event_series, [:slug] => :environment do |task, args|
-      event = Static::EventSeries.find_by_slug(args[:slug])
-      if event
-        event.import!
-      else
-        puts "Event Series with slug '#{args[:slug]}' not found."
+      benchmark = Benchmark.measure do
```
MatheusRich, line 28: do we want to leave this here or were you just using this for debugging?

ChaelCodes: I commit it in the first one, and remove it in the last commit. This PR's not done - I ran into issues with a line I thought I could remove from the sponsor setup. 

Thanks for pointing it out though!

app/models/static/event.rb
```diff
@@ -393,31 +376,31 @@ def import_sponsors!(event)
...
             end
 
-            s ||= ::Organization.find_by_name_or_alias(sponsor["name"]) || ::Organization.find_by_slug_or_alias(sponsor["slug"]&.downcase)
-            s ||= ::Organization.find_or_initialize_by(name: sponsor["name"])
+            organization ||= ::Organization.find_by_name_or_alias(sponsor["name"]) || ::Organization.find_by_slug_or_alias(sponsor["slug"]&.downcase)
+            organization ||= ::Organization.find_or_initialize_by(name: sponsor["name"])
 
-            s.update(
+            organization.assign_attributes(
               website: sponsor["website"],
```
ChaelCodes, line 402: A future area for improvement - we keep swapping from https://www.sponsorname.com to https://sponsorname.com as we seed so we're doing a lot of unnecessary database updates.

Maybe we move the website to the event-organization sponsor table, and then have a domain or select the first website for the organizations table?

Or a major data clean-up effort - perhaps during a refactor to sponsors?

Gemfile
```diff
@@ -262,6 +262,9 @@ group :development do
   # Guard for watching file changes and auto-importing [https://github.com/guard/guard]
   gem "guard"
 
+  # Benchmark is no longer included in the standard lib - used for benchmarking
+  gem "benchmark"
```
ChaelCodes, line 266: I keep adding and removing this, so I'll just add it and keep it. :)

PR #1645, opened 2026-04-15
Allow to follow an Event Series

marcoroth:
## Description

In order to prepare for the migration of the Meetup events we want to be able to transform the current "Attended as Visitor"  people have marked on an existing Event Series.

Since Meetups weren't actual events before, people could only mark themselves as a visitor of the whole series, and not of an indivual event.

Since can cannot automatically migrate that data, we'll introduce the concept of "subscriptions" for event series. The idea is that you can follow the series you are interested in and get notified about new events of that series.

Instead of loosing the data of people marking a "meetup series" as visited, we'll convert the data to be a subscription of that series.

## Screenshots

CleanShot 2026-04-15 at 03 56 24@2x

CleanShot 2026-04-15 at 03 56 30@2x

I went with "Subscription" for the model and "Follow/Unfollow" as the actions in the UI, happy to change either if you have a better idea.

## Testing Steps

* Open the show page of an event series
* Click "Follow Series" to create a subscription
* Click "Unfollow Series" to remove the subscription

## References

https://github.com/ChaelCodes/rubyevents/issues/11 

/cc @ChaelCodes (as we discussed this at Tropical on Rails)


ChaelCodes:
I have a lot of thoughts about this one. Maybe we can talk about it on Discord later - after you're back from Kaigi.

PR #1646, opened 2026-04-15
feat(events): add filter by kind to profile events

### LINKED ISSUE #1637: Filter by Event Kind on Profile
# Description

I think it would be good to filter Attended Events on the Profile by kind. It should be "All" + all kinds in the [enum on the event.](https://github.com/rubyevents/rubyevents/blob/4e4bdd5ec18e4956be03e9a59d0b535822852dd8/app/models/event.rb#L200)

Image

Let's use the toggle from the Events page for the UI.
Image

beatrizmitre:
## Description
This PR adds a filter by kind to the profile events page.
A few observations here:

- I have refactored the `app/views/shared/_filter_buttons.html.erb` to be used as the tabs with event kinds, because it seemed like they had similar responsibilities. If this refactor is not supposed to happen, I have a previous commit that I can use as well - that won't touch this partial.
- Should I also add an empty state message if `Past Events` or `Due Events` are empty?

Any other feedback, of course,  is more than welcome!

## Screenshots
https://github.com/user-attachments/assets/044536f5-1d7c-4209-8a80-ef8bcec164a3

## Testing Steps
1. Make sure you have a mix of past and future events, of a variety of kinds, in your seed logged user
2. Navigate to https://localhost:3000/profiles/<your_profile>/events
3. Navigate through the kinds in the same way of the video

## References
- closes #1637

### DIFF
--- a/app/javascript/controllers/events_filter_controller.js
+++ b/app/javascript/controllers/events_filter_controller.js
@@ -19,6 +19,7 @@ export default class extends Controller {
       const isActive = button.dataset.kind === this.currentValue
       button.classList.toggle('btn-active', isActive)
       button.classList.toggle('btn-primary', isActive)
+      button.classList.toggle('tab-active', isActive)
     })
   }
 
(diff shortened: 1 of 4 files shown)


ChaelCodes (approved):
Omg! This looks fantastic!! You absolutely killed it! 🙌  Love the PR description too!

PR #1648, opened 2026-04-15
Add validation for missing `speakers` field in `videos.yml`

Intrepidd:
cc @ChaelCodes , following our conversation in  #1618 

Probably a naive approach, let me know if there's anything that needs to be improved


lib/tasks/validate.rake
```diff
@@ -100,6 +100,62 @@ namespace :validate do
...
+
+    files.each do |file|
+      data = YAML.load_file(file)
+      relative_path = file.sub("#{Rails.root}/data/", "")
+
+      Array(data).each_with_index do |video, index|
+        video_id = video["id"] || video["video_id"] || "index #{index}"
+        video_title = video["title"] || video["event_name"] || "unknown video"
+
+        if video["talks"].nil? || video["talks"].empty?
```
ChaelCodes:
```suggestion
        if (video["talks"].nil? || video["talks"].empty?) && !video["video_provider"].in?(["children", "not_recorded"])
```

This works great! Some of our meetups are causing trouble though. The meetups are structured as a video with individual talks associated, so there's a lot of meetups that don't have talks. Filtering out these two providers catches most of those. Plus, if it wasn't recorded, we're unlikely to ever find out who those speakers were. :/

Intrepidd: Thanks, applied your suggestion !

marcoroth:
We might be able to improve the approach here with https://github.com/rubyevents/rubyevents/pull/1692 and https://github.com/rubyevents/rubyevents/pull/1693

ChaelCodes:
This looks great - unfortunately this has been resolved in the PR @marcoroth referenced, and then was refactored into a separate validator in #1706.

Thank you so much for your contribution! I wish I'd merged it earlier! Sorry. 😅

marcoroth:
Are you sure @ChaelCodes? If I understood right this is about a meetup nothing having a `speakers:` key, no? Or maybe I understood this wrong 🙈

ChaelCodes:
I thought this one was about speakers existing in speakers.yml.

Do we want to say that meetups can't have a speakers: [] key?

ChaelCodes:
https://github.com/rubyevents/rubyevents/pull/1618#pullrequestreview-4107881985

I did say this. 

So if the speakers: field exists, we want a - TODO so we can find it. 

But we should support it not existing, so things like Ruby Burgers and casual non-talk meetups are supported.

At least until the meetups refactor goes in.

PR #1649, opened 2026-04-16
Handle missing event visual asset configuration

### LINKED ISSUE #1638: Do Not Error on Missing Assets
# Description

When we add visual assets, we also add 3 fields to the event.yaml - `featured_background`, `featured_color`, and `banner_background`. These fields should **only** be set if the assets are available. However, the current setup - a big error message if the colors are set and the banner is missing is too aggressive.

We see this banner if you:
- Copy/Paste the event yaml without generating assets
- Have seeded the data and swapped to a branch without the assets
- Are progressively filling in data and haven't made it to the assets yet

These are all **common** use cases, and the error stops development and work in its track until assets are generated or the fields are removed and the database is updated. The /events page is not visible if _any_ upcoming event is missing assets.

Image

Instead of raising an error for a visual bug, it would good to _ignore_ those values unless a banner image is defined.

viniciusalonso:
## Description
Do not raise errors when event visual asset fields are configured but missing corresponding YAML data

## Screenshots
Screenshot 2026-04-15 at 23 00 57

## Testing Steps
* Pick an event (e.g., RailsConf 2017)
* Remove the fields `featured_background`, `featured_color`, or `banner_background` from `data/railsconf/railsconf-2017/event.yml`
* Restart the server
* Access the event page: http://localhost:3000/events/railsconf-2017

## References
- closes #1638


marcoroth (commented):
Hey @viniciusalonso, thanks for taking this on! 🙏🏼 

I wonder if instead we could validate the correctness/presence of the assets and the colors in the [`validate.rake`](https://github.com/rubyevents/rubyevents/blob/ebebee0bf0a3d95d54730365fc1f06d1653b0f1f/lib/tasks/validate.rake) file, so that we at least catch these inconsistencies in CI

viniciusalonso:
Hey @marcoroth I updated my PR.

The task checks that every event with visual configuration (featured_background or featured_color set in event.yml) has all 5 asset files
  present under app/assets/images/events/<series>/<event>/.

  The required assets are: banner.webp, card.webp, avatar.webp, featured.webp, poster.webp.

  134 events are failing, split into two groups:

  Group 1 — Only poster.webp is missing (~110 events)
  These events already have banner.webp, card.webp, avatar.webp, and featured.webp. poster.webp appears to be a newly introduced asset type
  that hasn't been generated for existing events yet.

  Group 2 — All 5 assets are missing (~24 events)
  These are mostly older events (e.g. euruko-2003 through euruko-2015, rubyconf-2001 through rubyconf-2008, railsconf-2006/2007, etc.) that
  have color configuration in event.yml but no image assets at all.

  Questions:
  - Should poster.webp be removed from the required assets list until it has been generated for all existing events?
  - For the older events missing all assets — is there a generation script/rake task that should be run, or should the visual configuration be
  removed from those event.yml files?


https://github.com/rubyevents/rubyevents/actions/runs/24543631888/job/71754497842?pr=1649

ChaelCodes:
Hey! I love the validation direction here. But I think there's been some confusion.

1. We want to _ignore_ the configured values when there aren't assets present.
2. For rubyconf + friends, there are default assets. So for [RubyConf 2001](https://www.rubyevents.org/events/rubyconf-2001) the configuration is correct. And [Euruko](https://www.rubyevents.org/events/euruko-2013).
3. We only care about banner.webp and featured.webp for the yaml configured values.

Those values are used to add padding to the banner, and to color text, etc in our featured list and on the homepage.
People copy/paste them a lot without setting up assets. We don't want an error, because that's blocking development, but we also don't want to use/set those values if we're using the default RubyEvents assets.

viniciusalonso:
I've just updated the PR @ChaelCodes  @marcoroth

lib/tasks/validate.rake
```diff
@@ -499,6 +529,147 @@ namespace :validate do
...
+          current << char
+        end
+      end
+      layers << current
+      return true if layers.all? { |layer| layer.strip.match?(/\A[\w-]+\(/) }
+    end
+
+    # CSS named colors
+    CSS_NAMED_COLORS.include?(normalized.downcase)
+  end
```
ChaelCodes: This is all a bit much. I think we just wanted to check if the values were present. Does this support gradient functions also?

viniciusalonso: Yes, it accepts CSS colors and functions. I’d prefer to keep it simple and only validate their presence.

viniciusalonso:
@ChaelCodes fixed your last comment.

ChaelCodes (approved):
Thank you for this! I'm sorry for the back and forth, but this will be incredibly helpful when building new events!

PR #1655, opened 2026-04-17
fix(events): hide profile events filter if profile has zero events

beatrizmitre:
## Description
Noticed that, on [my PR](https://github.com/rubyevents/rubyevents/pull/1646) that was merged earlier, I forgot to address the case when the user has no participated events. So this is a tiny fix to make sure we only display the filter bar if at least one event is present.

## Screenshots
Before:
image

After:
image

## Testing Steps
1. Delete all the `participated_events` from your seeds logged user
2. Navigate to https://localhost:3000/profiles/<your_profile>/events
3. Check if the filter bar showing `All` is gone

## References
Related to the previously [closed issue](https://github.com/rubyevents/rubyevents/issues/1637)

### DIFF
--- a/app/views/profiles/_events.html.erb
+++ b/app/views/profiles/_events.html.erb
@@ -6,7 +6,9 @@
 
 <% cache [user, events, Current.user] do %>
   <div data-controller="events-filter" data-events-filter-current-value="all">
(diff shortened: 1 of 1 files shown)


ChaelCodes (approved):
Thank you!

PR #1666, opened 2026-04-24
ruby.uy past events

elcuervo:
## Description

From https://ruby.uy/feed.xml

- **Add Ruby UY events**
- **Add Ruby UY speakers**

## Testing Steps

https://www.rubyevents.org/events/ruby-montevideo


marcoroth (approved):
Thank you for adding these events, @elcuervo! 🙏🏼

PR #1669, opened 2026-04-26
Update tickets URL for Blastoff Rails 2026 event

travisdock:
## Description
There was an old ticket url that I neglected to update. It is broken now so this fixes it.


marcoroth (approved):
Thanks @travisdock! 🙏🏼

PR #1670, opened 2026-04-27
First contribution - Fix: ActionView::StrictLocalsError when loading contributions via Turbo

cvaldivia83:
Hi @ChaelCodes , 

It was great meeting you at Rails Girls São Paulo! You mentioned that I could take a look at the contributions section of your website, so I've put together this PR with the fixes for the rendering issue. 

Let me know if you’d like me to change anything. Hope you're doing well!

Best,
Carla Valdivia

## Description
`contributions/views/show.html.erb `- replaced strict locals with instance variables in contributions partials to ensure proper rendering when Turbo frames load them dynamically.   

  When Turbo frame requests rendered partials with strict locals declarations, Rails raised                             
  `ActionView::StrictLocalsError `because the local variables weren't being passed correctly through Turbo's rendering    
  pipeline.  

## Screenshots
Before changes: 
Screenshot 2026-04-26 at 22 09 46

After changes:
Screenshot 2026-04-26 at 21 54 24

## Testing Steps
Visit `/contributions` and click any menu option (speakers without github handles, talks without slides)

### DIFF
--- a/app/views/contributions/_events_without_dates.html.erb
+++ b/app/views/contributions/_events_without_dates.html.erb
@@ -1,12 +1,12 @@
-<%# locals: (events_without_dates:) %>
+<%# locals: () %>
 
 <%= turbo_frame_tag "events_without_dates" do %>
-  <h2 class="mb-4">Events without conference dates (<%= events_without_dates.flat_map(&:last).count %>)</h2>
+  <h2 class="mb-4">Events without conference dates (<%= @events_without_dates.flat_map(&:last).count %>)</h2>
 
   <article class="prose mb-6">This section lists events that are missing conference dates. Adding conference dates allows us to show when this event took place, so we could show them in a calendar or similar.</article>
 
   <div id="events-without-dates" class="grid sm:grid-cols-2 lg:grid-cols-3 xl:grid-cols-3 gap-4 min-w-full mb-6">
-    <% events_without_dates.each do |file_path, events| %>
+    <% @events_without_dates.each do |file_path, events| %>
       <%= link_to "https://github.com/rubyevents/rubyevents/edit/main/#{file_path}", id: "event-#{file_path}", class: "hover:bg-gray-100 p-4 rounded-lg border bg-white", target: :_blank do %>
         <div class="flex flex-col">
           <h3 class="line-clamp-1"><pre>Data File: <%= file_path %></pre></h3>
--- a/app/views/contributions/_events_without_location.html.erb
+++ b/app/views/contributions/_events_without_location.html.erb
@@ -1,12 +1,12 @@
-<%# locals: (events_without_location:) %>
+<%# locals: () %>
 
 <%= turbo_frame_tag "events_without_location" do %>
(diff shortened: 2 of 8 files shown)


cvaldivia83:
The CI failure is due to the `erb-no-instance-variables-in-partials` Herb linter rule, which enforces that partials should use strict local variables instead of instance variables.
My PR intentionally uses instances variables because: 
- I found that when turbo frames dynamically rendered the partials it did not pass strict local variables, causing the `ActionView::StrictLocalsError`. 

I am sorry, but I could not find any other solution. :/

ChaelCodes:
Thanks so much for your contribution! Because this breaks CI, it's going to take me a bit longer to review it. I have to evaluate if there's an alternate approach that could pass CI and if Turbo's rendering pipeline can't handle strict_locals, we should send them an issue (which I'll happily handle after verifying).

JuanitoFatas:
Hi @cvaldivia83 👋🏻 — drive-by from another contributor here. 👋 Thanks for taking this on, and congrats on your first contribution!

Saw your CI note about the Herb rule `erb-no-instance-variables-in-partials` — I think there's actually a path that could fix the bug *and* gets CI green, which I would like to share.

The original error was a small misuse of Rails' `render` API. From the [API docs][render-api], the signature is:

```ruby
render(options = {}, locals = {}, &block)
```

> Without the rendering mode, the second argument can be a Hash of local variable assignments for the template.

So when we call `render "partial", locals: {...}` with a string as the first argument, Rails takes the trailing hash as the locals directly — meaning `locals:` itself ended up as a single local named `:locals`. That's what tripped strict locals, not Turbo.

And from the [Rails Guide on strict locals][strict-locals]:

> Action View partials, by default, accept any locals as keyword arguments. To define what locals a partial accepts, add a `locals:` magic comment to the partial's source code.

So the minimal fix could be dropping the `locals:` wrapper but keep the values:

```erb
<%= render "contributions/speakers_without_github", speakers_without_github: @speakers_without_github %>
```

That keeps strict locals in the partials, and the partials themselves can stay on their original `<%# locals: (foo:) %>` declarations using local `foo` (no `@foo`). Turbo frames work fine with strict locals — it was simply the `render` call shape that was off. That’s how I tried to do in this PR https://github.com/rubyevents/rubyevents/pull/1687/changes before seeing yours 😍

The nice side effect is that this lines up with what the Herb rule is asking for. Its error message — *"Pass `@speakers_without_github` as a local variable instead"* — is essentially pointing at the same convention, so reverting the partials to strict locals also gets CI green without disabling any rules.

Thanks again for the PR! 🙇🏻 

[render-api]: https://api.rubyonrails.org/classes/ActionView/Helpers/RenderingHelper.html
[strict-locals]: https://guides.rubyonrails.org/action_view_overview.html#strict-locals

ChaelCodes:
Thanks @JuanitoFatas for looking into this! It's been a busy few weeks, and this is the first chance I've had to investigate. @cvaldivia83, can you try this out and let me know how it goes?

PR #1675, opened 2026-05-01
Add Tech Kafi #ruby meetup

codez:
## Description
Add Tech Kafi #ruby meetup in Bern, Switzerland.


marcoroth (approved):
Merci Pascal! 🙏🏼

Issue #1317, opened 2026-01-14
Meetup CFP for Toronto Ruby


marcoroth:
https://github.com/toronto-ruby/talks

We should add it, similar to how we have SF Ruby and LRUG:

<img width="2706" height="626" alt="Image" src="https://github.com/user-attachments/assets/f89f2ab6-ccc6-4e65-a8d3-b0be8b83e361" />

esmale:
@marcoroth @ChaelCodes I think this issue should be closed as a result of #1684

marcoroth:
Thank @esmale! 🙏🏼

PR #1684, opened 2026-05-03
Add CFP for Toronto Ruby Meetup

lisgrilo:
## Description
This PR adds a Call for Proposals (CFP) for the Toronto Ruby meetup.

## Testing Steps
1. Run the seed command to create the CFP:
`bin/rails db:seed:event_series[toronto-ruby]
`

2. Visit the CFP page:  
http://localhost:3000/cfp  
- Confirm that the Toronto Ruby CFP appears in the list

3. Click on **"Submit your Talk Proposal"** for the Toronto Ruby meetup:  
http://localhost:3000/events/toronto-ruby-meetup/cfp  
- Verify that it redirects to:  
  https://github.com/toronto-ruby/talks  

## References
#1317

### DIFF
--- a/data/toronto-ruby/toronto-ruby-meetup/cfp.yml
+++ b/data/toronto-ruby/toronto-ruby-meetup/cfp.yml
@@ -1,0 +1,2 @@
+---
+- link: "https://github.com/toronto-ruby/talks"


data/toronto-ruby/toronto-ruby-meetup/cfp.yml
```diff
@@ -0,0 +1,2 @@
+---
+- link: "https://github.com/toronto-ruby/talks"
```
lisgrilo, line 2: I noticed that in some parts of the codebase we explicitly set the CFP name (even when it’s just "Call for Proposals"), while in others we rely on the default fallback.

Opening this thread to discuss which approach you all think is best 🙂

marcoroth: Thanks for this pull request @lisgrilo! 🙏🏼 

Great catch! Initially we didn't show the CFP name in the UI anywhere, but now we are actually showing it and implemented the "Call for Proposals" fallback for CFPs that didn't define it yet.

Currently we show it here in the UI:

CleanShot 2026-05-04 at 13 28 21@2x


And on the event CFP page:

CleanShot 2026-05-04 at 13 28 46@2x


I think its fine for now to keep it blank, though it might be a great idea to retrofit all the `cfp.yml` files and the [`CFPSchema`](https://github.com/rubyevents/rubyevents/blob/4eeb4b0b01fef7830bc6413e82caad808f083383/app/schemas/cfp_schema.rb) to enforce a name, even if its just `"Call for Proposals"` so they are explicit and consistent.

It doesn't need to be in the pull request, but feel free to work on it as a follow up 🙌🏼 

Thank you! 🙏🏼

PR #1687, opened 2026-05-04
Fix "Content missing" on /contributions sub-tabs

JuanitoFatas:
The render calls in contributions/show.html.erb used `render "partial", locals: {...}`. With the positional-string form, the trailing hash *is* the locals — so `locals:` became a single local named `:locals` instead of being unpacked. Strict-locals partials raised StrictLocalsError, the show action 500'd, and Turbo rendered "Content missing" in the affected frames.

Switched to the shorthand `render "partial", key: value` so locals are passed through correctly.

## Testing Steps

- https://www.rubyevents.org/contributions

  https://github.com/user-attachments/assets/ffebab42-4a5c-453d-8a0b-39c2d36a1f5f

### This PR

- http://127.0.0.1:3000/contributions
- click all sub tabs

https://github.com/user-attachments/assets/c9a9e0b7-a39b-4ed5-bf5e-e5280342541f

### DIFF
--- a/app/views/contributions/show.html.erb
+++ b/app/views/contributions/show.html.erb
@@ -1,16 +1,16 @@
 <% case @step %>
 <% when :speakers_without_github %>
-  <%= render "contributions/speakers_without_github", locals: {speakers_without_github: @speakers_without_github} %>
+  <%= render "contributions/speakers_without_github", speakers_without_github: @speakers_without_github %>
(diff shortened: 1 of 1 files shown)


JuanitoFatas:
Just discovered there is already an PR for this
https://github.com/rubyevents/rubyevents/pull/1670/

Closing this as duplicate 😅

PR #1696, opened 2026-05-05
Latest Talks should display the talks that occurred most recently

### LINKED ISSUE #1674: Homepage doesn't show latest talks/events
When viewing the homepage the sections with latest talks and events do not show the actual latest talks or events. 

The reason this is a problem is that it gives a false first impression. It appears the community is less active than it actually is. Yes, a user can obviously click through and see more accurate information but how many people are not just based of the assumption nothing has occurred this year?

These are from 2024 even though it's currently 2026:
Image

But if we click the See All Talks link we get the expected view:

Image

The same happens with Events; nothing from 2026 appears under latest:

Image

esmale:
Ordering on `published_at` results in the talks that were published most recently being displayed as the "Latest Talks", meaning if there's a significant delay between giving the talk and the video being published, it would be considered more recent than talks that were given since then.  Ordering by date seems like the more desired behavior here.

## Description

Changed `Talk.watchable.with_speakers.order(published_at: :desc)` to `Talk.watchable.with_speakers.order(date: :desc)`

## Testing Steps

1. Bring up the main page: localhost:3000
2. Scroll down to the "Latest Talks" section
3. The displayed talks should all be from "Ruby Community Conference Winter Edition 2026", as that currently appears to be the most recent set of talks that are watchable.  At the time of writing this PR the displayed latest talks are all from "Ruby Montevideo Meetup July 2025"
Screenshot 2026-05-05 at 7 16 22 PM

## References

- closes #1674

### DIFF
--- a/app/controllers/page_controller.rb
+++ b/app/controllers/page_controller.rb
@@ -3,7 +3,7 @@ class PageController < ApplicationController
 
   def home
     home_page_cached_data = Rails.cache.fetch("home_page_content", expires_in: 1.hour) do
-      latest_talks = Talk.watchable.with_speakers.order(published_at: :desc).limit(10)
+      latest_talks = Talk.watchable.with_speakers.order(date: :desc).limit(10)
       {
         talks_count: Talk.count,
         speakers_count: User.speakers.count,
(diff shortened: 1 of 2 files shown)


ChaelCodes:
Thank you for updating this! Looks great. Would you mind updating the heading from "Latest Talks" to "Latest Videos". I think that might've caused some confusion as well.

esmale:
@ChaelCodes done!

ChaelCodes (approved):
Thank you for picking this one up! 🙌  Nice work!

PR #1697, opened 2026-05-06
Show Today on the Schedule

### LINKED ISSUE #1628: Show Today on the Schedule
# Description

If I'm viewing the schedule on RubyEvents at a conference, I have to keep selecting the correct day. 

If the schedule page is loaded before or after the conference, default to the first day. 

If the schedule page is opened during the event, default to the current day.

esmale:
Add selected_schedule_day helper method, and use it to determine the day to show:

If today is:
* before or after the event, display the first day of the event
* during the event, display today

## Description

Added:
* a helper (Events::SchedulesHelper) with the `selected_schedule_day` method
* invoking it from the events/schedules_controller
* tests for the new helper

## Testing Steps

1. Visit the schedule page for an event that has already happened, and the first day of the schedule should display by default
2. Visit the schedule page for an event that hasn't happened yet, and the first day of the schedule should display by default
3. Visit the schedule page for an event that is currently in progress, and whichever tab in the schedule corresponds to "today" should be displayed by default.  This is the trickiest part of the process to test since it requires a multi-day event to currently be in progress and on day 2 or higher in order to see behavior different from cases 1 and 2 above.

## References

- closes #1628

### DIFF
--- a/app/helpers/events/schedules_helper.rb
+++ b/app/helpers/events/schedules_helper.rb
@@ -1,5 +1,6 @@
 module Events::SchedulesHelper
-  def selected_schedule_day(days, today: Time.zone.today.to_s)
+  def selected_schedule_day(days)
+    today = Time.zone.today.to_s
     days.detect { |day| day["date"] == today } || days.first
   end
 end
--- a/app/helpers/events/schedules_helper.rb
+++ b/app/helpers/events/schedules_helper.rb
@@ -1,5 +1,5 @@
 module Events::SchedulesHelper
   def selected_schedule_day(days, today: Time.zone.today)
-    days.detect{ |day| day["date"] == today } || days.first
+    days.detect { |day| day["date"] == today } || days.first
   end
 end
--- a/app/helpers/events/schedules_helper.rb
+++ b/app/helpers/events/schedules_helper.rb
@@ -1,0 +1,5 @@
+module Events::SchedulesHelper
+  def selected_schedule_day(days, today: Time.zone.today)
+    days.detect{ |day| day["date"] == today } || days.first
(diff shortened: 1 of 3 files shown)


app/helpers/events/schedules_helper.rb
```diff
@@ -0,0 +1,5 @@
+module Events::SchedulesHelper
+  def selected_schedule_day(days, today: Time.zone.today)
+    days.detect { |day| day["date"] == today } || days.first
```
marcoroth, line 4: Don't we need to compare strings with each other here? 

```suggestion
    days.detect { |day| day["date"] == today.to_s } || days.first
```
Looks like comparing a string against a Date is not equal:
```
rubyevents(dev):004> Time.zone.today == "2026-05-06"
=> false
rubyevents(dev):005> "2026-05-06" == Time.zone.today
=> false
rubyevents(dev):006>
```

esmale: Yep!  We definitely don't want to be comparing Dates vs Strings.  That was very silly of me.  

I've also updated the tests to more accurately reflect how the helper should be used (not passing the `today` date in directly).  Though now that I've done that, I'm beginning to think that `today` shouldn't even be a method parameters/kwarg and should instead just be set in the body of the helper...

ChaelCodes: Yeah, I think I like it in the body of the helper too. I don't want people passing it in - if it's in the helper, there's less chance of a mistake.

ChaelCodes (approved):
Thank you!! 🙌  I'm so excited to try this out at my next event!

PR #1706, opened 2026-05-07
Errors and validations

ChaelCodes:
## Description

It would be good to have validations in multiple places. We can confirm the schema of what the generator produces, and we can confirm the output in validate.rake. This will also help with establishing a standard format for error messages - which will assist in ensuring GitHub Actions reports errors in an easy to view way.

## Screenshots

Screenshot 2026-05-07 at 6 58 27 AM
Screenshot 2026-05-07 at 6 58 54 AM
Screenshot 2026-05-07 at 7 37 40 AM

### Errors
image
image
image
image
image

#### Needs Improvement
image
image

## Testing Steps

1. push an evil commit with lots of bad yaml
2. See failing CI
3. revert the bad commit

## References

- https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message
- https://github.com/marcoroth/yerba#structured-navigation
- https://github.com/marcoroth/yerba/commit/d5de188862fbc286b42c1c8a3739e406e6460bb5


app/models/static/validators/schema.rb
```diff
@@ -0,0 +1,45 @@
...
+      def initialize(file_path:, schema:)
+        @file_path = file_path
+        @schema = schema
+      end
+
+      def applicable?
+        return false unless File.exist?(@file_path)
+        @file_path.match?(/data\/[^\/]+\/[^\/]+\/(event|venue|schedule)\.yml$/) ||
+          @file_path.match?(/data\/[^\/]+\/series\.yml$/)
+      end
```
marcoroth, line 24: What do you think about something like:
```suggestion
      PATTERNS = [
        "data/*/*/event.yml",
        "data/*/*/venue.yml",
        "data/*/*/schedule.yml",
        "data/*/series.yml"
      ].freeze
      
      def applicable?
        return false unless File.exist?(@file_path)
      
        PATTERNS.any? do |pattern|
          File.fnmatch?(pattern, @file_path, File::FNM_PATHNAME)
        end
      end
```

marcoroth:
@ChaelCodes love the direction where this is headed! 😍

app/models/static/validators/schema.rb
```diff
@@ -30,14 +30,17 @@ def errors
...
 
         raw_errors.map do |e|
+          data_pointer = e["data_pointer"].gsub(/\A\//, "").tr("/", ".") || ""
+          location = document[data_pointer]&.location
           Static::Validators::Error.new(
             "#{e["error"]} at #{e["data_pointer"]}",
             file_path: @file_path,
-            line: 1
+            line: location&.start_line || 1,
+            end_line: location&.end_line
```
ChaelCodes: I got the locations working! This is going to look so good!

app/models/static/validators/speaker_exists.rb
```diff
@@ -0,0 +1,65 @@
...
+                errors << Static::Validators::Error.new(
+                  "Speaker '#{name}' not found in speakers.yml",
+                  file_path: @file_path,
+                  line: 1,
+                  end_line: 1
+                )
+              end
+            end
+          end
+        end
```
ChaelCodes, line 81: Location logic doesn't currently work.

app/models/static/validators/validator.rb
```diff
@@ -0,0 +1,15 @@
+module Static
```
ChaelCodes: Skip this file - not ready for it yet.

app/models/static/validators/speaker_exists.rb
ChaelCodes: Fixed it. >)

PR #1715, opened 2026-05-12
Upgrade to Yerba 0.5 and cleanup validators

marcoroth:
### Summary

  - Upgrade yerba from 0.4.2 to 0.5.0
  - Add JSON schema validation steps to all data file types in the Yerbafile
  - Migrate Ruby API calls from removed methods (`get`, `get_value`, `at_path`) to new API (`value_at`, `[]`)
  - Simplify `SpeakerExists` validator from manual index-walking per file to using `SpeakersFile#missing_speaker_references` with `Collection.get`, this validates all video files in parallel with file/line metadata
  - Extract `VIDEO_SPEAKER_SELECTORS` constant and `all_speaker_references` / `missing_speaker_references` methods on `SpeakersFile`
  - Make `Error#end_line` optional

### DIFF
--- a/app/models/static/validators/speaker_exists.rb
+++ b/app/models/static/validators/speaker_exists.rb
@@ -3,6 +3,26 @@
 module Static
   module Validators
     class SpeakerExists
+      PATTERNS = ["**/videos.yml"].freeze
+
+      def initialize(file_path:)
+        @file_path = file_path.to_s.sub("#{Rails.root}/", "")
+      end
+
+      def applicable?
+        return false unless File.exist?(@file_path)
+
+        PATTERNS.any? do |pattern|
+          File.fnmatch?(pattern, @file_path, File::FNM_PATHNAME)
+        end
+      end
+
+      def errors
+        return [] unless applicable?
+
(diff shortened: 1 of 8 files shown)


app/models/static/validators/speaker_exists.rb
```diff
@@ -3,84 +3,18 @@
...
-        @file_path = file_path
-      end
-
-      def applicable?
-        return false unless File.exist?(@file_path)
-
-        PATTERNS.any? do |pattern|
-          File.fnmatch?(pattern, @file_path, File::FNM_PATHNAME)
-        end
-      end
```
ChaelCodes, line 19: Can you keep the applicable? I'm thinking about throwing all our files at an array of validators.

ChaelCodes (commented):
😍👑

app/models/static/validators/speaker_exists.rb
marcoroth: Do we just want to statically return `true`?

Since we are doing it globally and not on a per-file basis I wasn't sure if it made sense to keep it. But if want to follow the same API for all validators then it might make sense to just make it compatible.

PR #1717, opened 2026-05-13
CfpGenerator can mutate values

ChaelCodes:
## Description

Use new Yerba library to find and delete an existing CFP record.
Use the existing record's values to populate and append the new record.
Better USAGE examples.

When Yerba deletes a Record, the whitespace stays around, so the cfp record doesn't introduce more whitespace. Otherwise it'll just add lots and lots of whitespace on updates.

## Testing Steps

- Tests, also modify an existing CFP

```bash
bin/rails g cfp --event <event-slug> --name "Call for Proposals" --end-date <new-end-date>
```

## References

- closes #


lib/generators/cfp/cfp_generator.rb
```diff
@@ -5,19 +5,44 @@
...
+  class_option :close_date, type: :string, desc: "CFP close date (YYYY-MM-DD)", group: "Fields"
+
+  CFP = Struct.new("CFP", :name, :link, :open_date, :close_date)
+
+  def cfp_file
+    @cfp_file = File.join([event_directory, "cfp.yml"])
+  end
+
+  def ensure_cfp_file_exists
+    template "header.yml.tt", @cfp_file unless File.exist?(@cfp_file)
```
ChaelCodes, line 20: Call `cfp_file` everywhere instead of `@cfp_file`. Thor automatically runs all public methods, but it's still nice to have clean code.

PR #1724, opened 2026-05-15
Fix sticker count on tabs

shkm:
## Description
The sticker count on the tab in user profiles was incorrectly counting
only the actual number of events, not the number of stickers (some have
more than one).

In doing this, I introduced a `Sticker.for_user` as a small abstraction,
similar to the `Stamp.for_user` helper. This also makes it more
testable.

None of the fixtures included events with multiple stickers, so I pulled
in railsconf 2025 which has two stickers. This broke the
archive_controller_test, hence the change there.

## Screenshots
image

## Testing Steps
- Browse to a user with more stickers than events (e.g. anyone who attended Rubycon Italy, probably)
- Check sticker count in tab vs actual stickers shown

## References
None

### DIFF
--- a/app/controllers/concerns/profile_data.rb
+++ b/app/controllers/concerns/profile_data.rb
@@ -52,7 +52,7 @@ module ProfileData
 
   def set_mutual_events
     @mutual_events = if Current.user
-      @user.participated_events.where(id: Current.user.participated_events).distinct.order(start_date: :desc)
+      @user.participated_events.where(id: Current.user.participated_events).order(start_date: :desc)
     else
       Event.none
     end
--- a/app/controllers/concerns/profile_data.rb
+++ b/app/controllers/concerns/profile_data.rb
@@ -17,7 +17,7 @@ module ProfileData
     @talks = @user.kept_talks
     @events = @user.participated_events
     @stamps = Stamp.for_user(@user)
-    @events_with_stickers = @events.select(&:sticker?)
+    @stickers = Sticker.for_user(@user, events: @events)
     @involvements_by_role = @user.event_involvements.group_by(&:role).transform_values(&:any?)
     @countries_with_events = @events.group_by(&:country_code).any? ? [true] : []
     @topics = @user.topics.approved.tally.sort_by(&:last).reverse.map(&:first)
--- a/app/controllers/page_controller.rb
+++ b/app/controllers/page_controller.rb
@@ -106,6 +106,7 @@ class PageController < ApplicationController
 
   def stickers
     @events = Event.all.select(&:sticker?)
+    @stickers = @events.flat_map(&:stickers)
   end
 
   def contributors
--- a/app/controllers/profiles/events_controller.rb
+++ b/app/controllers/profiles/events_controller.rb
@@ -2,7 +2,7 @@ class Profiles::EventsController < ApplicationController
   include ProfileData
 
   def index
-    @events = @user.participated_events.includes(:series).distinct.in_order_of(:attended_as, EventParticipation.attended_as.keys)
+    @events = @user.participated_events.includes(:series).in_order_of(:attended_as, EventParticipation.attended_as.keys)
     event_participations = @user.event_participations.includes(:event).where(event: @events)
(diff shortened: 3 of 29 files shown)


marcoroth (commented):
Thanks @shkm! 🙏🏼 

It looks like its now showing the stickers for certain events twice:

Image

I also noticed the sticker count is not the same on the `/profiles/:slug` page compared to any of the tab subpages, like `/profiles/:slug/stickers`:

`/profiles/:slug`:
Image

`/profiles/:slug/stickers`:
Image

shkm:
Good catches, will see what's up!

shkm:
Okay, so this looks like a pre-existing bug.

The issue is when a user participates in an event multiple times; i.e. in different roles. From seed data, `crmne` has this problem with SFRuby: participating both as a speaker _and_ a keynote speaker. This gives us an event count of 7, with one being a duplicate. You'd notice this on the "Attended Events" tab, and relevant to our current issue, in duplicate stickers.

Deduplication (distinct) was happening in some places previously, which led to counts being correct in some places and not others.

Since the sticker fix is quite relevant here, I've added a new commit for that which fixes it at the root: making participated events distinct. From what I can tell this won't cause issues elsewhere, but I'm lacking as much context as you folks!

This happens to also fix the problem with attended events numbering.

There might be a potential follow-up here in handling speakers with multiple roles per event. Is it relevant that someone shows up as both a keynote speaker and a speaker, or is a keynote speaker good enough? And if they're showing up as both, should we not choose the "better" role for them when displaying their events? Right now I'm seeing that Carmine attended SFRuby as a speaker, but it probably makes more sense to show as a keynote speaker.

Chose not to fix that here, though, as it's irrelevant and might be a bit tricky.

Whew, sorry for the wall of text. Hope it's useful, though.

marcoroth (approved):
Awesome, thanks for catching and digging into this @shkm! 🙏🏼 

I think this PR is good to merge as-is as it's really already.

It would be great to also solve this issue you mentioned where speakers have two `event_participation` records. This also happens when you "mark your attendance" for an event and then you get also published as a speaker, then you end up with both a `visitor` and `speaker` participation record. We probably want to use the "highest" type when deduplicating the records.

If you are up for it, it would probably make sense to fix this in the `backfill:speaker_participation` rake task:
https://github.com/rubyevents/rubyevents/blob/05b2b8b37ec4a815679f3ebeee8be48e92442bcd/lib/tasks/backfill_speaker_participation.rake#L17

And actually, we might want to run this rake task daily in `recurring.yml`: https://github.com/rubyevents/rubyevents/blob/05b2b8b37ec4a815679f3ebeee8be48e92442bcd/config/recurring.yml so we don't have to manually go an clean these records up.

Thank you! 🙌🏼

shkm:
> Awesome, thanks for catching and digging into this @shkm! 🙏🏼
> 
> I think this PR is good to merge as-is as it's really already.
> 
> It would be great to also solve this issue you mentioned where speakers have two `event_participation` records. This also happens when you "mark your attendance" for an event and then you get also published as a speaker, then you end up with both a `visitor` and `speaker` participation record. We probably want to use the "highest" type when deduplicating the records.
> 
> If you are up for it, it would probably make sense to fix this in the `backfill:speaker_participation` rake task:
> 
> https://github.com/rubyevents/rubyevents/blob/05b2b8b37ec4a815679f3ebeee8be48e92442bcd/lib/tasks/backfill_speaker_participation.rake#L17
> 
> And actually, we might want to run this rake task daily in `recurring.yml`: https://github.com/rubyevents/rubyevents/blob/05b2b8b37ec4a815679f3ebeee8be48e92442bcd/config/recurring.yml so we don't have to manually go an clean these records up.
> 
> Thank you! 🙌🏼

Sounds good! I'd love to contribute more, but have a hard time making commitments because of time and life. I'll open an issue based on this if I get the chance though!

marcoroth:
@shkm no worries and rush. Feel free to just open an issue with the "Reference in new issue" option on the PR comment itself 🙌🏼

PR #1714, opened 2026-05-12
Use dates without timezones for full date coverage in ical export

dennispaagman:
This is a follow up of #1007 to fix a small bug, where events are shown with an offset if you are in a different time zone. I didn't catch this in development for some reason.

## Description

Currently the dates are returned with a timezone:

```
BEGIN:VEVENT
DTSTAMP:20260512T080332Z
UID:RUBYEVENTS-331
DTSTART:20260515T000000
DTEND:20260517T000000
DESCRIPTION:Balkan Ruby is a yearly conference held in Bulgaria and feature
 s 13 talks from various speakers.\n
LOCATION:Sofia\, Bulgaria
SUMMARY:Balkan Ruby 2026
URL;VALUE=URI:https://balkanruby.com
END:VEVENT
```

To make sure they show up as full day for everyone, we need dates without timezones. The `Icalendar::Values::Date` class is there for that.

After:

```
BEGIN:VEVENT
DTSTAMP:20260512T080706Z
UID:RUBYEVENTS-1
DTSTART;VALUE=DATE:20260515
DTEND;VALUE=DATE:20260517
DESCRIPTION:Balkan Ruby is a yearly conference held in Bulgaria and feature
 s 12 talks from various speakers.\n
LOCATION:Sofia\, Bulgaria
SUMMARY:Balkan Ruby 2026
URL;VALUE=URI:https://balkanruby.com
END:VEVENT
```

Also added a test to compare the full string output, which makes it easier to catch potential regressions in the output.

## Screenshots

Currently:

CleanShot 2026-05-12 at 10 01 02@2x

PR #1727, opened 2026-05-17
Small fixes/improvements for ical export

dennispaagman:
Another round, follow up of #1714.

Four things:
- No longer +1 the end date as that's not needed now we're using full dates instead of datetime ranges.
- Add support for cancelled events, with Baltic 2026 cancelled this now shows up properly in the exported event as well.
- Strips an unnecessary new line from the end of the description.
- Added a 'last modified' property, hopefully this will help with cache invalidation as I still see old data in my gcal from days ago (the full date fix is not propagated to all events for some reason). Not sure though, this is a bit of a shot in the dark 🤷

@ChaelCodes would you mind another look? 😇


marcoroth (approved):
Thank you @dennispaagman!

PR #1728, opened 2026-05-17
Validate Schema using Yerba

ChaelCodes:
## Description

I want the validations for files to match the validations for generator tests. Replace Static::Validators::Schema with same Yerba check logic OR replace generator tests with same Yerba check logic.

## References

- closes #


ChaelCodes:
- Couldn't find Yerba::Document.check
- Yerba::Document.apply raised -> hard to iterate over collection
- GitHub Actions annotations aren't linked to line number for sort_keys
- Multiple keys in one error
- schema errors (missing keys) is a bigger error than sort_keys - probably should come first
- Need ruby interface so generator validations match file validations

PR #1735, opened 2026-05-19
Fix copy for calendar url to not include whitespace

### LINKED ISSUE #1731: "Subscribe to calendar" copies URL with leading and trailing whitespace
Image

Image

dennispaagman:
`<pre>` preserves whitespace, so the copied text would have whitespace around it.

Fixes #1731

cc @marcoroth


marcoroth (approved):
Great, thanks for the quick fix @dennispaagman! 🙏🏼

Issue #1753, opened 2026-05-30
Ruby Retreat AU 2026 needs a venue URL


bigtiger:
The venue for Ruby Retreat AU 2026 ([`data/ruby-retreat-au/ruby-retreat-au-2026/venue.yml`](https://github.com/rubyevents/rubyevents/blob/main/data/ruby-retreat-au/ruby-retreat-au-2026/venue.yml)) is missing a website URL. It previously held the generator placeholder (`"Venue website url - Optional"`), which has been removed.

The venue is Land's Edge Chowder Bay, Sydney. Someone with the correct link should add the `url` field back with the real venue website.

marcoroth:
is this it?

https://www.landsedge.com.au

PR #1754, opened 2026-05-30
Add lint check for default venue values

bigtiger:
## What

Adds a venue linter that fails when a `venue.yml` still contains the placeholder values left behind by the venue generator. This catches venues that were scaffolded but never filled in before being committed.

## Details

- New validator `Static::Validators::DefaultVenue` flags default placeholders for `name`, `description`, `instructions`, `url`, and `address.display`. Where possible it sources the defaults directly from `VenueGenerator.class_options` so they can't drift from the generator; template-only literals and the bare `TODO` placeholder are listed explicitly.
- Wired into `lib/tasks/validate.rake` via a new `validate:venues` task and added to `validate:all` (the CI path).
- Added test coverage in `test/models/static/validators/default_venue_test.rb`.

## Data fix

The linter caught one pre-existing offender: `data/ruby-retreat-au/ruby-retreat-au-2026/venue.yml` still had the placeholder `url`. Removed the placeholder line (`url` is optional in `VenueSchema`). Follow-up to supply the real URL tracked in #1753.

## Testing

- `bin/rails test test/models/static/validators/default_venue_test.rb` — 11 runs, 0 failures
- `bin/rails validate:venues` — all venue.yml files pass

### DIFF
--- a/app/models/static/validators/default_venue.rb
+++ b/app/models/static/validators/default_venue.rb
@@ -1,0 +1,107 @@
+# frozen_string_literal: true
+
+require "generators/venue/venue_generator"
+
+module Static
+  module Validators
+    class DefaultVenue
+      def initialize(file_path:)
+        @file_path = file_path
+      end
+
+      PATTERNS = [
+        "**/venue.yml"
+      ].freeze
+
+      # Placeholder values that the venue generator leaves behind and that a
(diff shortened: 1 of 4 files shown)


bigtiger:
@ChaelCodes trying to protect myself in the future. 😆

ChaelCodes (approved):
Love this!! Great idea!

PR #1761, opened 2026-06-01
Calendar: fix multi day events by making dtend inclusive again

dennispaagman:
I removed this in #1727 based on wrong assumptions. I don't remember what caused me to remove it. This adds it back, but only in case the event span multiple days.

## Screenshots
Before:

- You can see Blastoff Rails, Baltic Ruby and Ruby Retreat being one day too short. The one day events are displayed properly.

CleanShot 2026-06-01 at 15 53 40@2x

After:
CleanShot 2026-06-01 at 15 49 48@2x

## References
- #1727

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -362,7 +362,11 @@ class Event < ApplicationRecord
       event.uid = "RUBYEVENTS-#{id}"
       event.last_modified = updated_at
       event.dtstart = Icalendar::Values::Date.new(start_date)
-      event.dtend = Icalendar::Values::Date.new(end_date)
+
+      if end_date > start_date
+        event.dtend = Icalendar::Values::Date.new(end_date + 1.day) # dtend is exclusive, add 1 day to make it inclusive
+      end
+
       event.summary = name
       event.description = description.strip
       event.location = static_metadata.location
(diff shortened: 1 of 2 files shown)


app/models/event.rb
```diff
@@ -362,7 +362,11 @@ def to_ical
       event.uid = "RUBYEVENTS-#{id}"
       event.last_modified = updated_at
       event.dtstart = Icalendar::Values::Date.new(start_date)
-      event.dtend = Icalendar::Values::Date.new(end_date)
+
+      if end_date > start_date
```
dennispaagman, line 366: do you think we should guard against events without `end_date` or `start_date`? There are a few in the database, but they never get rendered on the calendar export because of how the `EventsController` loads the records right now.

marcoroth (commented):
Thank you @dennispaagman!

Looks like this test failure might be legit?

```
Failure:
EventTest#test_to_ical_serializes_to_a_ical_formatted_string_with_the_event_details [test/models/event_test.rb:284]:
--- expected
+++ actual
@@ -2,7 +2,6 @@
 DTSTAMP:20260101T000000Z\r
 UID:RUBYEVENTS-822853486\r
 DTSTART;VALUE=DATE:20231026\r
-DTEND;VALUE=DATE:20231026\r
 DESCRIPTION:RailsWorld is a yearly conference held in Netherlands.\r
 LAST-MODIFIED:20260101T000000\r
 LOCATION:Amsterdam\\, Netherlands\r


bin/rails test test/models/event_test.rb:279
```

dennispaagman:
> Looks like this test failure might be legit?

Ah yeah you're right, fixed and added a test for multi day events as well.

marcoroth (approved):
Thanks! 🙏🏼

PR #1766, opened 2026-06-02
Remove Suggestions system

marcoroth:
Similar to #1765, the concept of suggestions is misleading, especially since we haven't been really maintaining and looking at them. I'm suggesting to remove them.


ChaelCodes:
I like this idea!

PR #1776, opened 2026-06-07
Normalize Mastodon handle into profile URL and reject dangerous schemas.

connorshea:
## Description

AI Disclosure: Changeset generated with Claude Code, reviewed and tested by me.

This isn't actually realistically abusable by an attacker right now because the link is `target="_blank"` which removes all ability to use the `javascript:` schema in modern browsers, but I figure it's best to be safe here just in case.

This normalizes the Mastodon URL on user profiles to prevent any URLs outside the normal `http`/`https` schema. It still allows the `@username@mastodon.host` format as well.

## Screenshots

Previously I was able to set the value to a malicious javascript command, but it will not save that to the database anymore after this PR:

Screenshot 2026-06-07 at 3 27 28 PM

But normal usernames/URLs continue to work fine:

Screenshot 2026-06-07 at 3 28 46 PM

## Testing Steps

- Set a Mastodon username using a URL, ensure the user page renders and the link works.
- Set a Mastodon username using the `@name@host.com` format, ensure the user page renders and the link works correctly.
- Set a Mastodon username as `javascript:alert(document.cookie)`, see that it gets reset to blank string. 

## References

N/A

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -153,7 +153,11 @@ class User < ApplicationRecord
   normalizes :linkedin, with: ->(value) { value.gsub(%r{https?://(?:www\.)?(?:linkedin\.com/in)/}, "") }
 
   normalizes :mastodon, with: ->(value) {
-    return value if value&.match?(URI::DEFAULT_PARSER.make_regexp)
+    return "" if value.blank?
+
+    # Only allow http(s) URLs through verbatim; reject javascript: and other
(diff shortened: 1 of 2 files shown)


marcoroth (approved):
Thank you for the fix @connorshea! 

Also good to know there wasn't really a way for this the be actually dangerous in production 🙌🏼

PR #1806, opened 2026-06-19
Fix latest event sampling on home page

### LINKED ISSUE #1749: Latest events not updating?
on https://www.rubyevents.org/ "Latest events" section

Shouldnt say Blue Ridge Ruby be appearing before RubyConfTH since it was more recent?

Image

sudeeptarlekar:
## Description
Sampling 4 events from last 10 events ordered by descending dates was causing confusion.
This commit fixes the issue and adds the scroll to last 10 events.

## Screenshots
Screenshot 2026-06-20 at 01 01 54

fixes: #1749

### DIFF
--- a/app/controllers/page_controller.rb
+++ b/app/controllers/page_controller.rb
@@ -10,7 +10,7 @@ class PageController < ApplicationController
         events_count: Event.count,
         latest_talk_ids: latest_talks.pluck(:id),
         upcoming_talk_ids: Talk.with_speakers.where(date: Date.today..).order(date: :asc).limit(15).pluck(:id),
-        latest_event_ids: Event.order(date: :desc).limit(10).pluck(:id).sample(4),
+        latest_event_ids: Event.conference.past.limit(10).pluck(:id),
         featured_speaker_ids: User.with_github
           .joins(:talks)
           .where(talks: {date: 12.months.ago..})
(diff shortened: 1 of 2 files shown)


sudeeptarlekar:
@marcoroth review pls. Thanks in advance.

app/controllers/page_controller.rb
```diff
@@ -10,7 +10,7 @@ def home
         events_count: Event.count,
         latest_talk_ids: latest_talks.pluck(:id),
         upcoming_talk_ids: Talk.with_speakers.where(date: Date.today..).order(date: :asc).limit(15).pluck(:id),
-        latest_event_ids: Event.order(date: :desc).limit(10).pluck(:id).sample(4),
+        latest_event_ids: Event.order(date: :desc).limit(10).pluck(:id),
```
marcoroth: Maybe we want something more like:
```suggestion
        latest_event_ids: Event.conference.past.order(end_date: :desc).limit(10).pluck(:id),
```

sudeeptarlekar: Makes sense, I also removed order as `past` is already doing the ordering...

marcoroth (commented):
Thank you @sudeeptarlekar! 🙏🏼

PR #1807, opened 2026-06-20
Validate `github_handle` against GitHub username format

connorshea:
## Description

AI Disclosure: Found with Claude Code, and this fix was Generated with Claude Code. Reviewed and tested by me.

Previously, you could set `github_handle` to a value containing path-injection characters (`/`, `?`, `#`) which would then modify the GitHub API requests that were made by rubyevents in some cases.

I'm not aware of anything harmful you could do with this despite asking Claude to try as much as it could, so I'm opening this PR publicly. The `../` case isn't actually usable as an attack because the URL isn't built in a way that would enable "path traversal" of the URL. And the OAuth token doesn't get enough permissions to really do damage anyway.

Also, underscores aren't allowed in GitHub usernames so some test fixtures were updated (I didn't know this until Claude mentioned it, I assumed it was hallucinating but it turns out it is true).

## Screenshots
N/A

## Testing Steps
- Ensure that usernames can be set still.
- Ensure that GitHub URLs/API requests still work fine

## References
N/A

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -139,6 +139,10 @@ class User < ApplicationRecord
 
   validates :email, format: {with: URI::MailTo::EMAIL_REGEXP}, allow_blank: true
   validates :github_handle, presence: true, uniqueness: true, allow_blank: true
+  GITHUB_HANDLE_PATTERN = /\A[A-Za-z0-9](?:[A-Za-z0-9-]{0,37}[A-Za-z0-9])?\z/
+  validates :github_handle,
+    format: {with: GITHUB_HANDLE_PATTERN, message: "is not a valid GitHub username"},
+    allow_blank: true
   validates :canonical, exclusion: {in: ->(user) { [user] }, message: "can't be itself"}
   validates :distance, comparison: {less_than_or_equal_to: 20_000, greater_than_or_equal_to: 0}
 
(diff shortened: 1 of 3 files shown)


app/models/user.rb
```diff
@@ -139,6 +139,10 @@ class User < ApplicationRecord
 
   validates :email, format: {with: URI::MailTo::EMAIL_REGEXP}, allow_blank: true
   validates :github_handle, presence: true, uniqueness: true, allow_blank: true
+  GITHUB_HANDLE_PATTERN = /\A[A-Za-z0-9](?:[A-Za-z0-9-]{0,37}[A-Za-z0-9])?\z/
```
marcoroth: Can we move this constant up to L70? Thanks!

app/models/user.rb
```diff
@@ -139,6 +139,10 @@ class User < ApplicationRecord
...
+  validates :github_handle,
+    format: {with: GITHUB_HANDLE_PATTERN, message: "is not a valid GitHub username"},
+    allow_blank: true
```
marcoroth: I think we can merge this with the `validates` on L141, no?

connorshea: oops yep

marcoroth (approved):
Great, thanks @connorshea! 🙏🏼

PR #1812, opened 2026-06-21
removed description from 'Things I don't have time for' talk

cyberdevOM:
## Description
Removed the description from the "Things I don't have time for" talk, [Helsinki Ruby Brigade 2026.](https://www.rubyevents.org/talks/things-i-don-t-have-time-for-slow-ruby-and-learning-rust)
Description created a duplicate event name and date.


ChaelCodes (approved):
Thanks so much!! Now we won't have duplicate info! 🙌

PR #1842, opened 2026-06-29
Hide empty profile event sections after filtering

### LINKED ISSUE #1834: Don't show "Future Events" if there are no future events of this kind
Image

minhluuquang:
## Description

Moves the `events-filter` group target from the profile event grids to the Future/Past section wrappers. When a kind filter hides every card in one section, the Stimulus controller now hides that section heading with the grid instead of leaving an empty "Future Events" or "Past Events" heading visible.

## Screenshots

Not included; this is a small markup target change for the existing filter behavior.

## Testing Steps

- `bin/vite build --clear --mode=test`
- `bin/rails test test/controllers/profiles/events_controller_test.rb`
- `bundle exec erb_lint app/views/profiles/_events.html.erb`
- `git diff --check`

## References

- closes #1834

### DIFF
--- a/app/views/profiles/_events.html.erb
+++ b/app/views/profiles/_events.html.erb
@@ -12,10 +12,10 @@
 
     <div class="space-y-8">
         <% if future_events.any? %>
-          <div>
+          <div data-events-filter-target="group">
             <h3 class="text-lg font-semibold mb-4">Future Events</h3>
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
Thank you @minhluuquang! 🙏🏼

PR #1844, opened 2026-06-30
Add Nantes RB June 2026 meetup

francoisedumas:
Add Nantes RB event for June


marcoroth (approved):
Thanks for updating the Nantes.rb events, @francoisedumas! 🙏🏼

Issue #1836, opened 2026-06-28
Cannot Click tags on Newsletter


ChaelCodes:
If you click one of the tags on a [newsletter](https://www.rubyevents.org/announcements/may-2026-newsletter) - you get an error.

<img width="1601" height="686" alt="Image" src="https://github.com/user-attachments/assets/d4e52d36-5fea-4655-b024-2d3ab806136f" />

cato-hga:
👋🏾 I'd love to work on this issue!

ChaelCodes:
> 👋🏾 I'd love to work on this issue!

Absolutely! Thanks!

marcoroth:
Resolved via https://github.com/rubyevents/rubyevents/pull/1847

PR #1847, opened 2026-07-01
Fix tag filtering for non-string tags

yashika279:
## Description
Updated Announcement::Collection#by_tag to safely compare tags of different types.
Converted each tag to a string before performing a case-insensitive comparison using casecmp?.
Prevents NoMethodError when tags contain non-string values such as integers while preserving existing filtering behavior.

## Screenshots
image

## Testing Steps
https://www.rubyevents.org/announcements

## References
https://github.com/rubyevents/rubyevents/issues/1836

### DIFF
--- a/app/models/announcement.rb
+++ b/app/models/announcement.rb
@@ -8,7 +8,7 @@ class Announcement
     end
 
     def by_tag(tag)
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
Thank you @yashika279! 🙏🏼

PR #1852, opened 2026-07-02
Add Languages to profile

### LINKED ISSUE #1826: Add "Languages I speak" to profile
So that we can show/not show talks in these languages in recommendations etc. 

On the home page we show "More Talks in Portuguese" and "More Talks in Japanese" because I marked at least one of each as "watched". In both cases they were in-person as I could listen to the live translation. But I don't speak these languages so RubyEvents shouldn't recommend me more in these languages.

Image

yashika279:
## Description
Added a column in users table: spoken language
Added in strong params 
Added multi select option in profile

## Screenshots
Screenshot from 2026-07-02 15-14-02

Screenshot from 2026-07-02 15-13-44

## Testing Steps
Sign in
goto profile
system should show an option to choose languages

## References
Closes https://github.com/rubyevents/rubyevents/issues/1826

### DIFF
--- a/app/controllers/profiles_controller.rb
+++ b/app/controllers/profiles_controller.rb
@@ -142,7 +142,8 @@ class ProfilesController < ApplicationController
       :speakerdeck,
       :pronouns_type,
       :pronouns,
-      :slug
+      :slug,
+      spoken_languages: []
     )
   end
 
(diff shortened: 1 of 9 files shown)


marcoroth (approved):
Awesome, thank you so much @yashika279! 🙏🏼

PR #1861, opened 2026-07-03
Only show Open CFPs in event header

### LINKED ISSUE #1858: Only show Open CFPs in event header
# Description

When you're reviewing an event, it's helpful to have a call-out if the CFP is open, so we include that in the header. However, events are starting to have multiple CFPs - for regular talks and lightning talks. We display all CFPs under the CFP tab, so let's only show open CFPs in the callout on the event header.

Image

yashika279:
## Description
Added open scope in event header partial
and remove the checks for future, open and closed 

## Screenshots
image

## Testing Steps
goto events
looks for event which had future and closed cfps visible 
with this change it will now show only open cfps

## References
Closes https://github.com/rubyevents/rubyevents/issues/1858

### DIFF
--- a/app/views/events/_header.html.erb
+++ b/app/views/events/_header.html.erb
@@ -75,21 +75,13 @@
     <%= event.description %>
   </p>
 
-  <% event.cfps.each do |cfp| %>
+  <% event.cfps.open.each do |cfp| %>
     <% if event.start_date && event.start_date > Date.today && !event.static_metadata.cancelled? %>
       <div class="mt-6 p-4 bg-gray-100 rounded-lg max-w-[700px]">
         <h3 class="font-bold text-lg mb-2"><%= cfp.name.presence || "Call for Proposals" %></h3>
 
         <p class="text-[#636B74] mb-3">
-          <% if cfp.future? %>
-            The CFP opens on <%= cfp.formatted_open_date %><% if cfp.close_date.present? %> and closes on <%= cfp.formatted_close_date %><% end %>.
-          <% elsif cfp.open? %>
-            The CFP is currently open and closes on <%= cfp.formatted_close_date %>.
-          <% elsif cfp.closed? %>
-            The CFP closed on <%= cfp.formatted_close_date %>.
-          <% else %>
-            Call for Proposals information is available.
-          <% end %>
+          The CFP is currently open and closes on <%= cfp.formatted_close_date %>.
         </p>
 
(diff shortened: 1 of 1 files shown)


app/views/events/_header.html.erb
```diff
@@ -75,21 +75,13 @@
...
-          <% if cfp.future? %>
-            The CFP opens on <%= cfp.formatted_open_date %><% if cfp.close_date.present? %> and closes on <%= cfp.formatted_close_date %><% end %>.
-          <% elsif cfp.open? %>
-            The CFP is currently open and closes on <%= cfp.formatted_close_date %>.
-          <% elsif cfp.closed? %>
-            The CFP closed on <%= cfp.formatted_close_date %>.
-          <% else %>
-            Call for Proposals information is available.
-          <% end %>
+          The CFP is currently open and closes on <%= cfp.formatted_close_date %>.
```
marcoroth: Not all CFPs have close dates, so we might want to keep the `else` condition we had before to account for those too.

ChaelCodes: We might need a different one. Looks like it's using `cfp.open?`. That would just result in "The CFP is currently open and closes on unknown."

Could we check `if cfp.close_date.present?` for the formatted_close_date line, else "The CFP is currently open."?

marcoroth (commented):
Thanks for the update @yashika279! 🙏🏼 

I got one more thing, it looks like the `open` scope doesn't include CFPs that are open-ended (have no `start_date` and `end_date`). It looks like we either need to update the scope on the model or handle these specially:

Image

Thank you! 🙌🏼

app/views/events/_header.html.erb
```diff
@@ -75,28 +75,24 @@
     <%= event.description %>
   </p>
 
-  <% event.cfps.each do |cfp| %>
-    <% if event.start_date && event.start_date > Date.today && !event.static_metadata.cancelled? %>
+  <% if event.start_date.present? && event.start_date.future? && !event.static_metadata.cancelled? %>
```
ChaelCodes: I think we can get rid of this if statement entirely and these cfps will show up on meetups. Meetups don't have start_dates on events.

The open scope looks fine.
https://github.com/rubyevents/rubyevents/blob/main/app/models/cfp.rb#L26

marcoroth (approved):
Thank you @yashika279! 🙏🏼

PR #1885, opened 2026-07-06
Add Google Slides embed support for talk slides

### LINKED ISSUE #1856: Support Google Slides Embed
Currently on the Talks show page we show slides embeds with speakerdeck:

Image

For Google Slides we only have a button that links to an external URL:

Image

But, Google Slides also supports embedding slides, let's build the integration to also handle Google Slides in the same way.

ErnaneJ:
## Description

Google Slides embed support for the talk slides page, matching the existing SpeakerDeck pattern.

`google_slides_embed_url` normalizes any Google Slides URL (edit, view, pub, bare ID) to the `/embed` form. Non-Google Slides URLs return nil, leaving the plain-link fallback untouched.

The new `slides-embed` Stimulus controller shows a daisyUI skeleton while the iframe loads. When the `load` event fires, the skeleton hides and the iframe appears. If `load` never fires within 15 seconds, both hide and the fallback button stays visible.

`slides_external_link` was extracted from the view to avoid repeating the anchor tag across all three branches (SpeakerDeck, Google Slides, fallback).

## Screenshots

Demo for /talks/rubyists-can-make-games-too

## Testing steps

1. Navigate to a talk with a Google Slides URL as slides_url and click the Slides tab.
2. Verify the skeleton shows briefly, then the iframe appears.
3. To test the timeout: temporarily set an invalid Google Slides URL. The skeleton should hide after 15 seconds and leave only the button.

## References

- Closes #1856

### DIFF
--- a/app/helpers/talks_helper.rb
+++ b/app/helpers/talks_helper.rb
@@ -55,6 +55,31 @@ module TalksHelper
     resource["url"]
   end
 
+  def slides_external_link(slides_url)
+    host = URI(slides_url).host
+    link_to "See Slides on #{host}", sanitize_url(slides_url), target: "_blank", class: "btn btn-primary mt-6"
+  rescue URI::InvalidURIError
+    nil
+  end
+
+  def google_slides_embed_url(slides_url)
+    return nil if slides_url.blank?
+
+    uri = URI(slides_url)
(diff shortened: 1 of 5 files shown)


marcoroth (approved):
Awesome, thank you @ErnaneJ! 🙏🏼

Issue #1880, opened 2026-07-05
Checks for Involvements.yml


ChaelCodes:
# Description

When we create an [involvements.yml](https://github.com/rubyevents/rubyevents/blob/main/data/rails-world/rails-world-2025/involvements.yml), we provide either the names or organizations for each tier. However, this means both are optional, but each involvement should have at least one.

We could validate this with a [validator](https://github.com/rubyevents/rubyevents/tree/main/app/models/static/validators) or in the [Schema](https://github.com/rubyevents/rubyevents/tree/main/app/models/static/validators).

If we validate it with a validator, that validator should be added to validate.rake and the generator test.

PR #1916, opened 2026-07-08
Either name or organisations should be present

yashika279:
## Description
Added validation inside validator for name or organisation

## Screenshots
image

## Testing Steps
remove name or organisations from invoment.yml
run  bin/rails validate:involvements

## References
https://github.com/rubyevents/rubyevents/issues/1880

### DIFF
--- a/test/models/static/validators/involvement_role_name_test.rb
+++ b/test/models/static/validators/involvement_role_name_test.rb
@@ -71,6 +71,28 @@ class Static::Validators::InvolvementRoleNameTest < ActiveSupport::TestCase
     end
   end
 
+  test "does not return errors when names present and organisations are not present" do
+    with_temp_involvements([
+      {
+        "name" => "Organizer"
+      }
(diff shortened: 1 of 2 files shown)


test/models/static/validators/involvement_role_name_test.rb
```diff
@@ -46,6 +46,31 @@ class Static::Validators::InvolvementRoleNameTest < ActiveSupport::TestCase
     end
   end
 
+  test "does not return errors when both names and organisations are present" do
+    with_temp_involvements([
+      {
+        "name" => "Organizer",
+        "users" => ["Amanda Perino"],
+        "organisations" => ["Rails Foundation"]
```
marcoroth, line 54: @yashika279 would you mind adding two more tests where only `users` and one where only `organisations` are present

ChaelCodes (approved):
Thank you so much! This looks great! 🙌

PR #1931, opened 2026-07-12
Add dynamic Opengraph image

hschne:
## Description

Marco said he likes my guides OG image, so now I had to do it again for Rubyevents itself I guess? 

image

This adds dynamically generated OG images for the site and it's sub sites. OG images load upcoming/relevant events - slightly adapted from featured to guarantee a sufficient number of them - to display a nice little icon cloud for events. 

Why dynamically generated? Because it's cool and because we can use this to generate profile/organization/event OG images too! See follow ups. 

Details: 
- The approach is adapted from the existing Ferrum based screenshotting used for Wrapped 2025, with some slight changes. 
- OpengraphImage contains the logic. It's will come in handy if we want to use og image generation for other areas of the app.
- Contrary to Wrapped, the og image is generated by a recurring job on a weekly basis. Either run the generation after initial deploy or wait for a bit for the recurring job to run, until an image is present we'll fall back to the current one.  
- Extracted featured query to model and wrote a test for it

### Possible Follow Ups

I want to keep the PR small, so these are deferred for later. Maybe. 
- Refactor Wrapped OG image to align the image generation approaches
- Add OG images for profiles, events, organizations.... 

## Screenshots

image

## Testing Steps

Run ` bin/rails runner 'OpenGraphImage.new.generate!'` to generate the OG image. 

Navigate to http://localhost:3000/og_image to see the magic.


ChaelCodes:
I want to make sure I understand this flow. 

We have a recurring job, that generates the og image by navigating to an html window, screenshotting it, and stores it as a file attached to a model with timestamps. 

Then we query for that model or create one in order to get the og image.

From what I can see, old ones are never destroyed.

app/models/open_graph_image.rb
```diff
@@ -0,0 +1,138 @@
...
+    {cx: 622, cy: 576, size: 62, rotation: -5, opacity: 0.72},
+    {cx: 800, cy: 558, size: 58, rotation: 6, opacity: 0.7},
+    {cx: 980, cy: 574, size: 68, rotation: -7, opacity: 0.78},
+    {cx: 1124, cy: 544, size: 90, rotation: 9, opacity: 0.95}
+  ].freeze
+
+  has_one_attached :image
+
+  def self.instance
+    first || create!
```
ChaelCodes: I'm concerned this will mean we always use the first image generated and not the most recent?

hschne: This is (for now) designed as a singleton record. So yes, the record is always the first - it's the only one. However, the attached image is updated via  `generate!`. Calling `image.attach` leaves the record intact but replaces the old image with the new one.

hschne:
> I want to make sure I understand this flow.
> 
> We have a recurring job, that generates the og image by navigating to an html window, screenshotting it, and stores it as a file attached to a model with timestamps.
> 
> Then we query for that model or create one in order to get the og image.
> 
> From what I can see, old ones are never destroyed.

@ChaelCodes You got that mostly right. The image generation approach is based on what we already did for Wrapped 2025, which itself is based off [this Avo post](https://avohq.io/blog/open-graph-image-generation-rails).  

We never destroy or replace the singleton record, we only update attached images. Old attached images are purged automatically once they are replaced, so the attachment records also don't hang around. 

Want me to change stuff or add comments to make this more transparent? :slightly_smiling_face:

hschne:
@ChaelCodes Finally got around to simplify this, as per our conversation. The new og image is now simply an asset that can be checked in and regenerated via rake task :+1:

PR #1945, opened 2026-07-15
Note which Debian packages also need to be installed

ziggythehamster:
## Description

At RubyConf 2026's hack day, I found that my basically vanilla development environment (WSL2 Debian, RV-installed Ruby) would not successfully run bin/setup. This was because I have the libraries installed but not the headers, so bundle install fails to locate headers.

This PR adds a subsection to tell folks to do this.


ChaelCodes (approved):
Thank you for improving the docs for others!

PR #1946, opened 2026-07-15
Standardize event schema timezones (by always installing tzinfo-data)

### LINKED ISSUE #1893: Stop Generating event_schema.json with timezone changes
# Description

Every time we run bin/lint, we regenerate schemas. The problem is that the event schema uses timzeones, so it regenerates lib/schemas/event_schema.json with changes. This task involves investigating _why_ we've got discrepancies between timezones and coming up with some solutions that would give us a fixed set of timezones.

nevinera:
## Description
Always bundle tzinfo-data, instead of skipping on platforms that provide that information themselves.

This will prevent the events schema from getting diffs from bin/lint, when run on systems that have subtly different lists of timezones.

## Testing Steps
Run `bin/lint`, confirm that no diff is produced (especially, do this on a computer that was previously producing such a diff. @ChaelCodes can confirm that, I think?)

## References
closes #1893

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,8 +39,8 @@ gem "turbo-rails"
 # Use Active Model has_secure_password [https://guides.rubyonrails.org/active_model_basics.html#securepassword]
 gem "bcrypt", "~> 3.1.7"
 
-# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
-gem "tzinfo-data", platforms: %i[windows jruby]
+# Bundle tzinfo-data on all platforms, so TZInfo::Timezone.all_identifiers is deterministic
+gem "tzinfo-data"
 
 # Reduces boot times through caching; required in config/boot.rb
 gem "bootsnap", require: false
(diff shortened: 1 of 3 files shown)


nevinera:
Worth noting - there is a _more_ surgical approach available; if we wanted to just export that list to a yaml file (in `config/time_zone_names.yml` possibly), we could then just _read that list_ from [event_schema.rb](https://github.com/rubyevents/rubyevents/blob/main/app/schemas/event_schema.rb#L33) and be consistent without actually touching the app-wide configuration or Gemfile. I don't think this is a particularly risky change, but _that_ is even safer :-)

ChaelCodes (approved):
IT WORKS!! 🎉 ✨

ChaelCodes:
> Worth noting - there is a _more_ surgical approach available; if we wanted to just export that list to a yaml file (in `config/time_zone_names.yml` possibly), we could then just _read that list_ from [event_schema.rb](https://github.com/rubyevents/rubyevents/blob/main/app/schemas/event_schema.rb#L33) and be consistent without actually touching the app-wide configuration or Gemfile. I don't think this is a particularly risky change, but _that_ is even safer :-)

We could, but then we'd have to update it when timezones change, and it's a question of how we'd discover that timezones have changed. And really, that feels like the same solution - a static list of timezones - but we'd be storing it instead of the gem. (And I know we'd just regenerate it from the gem when things update.)

PR #1951, opened 2026-07-15
Update William Frey's GitHub

andycandrea:
## Description
I noticed that the picture and GitHub for Billy/William Frey, who I gave a talk with a few years back, was the wrong person. Since there's only that one talk attributed to him, I think it should be a safe change especially since the current GitHub account doesn't have any past Ruby activity.

## Screenshots

Screenshot 2026-07-15 at 3 02 03 PM

Screenshot 2026-07-15 at 3 04 21 PM

## Testing Steps

I verified that the GitHub link and profile picture were correct on the speaker show page for `will-frey` and that it looked correct on the show page for the talk we gave together, too.

### DIFF
--- a/data/speakers.yml
+++ b/data/speakers.yml
@@ -17245,7 +17245,7 @@
   github: ""
   slug: "william-bereza"
 - name: "William Frey"
-  github: "will-frey"
+  github: "bfrey08"
   slug: "william-frey"
 - name: "William Horton"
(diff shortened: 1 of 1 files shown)


ChaelCodes (approved):
Thank you for fixing this and sorry for the trouble! 🙏  When the seeds run, it'll update his account.

PR #1952, opened 2026-07-15
Compact speaker profile header for mobile viewports (RubyConf Hack Day)

stancheta:
## Description

Below `md` (< 768px), the speaker profile page trades the centered hero and full-width card grid for a compact 64px-avatar header and horizontal talk rows so ~5 talks fit above the fold instead of 0. Desktop is unchanged except the tab bar, which now uses the existing `tabs-boxed` treatment (same as the events kind filter) and hugs its content.

View-layer only — no controller, model, migration, JS, or system-test changes. Reuses existing partials (`talks/_card_horizontal`, `topics/_badge_list`, `profiles/_socials`) via additive locals so the change doesn't touch other callers.

Little bit of scope increase by using the grey background tab box instead of pills.

### Old
Screenshot 2026-07-15 at 5 39 03 PM

### New
Screenshot 2026-07-15 at 5 41 49 PM

## Testing Steps

- `/profiles/:slug` at 390px shows the compact header + horizontal talk list; no horizontal scroll.
- `/profiles/:slug` at ≥ 768px is visually identical to `main`.
- `/talks/:slug` topic tags still render with the ghost `+N more` chip (opt-in kept the outlined variant to the profile page).
- `bin/rails test`, `bundle exec standardrb`, `bundle exec erb_lint --lint-all` pass.


app/views/talks/_card_horizontal.html.erb
```diff
@@ -23,8 +27,17 @@
...
+  <div class="flex aspect-video shrink-0 relative <%= thumb_width_class %>">
+    <% if show_upcoming_placeholder %>
+      <div class="w-full h-full aspect-video rounded flex items-center justify-center bg-[repeating-linear-gradient(45deg,#e7e5e4_0_8px,#efedeb_8px_16px)]">
+        <span class="font-mono text-[10px] text-gray-500 bg-white rounded px-1.5 py-0.5">upcoming</span>
+      </div>
+    <% else %>
+      <%= image_tag talk.thumbnail_sm, srcset: ["#{talk.thumbnail_sm} 2x"], id: dom_id(talk), class: "w-full h-auto aspect-video border rounded #{talk.thumbnail_classes.presence || "object-cover"}", loading: :lazy %>
+      <% if show_duration_chip && talk.duration_in_seconds.to_i > 0 %>
+        <span class="absolute right-1 bottom-1 bg-black/75 text-white text-[10px] font-semibold px-1.5 py-px rounded"><%= talk.formatted_duration %></span>
+      <% end %>
```
ChaelCodes: Image

Uh oh!

stancheta: Ah, beans! Nice catch! That definitely shouldn't be talk.formatted_duration. I replaced it with seconds_to_formatted_duration(talk.duration_in_seconds), which is the nil-safe helper that's use elsewhere. While I was in there, I also took the liberty of aligning a few of these changes with the existing patterns in RubyEvents, so it should be a bit more consistent.

ChaelCodes (approved):
This looks great!! Thank you!

PR #1953, opened 2026-07-16
Respect PORT environment variable for development environment assets and mailer URLs

kcdragon:
## Description

The Rails server itself does respect the PORT environment variable but the assets don't load.

Also, I deleted something that seemed like a duplicate configuration. Let me know if there's some Rails magic going on and its not a duplicate.

## Screenshots

Before

Screenshot 2026-07-15 at 5 43 22 PM

## Testing Steps

Run this on the main branch to reproduce

```shell
PORT=2999 ./bin/dev
```

## References

- closes #

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -35,11 +35,10 @@ Rails.application.configure do
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
 
-  # Enable serving of images with full URLs
-  config.asset_host = "http://localhost:3000"
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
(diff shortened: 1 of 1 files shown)


config/environments/development.rb
```diff
@@ -35,11 +35,10 @@
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
 
-  # Enable serving of images with full URLs
-  config.asset_host = "http://localhost:3000"
```
kcdragon, line 39: I think this can be removed since it's immediately overwritten below but let me know if there's some Rails magic going on and it should stay.

ChaelCodes (approved):
Thanks! Interesting one.

PR #1955, opened 2026-07-16
Generate RubyLLM tools from Rails generators

andycandrea:
## Description
Generates RubyLLM tools based on each of the generators that exists in `lib/generators/` to enable LLMs to make easy and deterministic data updates that will hopefully also cost fewer tokens than having it do so more manually.

Left as a draft until the TODO items are resolved.

## Screenshots

Screenshot 2026-07-15 at 10 12 43 PM

Screenshot 2026-07-15 at 10 13 04 PM

## Testing Steps

- [x] Ran the event, sponsor and CFP tools by executing the corresponding tools manually in a Rails console
- [x] Ran `bin/mcp_server` and verified that 1) it didn't raise an error and 2) logged the additional tool classes that we generated
- [x] Ran `npx @modelcontextprotocol/inspector bin/mcp_server`, verified that all new tools were listed, had it run the event generator tool and verified the output

### DIFF
--- a/bin/mcp_server
+++ b/bin/mcp_server
@@ -5,6 +5,7 @@
 # Run with: bin/mcp_server
 
 require_relative "../config/environment"
+require_relative "../lib/generators/tool_generation"
 require "json"
 
 class MCPServer
@@ -36,11 +37,15 @@ class MCPServer
   private
 
   def discover_tools
-    Dir[Rails.root.join("app/tools/**/*_tool.rb")].each_with_object({}) do |file, tools|
+    discovered_tools = Dir[Rails.root.join("app/tools/**/*_tool.rb")].each_with_object({}) do |file, tools|
       require file
       class_name = File.basename(file, ".rb").camelize
       tool_class = class_name.constantize
       tools[tool_name(class_name)] = tool_class
+    end
+
+    ToolGeneration.generate_tools.each.with_object(discovered_tools) do |klass, tools|
+      tools[klass.name] = klass
     end
   end
 
--- a/bin/mcp_server
+++ b/bin/mcp_server
@@ -44,8 +44,13 @@ class MCPServer
       tools[tool_name(class_name)] = tool_class
     end
 
-    ToolGeneration.generate_tools.each.with_object(discovered_tools) do |klass, tools|
-      tools[klass.name] = klass
+    # Load all the generators, create a tool class associated with each one and
+    # then add them to the tools hash
+    Rails::Generators.sorted_groups
+    generators = Generators::EventBase.descendants
+    tool_classes = ToolGeneration.generate_tools(generators)
+    tool_classes.each.with_object(discovered_tools) do |klass, tools|
+      tools[tool_name(klass.name)] = klass
     end
   end
 
--- a/lib/generators/tool_generation.rb
+++ b/lib/generators/tool_generation.rb
@@ -1,22 +1,24 @@
 require "rails/generators"
 
 class ToolGeneration
-  # Generates RubyLLM tool classes for each Rails generator defined in
-  # lib/generators.
+  # Generates RubyLLM tool classes for each provided generator
   #
+  # @param generators [Array<Generators::EventBase>]
   # @return [Array<RubyLLM::Tool>]
-  def self.generate_tools
-    # Load all the generators
-    Rails::Generators.sorted_groups
-
-    Generators::EventBase.descendants.map do |klass|
+  def self.generate_tools(generators)
+    generators.map do |klass|
       # Grab the generator's options that we've defined manually so that they
       # can be provided as params to the new tool class.
       options = klass.class_options.select do |option_key, option|
         option.group == "Fields"
       end
 
+      generator_type = klass.name.gsub("Generator", "").downcase
+      description = tool_description(klass, generator_type)
+
       tool = Class.new(RubyLLM::Tool) do
+        desc description
+
         options.each do |option_key, option|
           # Document each param for the tool using what is defined on the
           # corresponding generator option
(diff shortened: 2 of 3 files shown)


bin/mcp_server
```diff
@@ -36,12 +37,16 @@ class MCPServer
...
-    Dir[Rails.root.join("app/tools/**/*_tool.rb")].each_with_object({}) do |file, tools|
+    discovered_tools = Dir[Rails.root.join("app/tools/**/*_tool.rb")].each_with_object({}) do |file, tools|
       require file
       class_name = File.basename(file, ".rb").camelize
       tool_class = class_name.constantize
       tools[tool_name(class_name)] = tool_class
     end
+
+    ToolGeneration.generate_tools.each.with_object(discovered_tools) do |klass, tools|
+      tools[klass.name] = klass
```
andycandrea: Does the format of the key matter here? If so, I think we'd want this to be something more like:

```rb
tools[tool_name(klass.name)] = klass
```

For context, the tools from `app/tools/` look like

```
"cfp_create" => CFPCreateTool
```

whereas the genned ones look like:

```
"GenerateCFPTool" => GenerateCFPTool
```

lib/generators/tool_generation.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      generator_type = klass.name.gsub("Generator", "").downcase
+
+      # If an argument's value is provided, ensure that it gets passed to the
+      # generator in the correct format. If it is not provided, discard it to
+      # ensure fallback values in the generator code are used properly.
+      argument_handling_code = options.map do |option_key, option|
+        <<~METHOD
+          if #{option_key}
+            cli_options << ["#{option.switch_name}", *Array.wrap(#{option_key})]
```
andycandrea, line 40: Since our initial version, I took out the `if #{option_key}.is_a?(Array)` branch in favor of the `*Array.wrap` to treat both array and non-array values the same.

lib/generators/tool_generation.rb
```diff
@@ -0,0 +1,71 @@
...
+class ToolGeneration
+  # Generates RubyLLM tool classes for each Rails generator defined in
+  # lib/generators.
+  #
+  # @return [Array<RubyLLM::Tool>]
+  def self.generate_tools
+    # Load all the generators
+    Rails::Generators.sorted_groups
+
+    Generators::EventBase.descendants.map do |klass|
```
andycandrea: RE: the testing conversation: if we extracted out this to a `generators` arg, we could more easily test this logic by passing in a single, contrived generator and then simply testing that the return value includes a class that's set up the way we'd expect.

ChaelCodes: I love that idea!

bin/mcp_server
ChaelCodes: From what I see it doesn't because the name is just used as a key in handle_request, but robots like consistency, and I don't see an issue with using tool_name.

lib/generators/tool_generation.rb
```diff
@@ -0,0 +1,71 @@
...
+      # Grab the generator's options that we've defined manually so that they
+      # can be provided as params to the new tool class.
+      options = klass.class_options.select do |option_key, option|
+        option.group == "Fields"
+      end
+
+      tool = Class.new(RubyLLM::Tool) do
```
ChaelCodes, line 19: We need to include a description for the MCP tool here. That's how the robots know if they should grab this tool. 🤔 

`desc "Generate or update a cfp for an event."`
[Ruby LLM example](https://rubyllm.com/tools/#creating-a-tool)

However, right now, I'm not using a simple description for the generators. I'm using a USAGE document which is way too long.

I'm thinking that generators probably support both? And I can add a short description to each of the generators to check.

~~Meanwhile, Thor's examples here have me thinking I'm not using the documentation/examples feature correctly or to its full effect.~~ Misunderstood what we were doing with example.

https://github.com/rails/thor/wiki/Getting-Started

ChaelCodes: That was harder than expected. #1966 Had to build a hacky solution that won't work for the generic Rails generators. Maybe need to explore upstream fixes?

andycandrea:
@ChaelCodes 

* Added some tests for the tool generation (though I have almost zero MiniTest experience, so feel free to be as brutal as you want when looking those over if there's anything weird or unpolished there)
* Added the `desc` using your `TOOL_DESC` constants with a super simple fallback if that constant isn't defined
* Standardized the tool keys in `bin/mcp_server`
* Did a little more testing

I left it as a separate commit to make reviewing the changes a little easier. Once you think it looks good, I'll squash and mark it as fully ready for review.

ChaelCodes:
@andycandrea it looks good!! ✨ 

I want to test it out using opencode tomorrow, but I see no issues with this. 

We normally squash-merge, so the commits aren't a big deal either. 

I want to merge as-is and start using it, but we might want to think about second pass items.

- arguments (in addition to options) 
- capturing and handling errors/success messages
- a more permanent solution to description

andycandrea:
@ChaelCodes Awesome! I marked it as ready for review, so it should be mergeable whenever you'd like.

On the subject of error messages, one thing that Claude flagged for me was the `exit` in `EventBase`. Since that raises an error that's not a `StandardError`, I think we run the risk of killing the entire MCP server process if we hit that with a tool call.

ChaelCodes:
Screenshot 2026-07-27 at 1 08 50 PM

It's working and errors aren't crashing it! But these are all parameter errors.

ChaelCodes:
It cheerfully generated this PR - https://github.com/rubyevents/rubyevents/pull/2005

ChaelCodes:
Screenshot 2026-07-27 at 9 36 39 PM

It definitely broke on a bad slug.

ChaelCodes:
I tried ask instead of error and now it's stuck. :/

Screenshot 2026-07-27 at 9 51 37 PM

ChaelCodes:
Screenshot 2026-07-27 at 10 01 06 PM

Well, it fails when it gets to the template.

By the way, my first attempt at raising an InvalidAgumentException resulting in the mcp tool server crashing on load.

PR #1960, opened 2026-07-18
Add Triangle.rb meetup and past videos

andycandrea:
## Description

There used to be two Ruby meetups in the Research Triangle region: Triangle.rb/Triangle Ruby Brigade and Thirsty.rb. I couldn't find much existing reference to Thirsty.rb aside from [this single Reddit post](https://www.reddit.com/r/triangle/comments/1u0j8v/thirstyrb_ruby_developer_meetup_at_crank_arm/) that implies their meetups were published on EventBrite, but a lot of the Triangle.rb stuff is still online.

For now, I just went ahead and added the meetup and the talks whose recordings I could find, but [their Meetup page](https://www.meetup.com/raleighrb/events/?type=past) is still live with all of their old events. I bet we could point Cursor/Claude that way to harvest a bunch of the non-published stuff (maybe you all have already done something similar in the past?). From a quick skim of the older events, there also seem to be lots of references to talks that were given to practice in advance of larger conferences like RubyConf ([example from 2004](https://www.meetup.com/raleighrb/events/3000748/?eventOrigin=group_events_list)) that might help us fill in some other events if we happen to lack data for them.

I based anything I added by hand on what was in the schemas and the ChicagoRuby files.

## Screenshots

One odd thing here: I notice that it says 10 talks rather than 5 which would be what I'd expect. Not sure if something in my added files is screwy or that's expected.

localhost_3000_events_triangle-rb-meetup

## Testing Steps

- [x] New meetup shows up in search
- [x] Clicking through to each new talk shows the link to the correct speaker and a working video
- [x] Clicking through to each new speaker shows their name and, if applicable, socials and profile pic


ChaelCodes (approved):
This looks perfect! You may be interested in this script for [automatically generating event assets.](https://github.com/rubyevents/rubyevents/blob/main/docs/ADDING_VISUAL_ASSETS.md#automated-asset-generation)

It takes a logo and some colors and generates the full suite of feature, card, and banner images.

PR #1971, opened 2026-07-21
Add alt text to image tags

andycandrea:
## Description
On hack day at RubyConf, I ran a Lighthouse a11y report and noticed that most of what it flagged were missing `alt`s on the images, so I went ahead and added any missing ones that I could find. Finally making the PR now.

## Screenshots

Screenshot showing the `alt` text audit passing for my local copy's home page
Screenshot 2026-07-21 at 12 27 02 AM

Screenshot showing the `alt` text audit failing for prod's home page
Screenshot 2026-07-21 at 12 27 33 AM

## Testing Steps

- [x] Lighthouse score increased from 88 to 96 for the home page (similar increases on the other pages I spot-checked, though a few pages have lower scores than the 96)
- [x] After clicking around through as much of the app as I could think of, no new errors surfaced in the server logs or console (I also had Claude do a sanity check for me in case there were any pages I missed)

## References
N/A

### DIFF
--- a/app/components/ui/avatar_component.html.erb
+++ b/app/components/ui/avatar_component.html.erb
@@ -6,7 +6,7 @@
       <% else %>
         <span class="<%= text_size %> <%= "hidden" if show_custom_avatar? %>"><%= initials %></span>
         <% if show_custom_avatar? %>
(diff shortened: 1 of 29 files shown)


ChaelCodes (approved):
Thank you for adding all of these! ✨

PR #1974, opened 2026-07-21
Improve event navigation

hschne:
## Description

This is an attempt at simplifying (?) the navigation across event pages. The various events pages are a bit of a tangle, so to keep the PRs small I'm just making tiny increments. 

Previously,  different event pages had fairly different navigation patterns. Some pills, some links, all changing across each view. Now this has been aligned with one persistent sub-navigation and a context-dependent pill components. 

Some things I'd like to mention: 
- I had to make some changes to the calendar page especially, so things match up.
- Mobile doesn't look particularly great. 

Follow ups: 
- I'm considering removing the list view. It's nearly identical to the preview view, and... does anyone even use that? Would simplify navigation. 
- Meetups are in a weird spot right now. IMO the upcoming events list should contain upcoming conferences and meetups, with a filter to toggle between them (like we have on the archive page now) - but I got a feeling that will be a fair bit of work, because meetups do work a bit differently then conferences. 

### After
| Upcoming   | Past   | Calendar   | Archive   | Meetups   |
| ---------- | ---------- | ---------- | ---------- | ---------- |
| meetups        | archive        | calendar        | past        | upcoming        |

### Before

| Upcoming   | Past   | Calendar   | Archive   | Meetups   |
| ---------- | ---------- | ---------- | ---------- | ---------- |
| meetups  |  archive   | calendar    | past   |  upcoming   |

## Testing Steps

Give it a spin on [/events](http://localhost:3000/events/)


ChaelCodes:
I prefer list to preview - especially for mobile, and preview causes UI issues in low bandwidth environments. The whole point of the fog is to have a fixed size for scroll with the preview image on the side, and in your screenshots the preview is cropped out.

ChaelCodes:
Just saw the screenshots and I am in love!! <3 They're gorgeous! I'm out of review time today, but tonight or tomorrow, I want to test.

app/views/events/years/_yearly_events.html.erb
```diff
@@ -1,21 +0,0 @@
-<div class="mt-8 md:mt-12">
-  <div class="flex flex-col md:flex-row md:items-baseline justify-between gap-1 md:gap-0 mb-4 md:mb-6">
-    <h2 class="text-xl md:text-2xl font-bold text-gray-900">Sometime in <%= year %></h2>
-    <span class="text-xs md:text-sm text-gray-400"><%= events.count %> events with unconfirmed dates</span>
```
ChaelCodes, line 4: Image
Image

It looks like we lost this partial entirely. I think it'd be good to add it back.

app/views/events/years/index.html.erb
```diff
@@ -1,17 +1,23 @@
...
+    <p class="text-sm text-gray-500">
+      <%= pluralize(@monthly_events.count, "event") %>
+
+      <% if @monthly_events.any? %>
+        across <%= pluralize(@events_by_month.keys.count, "month") %>
+      <% end %>
+    </p>
+
+    <span aria-hidden="true" class="text-gray-300">·</span>
+    <%= render "shared/subscribe_to_calendar", url: events_url(format: :ics), modal_id: "modal-middle" %>
```
ChaelCodes, line 25: It moves subscribe to the footer, but maybe that's okay?

app/views/events/years/_yearly_events.html.erb
hschne: I removed it because I felt it unnecessarily complicated the UI. When I open the calendar, I want to see which events happen when, and that partial doesn't help. 

It's a matter of taste. I think it should go, but if you think otherwise I'll add it back of course.

marcoroth: I wonder why you think this should go? I felt like this is actually useful if you want add an event that you know is going to happen, but we don't quite know any of the dates/details yet.

hschne: Nothing stops you from adding such an event as organizer. I'm talking from the  attendee perspective.

Like I said, I look at a calendar to see what is happening when. Seeing that an event happens sometimes this year adds minimal value. Not worth the UI complication. Having a simpler UI outweighs the marginal value added by this display. 

Look, that's just my opinion. It's no big deal, I'll add it back and that's that 😄

app/views/events/years/index.html.erb
hschne: I like it more, having text + navigation on top is a bit much 🤷

app/views/events/years/_yearly_events.html.erb
hschne: Added it back now :+1:

marcoroth (approved):
Awesome, great improvement! Thank you @hschne! 🎉

app/views/events/years/_yearly_events.html.erb
marcoroth: Thank you!

PR #2010, opened 2026-07-28
Polish the mobile speaker profile layout

stancheta:
Follow-up to #1952, which landed the compact mobile header. This is the polish pass on top of it, from a design handoff after reviewing main on a real device.

## What changed

- **Spacing** is one 16px gutter throughout. The header, topics, tabs and talks list were each adding their own padding on top of the container's, so they sat at 32px while every other profile tab sat at 16px. The large gap under the header was an empty actions block still holding its padding.
- **Thumbnails** keep their current 132px width but are now locked to 16:9, so the talk list has a straight left edge. They were previously free to keep their own aspect ratio, which started each row's text at a different point.
- **Upcoming talks** show their real thumbnail instead of a striped placeholder, the same as the desktop cards.
- **Topics** show 4 on mobile behind a plain "+N more" chip. Eight pills took three rows, and the old outlined chip was the only outlined control on the page, so it read as the main action. Desktop still shows 8.
- **Badges** stay on the name line instead of wrapping underneath it.
- **Tabs** show which one you're on. Nothing was marked active before.

## Notes for review

`active_link_to` is shared. The new `active:` keyword defaults to `nil`, so all ~25 existing call sites keep their current `current_page?` behaviour. The profile root renders Talks inline without navigating to the tab's own URL, which is why `current_page?` alone matched nothing there. The helper also now sets `aria-selected` on `role="tab"` links, which previously exposed no selected state to assistive tech at all — that reaches the events tab bar too, and is the one part of this PR that isn't purely cosmetic.

`_card_horizontal.html.erb` and `_badge_list.html.erb` are shared with the talk detail page. Changes are behind the existing `variant: :profile` flag or caller-passed locals; the talk-detail sidebar and its 8-topic badge list are unchanged.

The thumbnail fix needed care: the image carried `h-auto` alongside `aspect-video`, which let the intrinsic ratio win. `object-cover` is now applied only when `thumbnail_classes` doesn't already set an `object-fit`, so position-only values like `object-bottom` get a fit while the deliberate `object-contain bg-[#FEE3E5]` on Balkan Ruby 2026 is preserved.

The topic limit is a server-side decision and can't be scoped by a media query, so that list renders once per breakpoint.

Desktop is unchanged — header, topic count and tab metrics were measured against `main` and match exactly.

## Not included

The tab bar uses `role="tab"` on what are really navigation links, without `role="tabpanel"` or arrow-key handling. I added `aria-selected` to close the immediate gap, but the fuller fix is dropping `role="tab"` — that touches the desktop tab bar and `events/_navigation.html.erb`, so it felt out of scope here. Happy to follow up if you'd like it.

##### Old

CleanShot 2026-07-28 at 11 38 13

##### New

CleanShot 2026-07-28 at 16 00 10


ChaelCodes:
Oh no! CI errors! I'll re-review once those are resolved.

marcoroth:
CI was blocked on https://github.com/marcoroth/herb/pull/1971, which I now resolved by using the Herb Linter from main 🙌🏼

marcoroth (approved):
This is good to go, thanks @stancheta! 🙏🏼

PR #2016, opened 2026-08-02
Update LRUG with May 2026 & July 2026 videos, and add upcoming August 2026 meeting

h-lame:
## Description

Updates to LRUG `videos.yml` - this is the result of running `bin/update_lrug_meetups` - but broken across a few commits to clarify the changes (and make it easy to accept some of them and reject others).

- add videos for May 2026
- add videos for July 2026
- add upcoming August 2026 meeting
- correct formatting etc issues introduced by running most recent Yerba over the file

Notable is that this undoes the addition of `published_at` #1849 - the data isn't present in the lrug.org source export, so any run of `bin/update_lrug_meetups` will remove it if the data is re-introduced.  We can put the same "fake" date into the source if it's better to have a `published_at` though to avoid this churn (at least until we do some data mining on our own repo / server to get better "when was the video published" information).


ChaelCodes:
Looks like yerba failed in CI. `yerba apply` should fix it. Normally that's whitespace issues.

h-lame:
> Looks like yerba failed in CI. `yerba apply` should fix it. Normally that's whitespace issues.

Thanks @ChaelCodes I could have sworn it was clear when I pushed it, but a rebase and repush seems to have fixed it

ChaelCodes:
I decided to add back the published_at. I don't have a better process than manually with git restore.

h-lame:
> I decided to add back the published_at. I don't have a better process than manually with git restore.

Great thank you!  We'll do a thing over in https://github.com/lrug/lrug.org to put a `published_at` back in and hopefully avoid this churn next time.  Chances are I'll take the opportunity to mine our git logs etc to get "real" dates so expect a PR sometime soon that updates these.

marcoroth:
I opened https://github.com/lrug/lrug.org/pull/477 to help with the churn!

Though, we might need to think about a better format for these data sync in the future.

h-lame:
> Though, we might need to think about a better format for these data sync in the future.

Yeah, we have meeting data going back to 2006 that we could import, and while we don't have videos for most anymore, that still means one huge `videos.yml` for the whole of LRUG and it would be tens of 1000s of lines long (it's already 4.1k just for 2020 onwards).  We could model LRUG as an event with a series per year - that's less than a 1000 liner per file.  But I suspect I'm rushing to a solution with what is available, you may already be thinking about modelling meet-ups differently from conferences.

PR #2017, opened 2026-08-03
Update data for Ziggy Gable

ziggythehamster:
## Description

* Stop using my deadname where possible (if someone at Ruby Central and/or Confreaks want to fix this further upstream too, please do!)
* Prefer `ziggythehamster` as my slug primarily, but retain the other slugs I have had.

## Questions

* Alias names are required, but I only have conditions resulting in duplicates, so I did something that seems to pass validation but is probably wrong. Please let me know how this should have worked :)
* A few folks annotated the SF Ruby videos with Ziggy the Hamster, though I was me and not the cartoon character representing me, so those could be Ziggy Gable instead if anyone has an opinion in favor of changing them.
* Is there an established procedure for this that I'm missing? Should there be one?


marcoroth (approved):
Thanks for this pull request @ziggythehamster! 🙏🏼 

I will make sure this properly propagates on production, thank you! 🙌🏼

PR #2020, opened 2026-08-04
Map Generator Type to MCP Tool type

ChaelCodes:
## Description

Some models would just fix this for me, but we were passing the wrong type in the MCP Tool schema which was causing provider errors.

(particularly "numeric" from Schedule)

## References

- closes #

### DIFF
--- a/bin/mcp_server
+++ b/bin/mcp_server
@@ -141,7 +141,7 @@ class MCPServer
     if tool_class.respond_to?(:parameters)
       tool_class.parameters.each do |name, param|
         schema[:properties][name.to_s] = {
-          type: param.type || "string",
+          type: map_type(param.type) || "string",
           description: param.description
         }
         schema[:required] << name.to_s if param.required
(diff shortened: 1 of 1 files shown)


bin/mcp_server
```diff
@@ -151,6 +151,17 @@ class MCPServer
     schema
   end
 
+  def map_type(type)
+    case type.to_s
+    when "numeric", "number", "float", "double" then "number"
```
marcoroth, line 156: I wonder if this is a "bug" in RubyLLM::Schema?

ChaelCodes: What do you mean?

This is building from the generator options, not the schema.

ChaelCodes: I think we're still using param here.

ChaelCodes: Oh, wait. I'm wrong.

ChaelCodes:
I need this or I can't use Big Pickle. Keep breaking.

PR #2028, opened 2026-08-05
Install yarn during bin/setup

alimi:
## Description

If you're installing Node for the first time, you might not have yarn installed. Enabling corepack tells Node to install the package manager as specified by package.json. This makes local setup a little nicer.

See https://yarnpkg.com/corepack


marcoroth (approved):
Great catch, thanks for this @alimi! 🙏🏼

PR #2029, opened 2026-08-06
Fix tab overflow in tabs on /cfp

dennispaagman:
## Description
The selected tab overflows on /cfp. This is a regression introduced by commit 0f3ae467e (“Adopt Herb Formatter and Tailwind Class Sorter”, #2013), which rewrote the line from `class="tabs tabs-boxed"` to `class="tab tabs-boxed"` (singular `tab` vs `tabs`).

## Screenshots
CleanShot 2026-08-06 at 11 24 21@2x

### DIFF
--- a/app/views/cfp/index.html.erb
+++ b/app/views/cfp/index.html.erb
@@ -5,7 +5,7 @@
     </h1>
 
     <div class="mt-4 flex items-center gap-4 md:mt-0">
-      <div class="tab tabs-boxed">
+      <div class="tabs-boxed tabs">
         <%= link_to "All", cfp_index_path(kind: "all"), class: "tab #{"tab-active" if params[:kind].blank? || params[:kind] == "all"}" %>
         <%= link_to "Conferences", cfp_index_path(kind: "conference"), class: "tab #{"tab-active" if params[:kind] == "conference"}" %>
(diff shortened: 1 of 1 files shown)


marcoroth (approved):
Thanks for catching @dennispaagman, this is definitely also worth an issue for the Herb Formatter. Gonna have to look into that. Thank you! 🙏🏼

dennispaagman:
Did the formatter make the mistake? That's interesting for sure.

Btw I really love that it can sort tailwind classes now 🔥 learned that from this fix ☺️

marcoroth:
> Did the formatter make the mistake? That's interesting for sure.

I'm not 100% sure, but I think I only let the formatter touch the files. But maybe this slipped in another way.

marcoroth:
Okay good, looks like it wasn't the Herb Formatter itself, as it doesn't reproduce. That is relieving 😅  

I checked out rubyevents at 8bfedd28f (the commit before 0f3ae46), installed the exact formatter build that commit pinned (`pkg.pr.new/marcoroth/herb/@herb-tools/formatter@c5ca11c`) plus `tailwindcss@3.4.17` and `daisyui@4`, copied in the `.herb.yml` from that commit, and ran `herb-format app`.

Case closed, thanks for catching it! 🙏🏼

dennispaagman:
Could've been Claude hallucinating what introduced it!

PR #2031, opened 2026-08-06
Fix non-existing brand colors and streamline alphabet filters

dennispaagman:
While building something else I found that the alphabet filters use the non-existing `bg-brand-lighter` color, these have been dropped in 80de288 which dropped the legacy `extend.colors` block from tailwind.config.js without updating any of the view, so none of these pages had a visible selected state.

Also noticed that the alphabet filter is implemented a bit different in multiple places, this streamlines it into one partial that is re-used.

Leading to:

- Selected state works again.
- "All" is always shown and as the first option, consistent in all places.

I fixed the usage of the brand color in two other places as well. The blocks for missing sponsors has become a bit nicer with a transition and hover color for the whole block instead of just the title.

## Screenshots
Before:
CleanShot 2026-08-06 at 13 18 10@2x

After:
CleanShot 2026-08-06 at 13 17 49@2x

CleanShot 2026-08-06 at 13 16 14@2x

### DIFF
--- a/app/views/events/archive/index.html.erb
+++ b/app/views/events/archive/index.html.erb
@@ -30,7 +30,7 @@
     </div>
   </div>
 
-  <%= render "shared/alphabet_filter", extra_params: {kind: params[:kind]}, anchor_ids: true %>
+  <%= render "shared/alphabet_filter", extra_params: {kind: params[:kind]} %>
 
   <% if params[:letter].present? %>
     <div class="mb-12 font-medium">
--- a/app/views/events/archive/index.html.erb
+++ b/app/views/events/archive/index.html.erb
@@ -30,17 +30,7 @@
     </div>
   </div>
 
-  <div class="flex w-full flex-wrap justify-between gap-1 py-8 hotwire-native:hidden">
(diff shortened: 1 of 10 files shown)


app/views/events/archive/index.html.erb
```diff
@@ -30,17 +30,7 @@
...
-      all
-    <% end %>
-
-    <% ("a".."z").each do |letter| %>
-      <%= link_to archive_events_path(letter: letter, kind: params[:kind]), id: letter, class: class_names("flex items-center justify-center w-10 text-gray-500 rounded hover:bg-brand-lighter border", "bg-brand-lighter": letter == params[:letter]) do %>
-        <%= letter.upcase %>
-      <% end %>
-    <% end %>
-  </div>
+  <%= render "shared/alphabet_filter", extra_params: {kind: params[:kind]}, anchor_ids: true %>
```
dennispaagman: maybe we can drop the anchor_ids all together as well? Only archive/index uses them and i'm not sure what the use case is.

marcoroth: I thought it might be to link to the start of "where events with O begin", but it doesn't like that's the way it works.

dennispaagman: They were used in a test, I've changed the test to be more explicit on what to click and removed them from the partial.

marcoroth (approved):
Thank you @dennispaagman! 🙏🏼

PR #2032, opened 2026-08-06
Add continent filter pills to event collections

dennispaagman:
I think most people prefer to go to conferences near them (on the same continent) more often. 
This adds continent filters on top of the event pages (upcoming, past, calendar), making it easier to find 'closer' conferences. 

The filter 'pills' are designed similarly to the alphabet filters found on other pages. It persists the filter state across the pages.

## Screenshots

https://github.com/user-attachments/assets/00341f74-635e-4987-87c4-075b859c272a

### DIFF
--- a/app/models/continent.rb
+++ b/app/models/continent.rb
@@ -13,6 +13,8 @@ class Continent
     "south-america" => {name: "South America", alpha2: "SA", emoji: "🌎"}
   }.freeze
 
+  UNPOPULATED = %w[antarctica].freeze # Very unlikely to have an event here ;)
+
   BOUNDS = {
     "africa" => {southwest: [-25.0, -35.0], northeast: [60.0, 40.0]},
     "antarctica" => {southwest: [-180.0, -90.0], northeast: [180.0, -60.0]},
@@ -129,7 +131,11 @@ class Continent
 
   class << self
     def all
-      @all ||= CONTINENT_DATA.keys.map { |slug| new(slug) }
+      @all ||= slugs.map { |slug| new(slug) }
+    end
+
+    def populated
+      @populated ||= (slugs - UNPOPULATED).map { |slug| new(slug) }
     end
 
     def find(term)
--- a/app/controllers/concerns/continent_filterable.rb
+++ b/app/controllers/concerns/continent_filterable.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module ContinentFilterable
+  extend ActiveSupport::Concern
+
+  included do
+    helper_method :selected_continent
+  end
+
+  private
+
+  def selected_continent
+    return @selected_continent if defined?(@selected_continent)
+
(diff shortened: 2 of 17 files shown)


marcoroth (commented):
Great idea, thanks @dennispaagman!

I wonder if we should use this kind of subnav, since that's the one we have been using in the new pages.

Image

What do you think?

dennispaagman:
yeah I was doubting between the two. I'll implement it like those. Do you feel we should add counts as well?

marcoroth:
@dennispaagman I think it would also be fine without the counts, like the lower navigation in the screenshot

dennispaagman:
CleanShot 2026-08-07 at 11 28 36@2x

app/models/continent.rb
```diff
@@ -13,6 +13,8 @@ class Continent
     "south-america" => {name: "South America", alpha2: "SA", emoji: "🌎"}
   }.freeze
 
+  UNPOPULATED = %w[antarctica].freeze # Very unlikely to have an event here ;)
```
marcoroth, line 16: @jasonswett might have a different opinion 😅

dennispaagman: haha, better name more than welcome!

ChaelCodes:
How does this work with online events?

ChaelCodes:
Feel free to ignore this, but I think we're adding this event filter, when maybe we're actually looking for a new view of upcoming events. Could we improve the https://www.rubyevents.org/countries page and promote it here to support filtering events by country?

Map could belong with Upcoming/Past/Archive/Calendar/Meetup.

hschne:
> Feel free to ignore this, but I think we're adding this event filter, when maybe we're actually looking for a new view of upcoming events. Could we improve the https://www.rubyevents.org/countries page and promote it here to support filtering events by country?

> Map could belong with Upcoming/Past/Archive/Calendar/Meetup.

100% agree with @ChaelCodes here. But that shouldn't stop this PR from going out, it's a great improvement over the status quo! 👏

 Ship it :ship:

dennispaagman:
> How does this work with online events?

Oh good question! There aren't currently any planned so I missed that they were possible. 

These events have `country_code = nil` so fall out of every filter except 'All'. Do you think we should add a Online filter pill to be fully covered? I'm not sure if will be many exclusively online events in the future though.

dennispaagman:
> Map could belong with Upcoming/Past/Archive/Calendar/Meetup.

I like that idea! I might work on it a bit in the future.

dennispaagman:
@marcoroth can this be merged?

marcoroth (approved):
Thanks @dennispaagman!

Do you want to follow-up with the "Online" event, or should we create an issue for it so we don't loose it?

dennispaagman:
> Do you want to follow-up with the "Online" event, or should we create an issue for it so we don't loose it?

I'll create a PR for it soon!

PR #2049, opened 2026-08-11
Fix /jobs dashboard in development without authentication

dennispaagman:
I noticed the /jobs dashboard can't be reached in development if you're not authenticated.

## Description
It inherits from ApplicationController by default, which requires authentication via `authenticate_user!.

The dashboard itself is protected via routes and can be accessed without being authenticated in development, so the controller doesn't need to do that.

## Screenshots
CleanShot 2026-08-11 at 09 07 47@2x

I think the "no route matches" error will go away as well, it probably tries to calculate the route in the scope of the jobs dashboard routes instead of the main app. Since it won't inherit that controller anymore anyway this will go away.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -45,5 +45,7 @@ module RubyEvents
 
     # disable Mission Control auth as we use the route Authenticator
     config.mission_control.jobs.http_basic_auth_enabled = false
+    # do not inherit from ApplicationController (the default), because that calls `authenticate_user!`
+    config.mission_control.jobs.base_controller_class = "ActionController::Base"
   end
 end


ChaelCodes (changes requested):
So sorry. I believe that is actually working as designed. We use the admin authentication to determine access to /jobs in production.

For development, if you want access to that page, you can use the omniauth development flow. I checked, and locally, you would have access even if logged in as yourself.

Image
Image
Image

ChaelCodes:
If I'm misunderstanding the issue or what you've done in the PR - please let me know.

dennispaagman:
Hmm I assumed they should be accessible without logging in because two reasons:
1. The routes are specifically rendered without an `authenticate` block in development: https://github.com/rubyevents/rubyevents/blob/0a00ea49cb0e7e037e733db92b689ec43404bb68/config/routes.rb#L49-L59
2. I can access `/admin` without being authenticated.

Maybe we don't need that condition in the routes then and need to fix the redirect instead?

dennispaagman:
@ChaelCodes not sure if you got a notification so @'ing you here 😇

ChaelCodes (approved):
I've tested this in staging, and our admin access is intact. Admins have access, logged in users do not, and unauthenticated users do not. So this should be safe to merge.

PR #2050, opened 2026-08-11
Add data for RubyEnRails 2006

dennispaagman:
## Description
Did some archeology and dug up most info from this conference.

## Screenshots
localhost_3000_events_rubyenrails-2006 (1)
localhost_3000_events_rubyenrails-2006 (2)
localhost_3000_events_rubyenrails-2006 (3)
localhost_3000_events_rubyenrails-2006 (4)
localhost_3000_events_rubyenrails-2006 (5)


data/rubyenrails/rubyenrails-2006/schedule.yml
```diff
@@ -0,0 +1,54 @@
...
+        slots: 1
+        items:
+          - "Ontvangst en registratie"
+
+      - start_time: "09:30"
+        end_time: "09:45"
+        slots: 1
+        items:
+          - title: "Welkom"
+            description: "Gert Oortgiesen, directeur van sponsor Profict, opent de dag met een kort welkomstwoord."
```
dennispaagman, line 18: stuff like this is in Dutch, do you want to keep it like this to match the original schedule, translate it, or just remote it and just keep a "welcome" slot?

ChaelCodes: Definitely keep it in Dutch. For talks, we have a [separate process.](https://github.com/rubyevents/rubyevents/blob/main/docs/ADDING_UNPUBLISHED_TALKS.md#faq)

data/rubyenrails/rubyenrails-2006/videos.yml
```diff
@@ -1,6 +1,81 @@
...
+# Slides: https://web.archive.org/web/20060619184010/http://rubyenrails.nl/articles/2006/05/25/presentaties-online
 ---
-[]
+
+- id: "frank-oxener-rubyenrails-2006"
+  title: "Introductie Rails"
+  event_name: "RubyEnRails 2006"
+  date: "2006-05-18"
+  language: "Dutch"
+  slides_url: "https://web.archive.org/web/20060526174918/http://agiledovadi.com/blog/articles/2006/05/22/workshop-rubyenrails-2006"
```
ChaelCodes: I think this would be better as an Additional Resource because it's a blog post and not slides.

https://github.com/rubyevents/rubyevents/blob/4f34e3e00ed63838bcbd18bed8e1de4db1a57378/app/schemas/video_schema.rb#L36
https://github.com/rubyevents/rubyevents/blob/main/app/schemas/additional_resource_schema.rb

We haven't done a really good job documenting this capability.

ChaelCodes (commented):
Thank you so much for adding this!! I love when we get detail on these historical events! 

I left a note about the additional resources and wanted to give you the option to translate titles for the talks to match our guidance around talks in other languages.

Honestly, I think this is good to merge as-is, but I wanted to give you a chance to respond to the feedback.

ChaelCodes (approved):

dennispaagman:
@ChaelCodes thanks for the feedback, that was exactly what I was looking for.

I've added a small drive by commit to update the doc, as it actually needs the full language name instead of the language code.

ChaelCodes (approved):
Thank you so much!! 🙌

dennispaagman:
The talks aren't visible yet on https://www.rubyevents.org/events/rubyenrails-2006/talks, is that a problem or should a sync still happen?

Issue #1987, opened 2026-07-25
Allow to be able to mark a conference as "interested"


marcoroth:
Right now you can only mark an event as "attending", but you can't share that you are only "interested" or are "considering going".

Inspired by a conversation with @headius at RubyConf.

headius:
Thanks for getting this added! Sorry I forgot to open an issue.

You've basically described it, but the background is that I was marking conferences as "attending" so I would not lose track of them while considering upcoming CFPs etc. I really wanted a way to "bookmark" upcoming events, but that seems to dovetail well with "interested".

Could also be a great way for conferences to track interest... e.g. put a button on your conference announcement post that says "I'm interested!" linking to this feature on rubyevents.

Looking forward to it. I'd love to help but I can't promise I'll have any time.

PR #2054, opened 2026-08-13
Fix:1987 Added option to mark event as interested

yashika279:
## Screenshots

https://github.com/user-attachments/assets/e38924e1-3c6f-44e8-9cc5-cff346cf7e2d

## References
https://github.com/rubyevents/rubyevents/issues/1987

### DIFF
--- a/app/controllers/api/v1/embed/stickers_controller.rb
+++ b/app/controllers/api/v1/embed/stickers_controller.rb
@@ -9,7 +9,7 @@ module Api
           user ||= User.find_by_github_handle(params[:slug])
           raise ActiveRecord::RecordNotFound, "User not found" unless user
 
-          events = user.participated_events.includes(:series).select(&:sticker?)
+          events = user.attended_events.includes(:series).select(&:sticker?)
 
           stickers = events.flat_map { |event|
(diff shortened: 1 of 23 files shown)


ChaelCodes:
Hey! I just wanted to pop in and let you know that I'm reviewing this one, but I'm thinking about it from a feature perspective and considering things like:
- Where do we want to display events you're interested in?
- Do we need a new filter on event views? What should that look like? Would it be better to keep it on the profile?
- Should we display events that someone is interested in?
  - Should that be a setting on the user profile?
  - Will it look bad that some events have no people interested in them?

So it'll take longer than usual to get back to you with good feedback and/or merge it.

Thank you for taking on a larger chunk of functionality like this!

yashika279:
@ChaelCodes any update on this one.

ChaelCodes:
I'm thinking this should be like a bookmark - private, and not a kind of event participation, so it doesn't get you a sticker or stamp automatically instead of filtering for attendance.

Then it would pair with filters for events and cfps.

But I'm curious what @marcoroth thinks because it's his issue.

Right now, we're working on adoption of the event attendance feature and having two options for event attendance present does add some confusion.

PR #2055, opened 2026-08-14
Add North West Ruby User Group meetup and assets

tekin:
## Description

Adds the basic details for the [NWRUG](https://nwrug.org) meet-up. A populated videos.yml to follow...

## Screenshots

Screenshot 2026-08-14 at 12 09 13


data/speakers.yml
```diff
@@ -14177,6 +14177,10 @@
 - name: "Robb Kidd"
   github: ""
   slug: "robb-kidd"
+- name: "Rob Whittaker"
+  github: "purinkle"
+  mastodon: "https://ruby.social/@purinkle"
+  slug: "rob-whtittaker"
```
ChaelCodes, line 14180: Hi!! I'm so excited to see another meetup join the index. However, it looks like CI is failing. You'll need to call `yerba apply` locally, but I'm pretty sure it's because this one isn't alphabetized.

tekin:
Thanks @ChaelCodes, that should be resolved now.

ChaelCodes:
Thank you so much for adding this!! 🙌

tekin:
Thanks @ChaelCodes. How often does the app get deployed? I'm wondering when we'll see our event on the site.

marcoroth:
@tekin the app gets deployed whenever things merge on main, but we need to run the "seeds" manually. Hopefully that is changing once #1988 merges. But that should be live now.

Issue #2062, opened 2026-08-15
Validate the slug for each sponsor is the sponsor's name parameterized.


ChaelCodes:
# Description

Create a new Static::Validation that the slug for a sponsor is the sponsor's name parameterized. There's a `parameterize` method in ActiveSupport that converts spaces to `-` and lowercases everything, and removes weird symbols. That's the one we want to use.

PR #2066, opened 2026-08-17
Integrated sponser validator

yashika279:
## Description
Validate the slug for each sponsor is the sponsor's name parameterized.
 #2062

## Screenshots
image

## Testing Steps
To try it on a real file, open a Rails console and pass a sponsors.yml path:

Static::Validators::SponsorSlug.new(
  file_path: "data/rubyconf/rubyconf-2026/sponsors.yml"
).errors

## References
https://github.com/rubyevents/rubyevents/issues/2062


ChaelCodes:
So the first problem is mine. We don't actually verify sponsors in validate.rake yet - we need to fix that before this'll run. 😅 

The second problem is that I added it locally, and this took over 10 minutes to run on all the files.
Screenshot 2026-08-17 at 1 01 12 PM

yashika279:
> So the first problem is mine. We don't actually verify sponsors in validate.rake yet - we need to fix that before this'll run. 😅
> 
> The second problem is that I added it locally, and this took over 10 minutes to run on all the files. Screenshot 2026-08-17 at 1 01 12 PM

Hello @ChaelCodes Thank you for the feedback, I am looking into this.

yashika279:
@ChaelCodes I need help in fixing this at local
image

I ran `bin/rails talk_ids:backfill_old_ids` and still getting above errors.
 I added sponsors in validate.rake as well.

ChaelCodes:
@yashika279 I think a rebase and a `bin/rails db:reset` should clear those out. 

But you shouldn't need to get videos passing locally, just `bin/rails validate: sponsors`.

Side note, I think the performance problem is traversing the document. Yerba passes the Ruby/Rust boundary for every record. Getting all the values from the document in one call might help. 

I ran into this issue when I built a similar validation for speaker slugs.

yashika279:
@ChaelCodes on local as well this check is breaking I am not getting how to fix it

yashika279:
@ChaelCodes the CI checks are expected to be break as we have added a validation for sponser please check once you have enough time

ChaelCodes:
You have to update the data. Sorry.

PR #2078, opened 2026-08-22
Add DenHaag.rb meetup

dennispaagman:
DenHaag.rb (https://www.meetup.com/denhaagrb/) is a Ruby user group that
has been running in and around The Hague since 2012. Amsterdam.rb was the
only Dutch series in the repo until now.

## Screenshots
localhost_3000_events_denhaag-rb-meetup
localhost_3000_events_denhaag-rb-meetup (1)
localhost_3000_events_denhaag-rb-meetup (2)
localhost_3000_events_denhaag-rb-meetup (3)
localhost_3000_events_denhaag-rb-meetup (4)


data/denhaag-rb/denhaag-rb-meetup/videos.yml
```diff
@@ -0,0 +1,389 @@
...
+
+- id: "denhaag-rb-march-2013"
+  title: "DenHaag.rb March 2013"
+  location: "Rotterdam, Netherlands"
+  event_name: "DenHaag.rb March 2013"
+  date: "2013-03-05"
+  video_provider: "not_recorded"
+  video_id: "denhaag-rb-march-2013"
+  description: |-
+    Joint meetup: Rotterdam.rb and Denhaag.rb at Floorplanner.com. We will join Rotterdam.rb at March 5th.
```
dennispaagman, line 65: there are a couple of these jointly organised meetups (i'm also adding Rotterdam.rb later). What's the preferred solution for them, duplicates or only add them in one of the two meetups?

ChaelCodes: I think duplicate until we can add something a bit more sophisticated.

ChaelCodes: We'll be able to query for them if they're duplicated.

data/featured_cities.yml
```diff
@@ -530,6 +530,16 @@
   aliases:
     - "TPE"
 
+- name: "The Hague"
```
ChaelCodes, line 533: Can you tell me why you added this? Was it due to an error?

ChaelCodes (approved):

data/featured_cities.yml
dennispaagman: Hmmm not sure, Claude decided to do that. Should I make a PR to remove it? I think we only lose the aliases in search and potentially the validate lookup in `Static::Validators::EventCityNames`

Issue #2070, opened 2026-08-18
event-data skill's speaker slug guidance contradicts the SlugMatchesName validator


crmne:
`.agents/skills/event-data/SKILL.md:120` says:

> **Slug convention:** Prefer the speaker's GitHub handle as the slug (e.g., `slug: "tenderlove"` for `github: "tenderlove"`). Only transliterate or generate a slug from the name when the speaker has no GitHub handle.

`Static::Validators::SlugMatchesName` requires `slug == name.parameterize`, with no exception for GitHub handles, and applies the same rule to aliases.

The example in the doc is contradicted by the data it describes. `data/speakers.yml` has:

```yaml
- name: "Aaron Patterson"
  github: "tenderlove"
  slug: "aaron-patterson"
```

The guidance predates the validator: the line landed in #1813 (2026-06-24), the validator in #2065.

Anyone following the skill writes `speakers.yml` entries that fail `bin/rails validate:speakers`. I hit this while adding four speakers in #2069, and had to rewrite all the slugs.

Looks like a one-line doc fix. Happy to send a PR if useful.

ChaelCodes:
Sorry about that. 😅 The GitHub handle was only supposed to be preferred to transliterating - for the Japanese Ruby community.

If you want to handle it, feel free, otherwise I'll get to it sometime this week.

ChaelCodes:
This is related to #2060

PR #2079, opened 2026-08-23
Validate speaker slugs use GitHub if they cannot be parameterized AND Sync Speakers in the TalkGenerator

### LINKED ISSUE #2060: AI is using github profile when generating slugs
# Description

In several recent PRs, new speakers have been added with the GitHub profile as the slug instead of the parameterized name.

Solutions:

- [ ] Have the talk generator add speakers if they are missing.
- [ ] Warn on speakers missing GitHub profiles (not error) (passes CI)
- [ ] New SKILL for finding GitHub profiles
- [x] Have the Speakerfile add speakers if they are missing with the correct slug
- [x] New static::validation for speaker slug matching speaker's name slugified.

ChaelCodes:
### Speaker sync in the talk generator

TalkGenerator updates speakers.yml. For each talk, it adds the speaker with an empty GitHub handle if they are not already present.

### Slug validation for non-parameterizable names

The `Static::Validators::SpeakerSlugMatchesName` validator previously skipped validation entirely when a speaker's name parameterized to an empty string (e.g. names written in non-Latin scripts like `まつもとゆきひろ`).

This PR tightens that rule:

- If the parameterized name is empty and the speaker has a `github` handle, the slug must match the GitHub handle (otherwise an error is reported: "Name cannot be parameterized. Slug must be the GitHub handle, expected: <slug>.")
- If there's no GitHub handle, validation is still skipped (any slug allowed)
- Aliases remain exempt — they don't carry their own GitHub handle and may be used for redirecting from the prior name

Also renames the validator file/class from `SlugMatchesName` to `SpeakerSlugMatchesName` because there is 0 chance it will be reused now.

### Fixes event-data

Event-data used to encourage the AI to use github handles instead of parameterizing the name. We're only supposed to do that instead of transliterating the name.

## Testing Steps

- `bin/rails g talk --event "rails-camp-west-2026" --speakers "Caleb Owens"`

## References
- closes #2060
- references #2070

### DIFF
--- a/app/models/static/validators/speaker_slug_matches_name.rb
+++ b/app/models/static/validators/speaker_slug_matches_name.rb
@@ -2,7 +2,7 @@
 
 module Static
   module Validators
-    class SlugMatchesName
+    class SpeakerSlugMatchesName
       def initialize(file_path:, document: nil)
         @file_path = file_path
         @document = document
--- a/app/models/static/validators/speaker_slug_matches_name.rb
+++ b/app/models/static/validators/speaker_slug_matches_name.rb
@@ -30,7 +30,7 @@ module Static
         errors = []
 
         @document ||= Yerba.parse_file(@file_path)
-        name_slug_pairs = @document.pluck(:name).zip(@document.pluck(:slug))
+        name_slug_pairs = @document.pluck(:name).zip(@document.pluck(:slug), @document.pluck(:github))
         errors.concat(find_errors_for_name_slug_pairs(name_slug_pairs) do |name|
           @document.find_by(name: name)&.[]("slug")
         end)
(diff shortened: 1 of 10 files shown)


app/models/static/validators/speaker_slug_matches_name.rb
```diff
@@ -2,7 +2,7 @@
 
 module Static
   module Validators
-    class SlugMatchesName
+    class SpeakerSlugMatchesName
```
ChaelCodes, line 5: Renaming because there is 0 chance this will be shared now.

data/speakers.yml
```diff
@@ -2171,6 +2171,7 @@
 - name: "beta_chelsea"
   github: "betachelsea"
   twitter: "beta_chelsea"
+  speakerdeck: "betachelsea"
```
ChaelCodes, line 2174: Found these because index_for now searches by alias too.

PR #2086, opened 2026-08-28
Allow an event to remove talks

marcoroth:
This pull request adds a `removed_talk_ids` key to `event.yml` so an event can declare that a talk was deliberately dropped from its `videos.yml`.

`bin/rails validate:videos` flags every talk id that disappears from a `videos.yml`, because a vanished id is usually an unrecorded rename and production still knows the talk under the old id. 

Until now there was no way to say the removal was intentional. `Static::Validators::TalkRenames` only subtracted `old_id`s, so a deliberate deletion failed CI with a message that ended "If the talk was removed on purpose, this is fine.", a state you could not actually express.

This came out of #2085


ChaelCodes:
Do we remove ids from the event.yml after?

marcoroth:
@ChaelCodes yeah, I'd say so

PR #2095, opened 2026-09-01
Add Twente.rb meetup

dennispaagman:
Adds the [Twente.rb](https://www.meetup.com/twenterb/) series and its 27 meetup editions (2010-2023), sourced from meetup.com and the group's Google Group archive (https://groups.google.com/g/twente-rb), which covers the pre-meetup.com era and fills gaps in later editions.

They don't seem to have a logo or anything, so I took the picture from the Meetup page.

## Screenshots
twente-rb-meetup-overview
twente-rb-meetup-overview
twente-rb-meetup-overview-talks


ChaelCodes (approved):
Thank you adding these events! I'm not sure what to do about the lack of a logo. I wonder if the default banner would be a better option until we can find an organizer to supply one?

PR #2115, opened 2026-09-10
Add upcoming B'more on Rails events

alimi:
## Description
B'more on Rails has a couple of events coming up: one tonight and another in October.


marcoroth (approved):
Awesome, thank you @alimi! 🙏🏼

PR #2117, opened 2026-09-10
Withdrawn: Berlin.rb September 2026 venue

viktorianer:
Withdrawn. I had the venue wrong — the existing description is correct as it stands. No change needed.


viktorianer:
Closing this — I had the venue wrong. The evening is an ablefy event and the existing description is correct as it stands. Sorry for the noise.

PR #2122, opened 2026-09-12
Add theater mode to talk videos

kcdragon:
## Description

Adds a YouTube-style theater mode to talk pages. The video expands to fill the viewport width while the page keeps scrolling normally.

A Theater pill in the action row (next to Bookmark and Play on YouTube) toggles it, and the preference is remembered across talks via localStorage. Desktop only (lg and up); on smaller screens the video is already full width, so it's a no-op.

The button lives outside the player because YouTube's iframe owns both right-hand corners (volume/CC/settings on top, fullscreen and "More videos" on the bottom), so any in-player overlay collides with something.

### Desktop (has theater button, not in theater mode)

Screenshot 2026-09-13 at 9 25 37 PM

### Desktop (has theater button, in theater mode)

Screenshot 2026-09-13 at 9 25 47 PM

### Mobile (no theater button)

Screenshot 2026-09-13 at 9 26 06 PM

## Testing Steps

1. Open any talk page on a screen at least 1024px wide.
2. Click Theater in the button row below the video.
   - The video expands to fill the window width, and the related-talks sidebar disappears.
   - The button label changes to Exit theater and its icon flips to the compress arrows.
   - The page still scrolls normally — the title, tabs, and transcript are all reachable below.
3. Click Exit theater. The layout returns to normal and the sidebar comes back.
4. Turn theater mode on, then reload the page. It should still be on.
5. Navigate to a different talk. The preference should carry across.
6. Narrow the window below 1024px. Theater styling should not apply, and the Theater button should be hidden.
7. Check a few providers — a YouTube talk, an mp4 talk, and a talk with no video (the button should be absent on that last one).


app/views/talks/_video_player.html.erb
```diff
@@ -20,8 +20,8 @@
       } do %>
   <div
     class="
-      banner-img card-horizontal-img relative aspect-video w-full max-w-full overflow-hidden rounded-xl bg-gray-300
-      outline outline-1 outline-gray-300
+      banner-img card-horizontal-img theater-stage relative aspect-video w-full max-w-full overflow-hidden
+      rounded-xl bg-gray-300 outline outline-1 outline-gray-300
```
marcoroth, line 24: I think we want to remove the `rounded-xl` if we go edge-to-edge
```suggestion
      bg-gray-300 outline outline-1 outline-gray-300
```

app/views/talks/_video_player.html.erb
marcoroth, line 69: Looks like we'd also need to remove the `rounded-xl` here

app/views/talks/_video_player.html.erb
kcdragon: @marcoroth were you thinking we always remove the rounded corners or just when theater mode is enabled?

marcoroth: I think just in theater mode, as it looks weird without rounded corners. Otherwise You might have to add some padding on either side in theater mode

app/assets/stylesheets/components/video.css
```diff
@@ -32,4 +32,43 @@
...
+  }
+
+  .theater-mode .theater-player .theater-stage {
+    @apply lg:mx-auto lg:h-auto lg:max-h-[calc(100vh-5rem)] lg:w-auto lg:max-w-full lg:rounded-none lg:outline-none;
+  }
+
+  /* Layers that fill the stage keep their own corners once it goes square, so
+     their rounding shows through. Scoped to inset-0 so nested UI that is meant
+     to be round (the play button) is left alone. */
+  .theater-mode .theater-player .theater-stage > *,
```
kcdragon, line 69: @marcoroth here's the CSS rule I added to remove the rounded corners. Screenshots are in the PR. TBH I'm not that good with CSS and this is what Claude came up with. It does work though. Let me know your thoughts and if there's a better approach.

Issue #2119, opened 2026-09-11
XO Ruby Montréal doesn't show up in Montréal


ChaelCodes:
# Description

We've had recurring issues where events do not show up in the city, region, or country they're in. 

In this case, XO Ruby Montréal isn't showing up in Montréal.

![Screenshot_20260911-065117.png](https://github.com/user-attachments/assets/4d26cd51-1af5-4856-b7db-8eea5302b0e4)
![Screenshot_20260911-065126.png](https://github.com/user-attachments/assets/a6b14f49-d765-4765-944f-8a8f15c41a77)

PR #2126, opened 2026-09-15
Fixed: XO Ruby Montréal doesn't show up in Montréal

yashika279:
## Description
Static::Event#import_event! — Sets state_code from the location (e.g. QC from "Montréal, QC, Canada"), falls back to featured city data when the state is omitted, and uses the canonical featured city name.

Event#location_and_country_code — Skips appending series.default_country_code when the location already includes a country, so Canadian XO Ruby events no longer geocode as "…, Canada, US".

## Screenshots
image

## Testing Steps
None

## References
https://github.com/rubyevents/rubyevents/issues/2119


app/models/static/event.rb
```diff
@@ -253,6 +253,44 @@ def city
...
+
+      Static::City.all.detect do |featured_city|
+        if country.present?
+          next unless featured_city.country_code.to_s.upcase == country.alpha2
+        end
+
+        featured_city.name.casecmp?(city) ||
+          Array(featured_city.aliases).any? { |alias_name| alias_name.casecmp?(city) }
+      end
+    end
```
sudeeptarlekar, line 281: May be we can combine these methods? Looks like they are related so why split?

Moreover this patterns seems to be norm in LLMs (unnecessary method splitting), be careful about that or on other hand may be I should catch up with new Ruby way of writing a code...

And you can use `||=`

```
@feature ||= find_featured_city
```

sudeeptarlekar:
Looks good overall with some nit-picky things, I will leave approval to maintainers...

PR #2131, opened 2026-09-17
Ignore mise files

madisonsites:
## Description

Ignore the file that my [mise](https://mise.jdx.dev/) keeps trying to generate.


ChaelCodes (approved):
Thanks for updating this!

PR #2133, opened 2026-09-18
Migrate RubyLLM::Schema to Schematist

### LINKED ISSUE #2051: Migrate `RubyLLM::Schema` to `Schematist `
https://paolino.me/schematist/
https://github.com/crmne/schematist
https://github.com/crmne/schematist#migrating-from-ruby_llm-schema

samuel-sarmah:
## Description

RubyLLM::Schema was renamed to Schematist in its 1.0 release, so this swaps the gem and updates the app to the new API.
  - Gemfile: `ruby_llm-schema ~> 0.4.0` to `schematist ~> 1.1`. ruby_llm-schema stays
    in Gemfile.lock as a transitive dependency of ruby_llm, which still
    declares it.
  - ApplicationSchema now inherits from Schematist::Schema.
  - ApplicationSchema.json_schema drops the provider-envelope digging.
    to_json_schema used to return `{name:, description:, schema:, strict:}` (the
    shape OpenAI's response_format expects) and now returns the Draft 2020-12
    document itself, string-keyed, so `new.to_json_schema[:schema]` becomes
   ` new.to_json_schema`.
  - FeaturedCitySchema#to_json_schema reaches into the result to make state_code
    nullable; updated for the flattened, string-keyed shape.
  - Regenerated the 20 exported schemas in lib/schemas/ via bin/rails 
    schema:export.

The exported JSON changes in exactly three ways, in every file: "$schema" and
  "title" are added, and "strict": true is removed (it was an OpenAI
  response_format flag rather than a JSON Schema keyword, so the gem dropped
  it). Nothing semantic moved comparing each file before and after with key
  order normalised and those three keys ignored, all 20 are identical. Some
  diffs also show required shifting position within a then block, which is key
  ordering only

## Screenshots

Not applicable, not a user-facing change

## Testing Steps

These schemas back the data/ YAML validation, so the
  validators are the interesting part:

  1. bundle install
  2. bin/rails schema:export  should produce no diff beyond what is committed
     here
  3. bin/rails validate:all Yerbafile rules, event.yml, speakers.yml and
     videos.yml all pass, confirming the new Draft 2020-12 $schema declaration
     does not change how Yerba validates data/
  4. bin/rails test 1866 runs, 4708 assertions, 0 failures, 0 errors, 2 skips

  test/lib/schema_export_test.rb covers step 2 by asserting every schema is in
  sync with its exported JSON.

## References

- closes #2051
- Schematist/docs
- Migrating from ruby-llm-schema


marcoroth (approved):
Thank you @samuel-sarmah! 🙏🏼

samuel-sarmah:
You're welcome @marcoroth

